PR #14, opened 2023-02-11
adding mod info panel functionality

oceancabbage:
Adding functionality to show information about a mod when that mod if clicked on in either mod list. Easily editable to show more/less information.


twstagg (approved):
No issues found! Nice work!

PR #16, opened 2023-02-11
trying out using the RimPy db in sorting too

oceancabbage:
Initial onboarding for parsing RimPy's `db.json` and adding the rules in there into the dependency generation. In my testing of 407 mods, this resulted in a 30 count increase in dependency rules, showing that there indeed are rules in `db.json` that are not encapsulated in `communityRules.json` or in the workshop mods `About.xml`s. However, simply adding the dependencies in this manner to the toposort does not get us to parity, so there is still something missing. However, we may be one step closer because of this.

### DIFF
--- a/util/mods.py
+++ b/util/mods.py
@@ -1,8 +1,10 @@
+import json
 import os
 from typing import Any, Dict, List, Tuple
+
 from PySide2.QtWidgets import *
-import json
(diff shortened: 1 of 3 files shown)


twstagg:
continuing in `sorting`

PR #54, opened 2023-03-29
Added support for No Version Warning

emipa606:
Since I got requests of adding support for my [No Version Warning-mod](https://steamcommunity.com/sharedfiles/filedetails/?id=2599504692) to RimPy but never could since that one is closed source, so Ill try here instead.

If the mod is subscribed it adds the supported version-tags to the applicable mods when generating the cache.
Can probably be improved and/or put behind some option but I'm not that experienced with GUI-coding in Python to do that.
Hopefully you will get any use of this.

Nice initiative with the project!

### DIFF
--- a/view/main_content_panel.py
+++ b/view/main_content_panel.py
@@ -2,6 +2,7 @@ import logging
 from multiprocessing import active_children, Process
 import os
 import platform
+import re
 import subprocess
 from threading import Thread
 from typing import Any, Dict
@@ -53,7 +54,8 @@ class MainContent:
         self.main_layout.setContentsMargins(
             5, 5, 5, 5
         )  # Space between widgets and Frame border
-        self.main_layout.setSpacing(5)  # Space beteen mod lists and action buttons
+        # Space beteen mod lists and action buttons
+        self.main_layout.setSpacing(5)
 
         # FRAME REQUIRED - to allow for styling
         self.main_layout_frame = QFrame()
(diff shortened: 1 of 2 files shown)


view/main_content_panel.py
```diff
@@ -262,11 +268,60 @@ def refresh_cache_calculations(self) -> None:
...
             f"Combined {len(self.expansions)} expansions, {len(self.local_mods)} local mods, and {len(self.workshop_mods)}. Total elements to get dependencies for: {len(all_mods)}"
         )
 
+        # Check mods verified working with newer game-versions than defined
+        # https://steamcommunity.com/sharedfiles/filedetails/?id=2599504692
+        package_id_to_uuid = {}
+        for mod_uuid, modmetadata in all_mods.items():
+            package_id_to_uuid[modmetadata["packageId"]] = mod_uuid
+        if "mlie.noversionwarning" in package_id_to_uuid.keys():
+            no_version_mod = all_mods[package_id_to_uuid["mlie.noversionwarning"]]
```
oceancabbage, line 283: Minor: may want to extract this to a function in the module

oceancabbage:
This looks really good. Was able to test on my machine that this works as expected. Only left one minor comment but I'm happy with merging this.

twstagg (changes requested):
I like the idea. I agree with @oceancabbage previous comment. It is of my opinion that:

    This feature should be made configurable (default: off) so that users can opt out of using it entirely.

view/main_content_panel.py
emipa606: That code was cloned from another part of the mod so it would probably be a good idea

twstagg:
please cherry pick your commits and rebase off of main again - sprint-a1.0.3 has been merged

PR #74, opened 2023-05-08
Fix .xml extension issues exporting, fix typo in import func, addt'l autodetect paths for Linux

### LINKED ISSUE #52: FIXME: Rimsort writes .xml twice if you change the name of the modlist when exporting
- [x] Fix `_do_export`

OS: Windows 10 Pro 10.0.19045

1. Export your modlist somewhere
2. In my case, Windows autofills the name with the extension of the file
image
3. If the modlist is exported like this, rimsort will add a second .xml at the end of the files, even with the first one already being in place
image
4. Changing the file's name and not writing the extension at the end of the name will have similar results
image
image

cylian914:
re fix #52 
the issue don't seem to appear on linux, its an issue with the windows file manager who add the .xml after selection the file i think

### DIFF
--- a/view/game_configuration_panel.py
+++ b/view/game_configuration_panel.py
@@ -826,11 +826,18 @@ class GameConfiguration(QObject):
             f"/Users/{getpass.getuser()}/Library/Application Support/Rimworld/Config/",
             f"/Users/{getpass.getuser()}/Library/Application Support/Steam/steamapps/workshop/content/294100/",
         ]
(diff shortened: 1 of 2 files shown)


loveboylion:
I can conform your fix works on windows 
when you save to a new file, or if you choose an existing file to overwrite

loveboylion:
I can conform your fix works on windows
import now does not error an works

PR #75, opened 2023-05-09
Added a progress bar for steamcmd runner

cylian914:
Added a progress bar for the runner of steamcmd on top of the screen as it was confusing at when it ends


twstagg:
I checked this out locally and it works flawlessly for me on Linux. I like it! :man_dancing:

Issue #72, opened 2023-05-08
Implement feature related to collection


cylian914:
- [x]  Fix download mod from a collection (im doing it)

- [ ] Allow to convert a collection into a mod list

ask and test before removing issues

PR #77, opened 2023-05-10
Adding a warning box when mod and add value to processbar when failed

cylian914:
add a warning at the end of downloading mods id some has file


cylian914:
It also fix #72 2nd checkbox

twstagg:
Here is an additional case that needs factored for this as well:

image

[steamcmd.txt](https://github.com/oceancabbage/RimSort/files/11501483/steamcmd.txt)

window/runner_panel.py
```diff
@@ -17,6 +17,11 @@
     QProgressBar,
 )
 
+from model.dialogue import show_warning
+from util.steam.webapi.wrapper import (
+    ISteamRemoteStorage_GetPublishedFileDetails as get_mod_data,
```
twstagg, line 22: Please don't mask this and just call it by it's actual name.

window/runner_panel.py
```diff
@@ -29,7 +34,7 @@ def __init__(self, todds_dry_run_support=False):
         logger.info("Initializing RunnerPanel")
         self.ansi_escape = compile(r"\x1B\[[0-?]*[ -/]*[@-~]")
         self.system = system()
-
+        self.warningmod = []
```
twstagg, line 37: Let's call this `self.steamcmd_failed_mods` or something like that. Please also add a comment describing what it is being used for.

window/runner_panel.py
```diff
@@ -236,3 +247,16 @@ def finished(self):
...
                 self.message("Subprocess completed.")
+                if "steamcmd" in self.process.program():
+                    if self.warningmod != []:
+                        tempdata = ""
+                        for i in get_mod_data(self.warningmod)["response"][
+                            "publishedfiledetails"
+                        ]:
+                            tempdata = tempdata + i["title"] + "\n"
+                        show_warning(
+                            title="WARNING steamcmd",
```
twstagg, line 258: Please change this to:

```
                        show_warning(
                            title="SteamCMD downloader",
                            text="SteamCMD failed to download mod(s)! Please try to download these mods again!",
                            information='Click "Show Details" to see the full report.',
                            details=tempdata,
                        )
```

twstagg (changes requested):
Overall looks good! Nice work so far!

Please see my additional comment on the PR for an additional case that will need handled as well. I'd like to do some testing to make sure we are parsing the SteamCMD output for any kind of failure we can create.

cylian914:
Every thing should be fix now

twstagg (approved):

window/runner_panel.py
```diff
@@ -236,3 +252,23 @@ def finished(self):
                 self.process_killed = False
             else:
                 self.message("Subprocess completed.")
+                if "steamcmd" in self.process.program():
+                    if self.steamcmd_failed_mods != []:
+                        tempdata = "Failed to connect"  # Default value
+                        GetPublishedFileDetails = (
```
twstagg, line 258: This relies on there being an internet connection, and if a mod failed because no internet connection... we get crash. I will add sanitization to avoid the crash, but additional handling would still need done, and since we don't really need to display the names to the user to prompt them to redownload, instead of doing this warning, we could possibly instead do a prompt to redownload or retry  button for the failed operation?

I think that'd be a great compromise, and would actually be even more convenient for the user. You could just use `model.dialogue.show_dialogue_conditional` to prompt a No/Yes prompt like we do for the "Setup SteamCMD" stuffs when it prompts to reinstall or do the symlink and have it say:

"X mod(s) failed to download. Do you want to try to redownload these mods?"

And then skip the lookup with WebAPI because you basically have to just iterate through the downloader_list and remove the mods that succeeeded, based on the pfids added to your list `self.steamcmd_failed_mods`

We should also be doing a `runner.message(self.steamcmd_failed_mods)` of the failed mods so that the user can retrieve the pfids from the `RunnerPanel` if push comes to shove

twstagg (changes requested):

window/runner_panel.py
twstagg: In addition to the above, please also add an effect to the Progress Bar that indicates that not all mods failed. I think that changing the color of the progress to a yellow or red would suffice. The percentage should also not add up to 100% if a mod failed. It should only show the percentage for mods that were successfully downloaded out of the total mods being downloaded.

PR #110, opened 2023-06-16
merge  Rw log (finaly)

cylian914:


twstagg:
Please split each major change into separate requests so we can review each individual fix/feature

PR #111, opened 2023-06-16
Change UI

cylian914:
all change in the UI (and thing that i forgot to push before)


twstagg:
Please split each major change into separate requests so we can review each individual fix/feature

PR #113, opened 2023-06-17
[FIX] bugfix for missing loadAfter/loadBefore warnings on the UI

oceancabbage:
### Notes

There's a bug where `loadAfter` and `loadBefore` violations aren't showing in the UI as warning/error labels (on individual mod cells and also on the warnings/error summary). This was due to code looking in a `uuid`-keyed dict for mods with package IDs. I've made a small fix so we're looking in the right place.

### Testing

Manual testing and verification through running RimSort locally and dragging around mods

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,5 +17,6 @@
         "--enable=W0611", "--enable=W0614"
     ],
     "python.terminal.activateEnvironment": true,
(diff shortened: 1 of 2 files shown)


twstagg (approved):
Thank you sir

Really appreciate this :clap:

Issue #162, opened 2023-09-06
FIXME: RimSort isn't picking up my mod's version information


jefferyharrell:
This is in reference to this mod: https://steamcommunity.com/sharedfiles/filedetails/?id=3027381000. RimSort says the mod version is "Not specified."

<img width="452" alt="image" src="https://github.com/RimSort/RimSort/assets/8659114/81f7fdd2-6234-4969-a735-55b100aa65fe">

Except it is so. This is the relevant part of my About.xml:

```xml
<ModMetaData>
  <packageId>CaptainArbitrary.CompactWorkTab</packageId>
  <name>Compact Work Tab</name>
  <author>CaptainArbitrary</author>
  <description>Adds vertical column labels and scroll-wheel support to make the Work tab a little more user-friendly.</description>
  <modVersion>1.1.2</modVersion>
  ...
```

twstagg:
Fixed with #168

PR #163, opened 2023-09-07
FIX: RimSort isn't picking mod's version information #162

LionelColaso:


oceancabbage:
Could you add a short description providing some context for this change? From what I can gather fix here is changing `modVersion` to `modversion`. Is the issue still persisting? Sorry been out of the loop for a while

LionelColaso:
Yes for Some Reasone it Doesnt like uppercase just renaming modVersion to modversion Works also the rest is just Cosmetics i had missed one change so i am closing this and reopening another pull request

PR #168, opened 2023-09-27
FIX: RimSort isn't picking mod's version information #162

LionelColaso:
Seems like it Does not Like Upper Case Characters

### DIFF
--- a/model/mod_list_item.py
+++ b/model/mod_list_item.py
@@ -184,9 +184,9 @@ class ModListItemInner(QWidget):
             author_line = f"Author: {authors_tag if authors_tag else 'Not specified'}\n"
 
         package_id_line = f"PackageID: {self.json_data.get('packageid')}\n"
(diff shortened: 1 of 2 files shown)


Chunnyluny:
sorry, but how am am I able to do a merge ? we as contributors shouldn't be able to do changes on merge requests or pull requests ?

oceancabbage:
> sorry, but how am am I able to do a merge ? we as contributors shouldn't be able to do changes on merge requests or pull requests ?

Whoops, didn't see this until just now. If a PR has enough approvals there should be a big green button near the bottom `Squash & Merge`. If you don't see it that might be because of the way permissions are set up for the repo. I'll make a note to take a look at that-- imo authors of PRs should be able to merge their own PRs and potentially any contributor can merge an approved PR (may have to raise review requirement to at least 2 for that, to be safe).

Issue #171, opened 2023-09-27
FIXME: Conflict with gog Mac version as it is named RimWorld.app not RimWorldMac.app


Ant-V:
**Describe the bug**
It is not finding the game directory, since it is named RimWorld.app instead of RimWorldMac.app

**To Reproduce**
On any new installation from gog on macOS it cannot autodetect the `game_folder` or the `local_folder`.
Setting the paths by editing `settings.json` to 
```
    "game_folder": "/Applications/" **OR** "/Applications/RimWorld.app"
    "local_folder": "/Applications/RimWorld.app/Mods",
```

causes a pop up with the error:
> ***Issue Getting Game Version***
> RimSort is unable to get the game version at the expected path: [/Applications/RimWorldMac.app/Version.txt]. Is your game path [/Applications/RimWorldMac.app] set correctly? There should be a Version.txt file in the game install directory.

Followed by a crash once `OK` is clicked

**Expected behavior**
Simple solutions may include: 
* remove default name appending in path
* allow selection of folders or .app when selecting path
* check for both naming schemes

**Please complete the following information:**
 - OS: macOS 14.0 
 - RimSort version: alpha v1.0.6.2-hf
 - Please provide a `RimSort.log`: none created
 - Are you running from Python interpreter or from a compiled RimSort release? RimSort release.

oceancabbage:
Looking into this now. Thinking of going with your suggestion 2 and 3: adding autodetect support for people who play the GoG version (I'm assuming the default path is how you have it, in `Applications`; and, allowing manual selection of folders or `.app`.

Not 100% sure what you mean by "remove default name appending in path", do you want to clarify on that?

Ant-V:
Sorry, I may have not been clear enough, because I had to post it twice because of a timeout error on githubs part.

When I enter a custom path directly in the settings, it appends the "RimWorldMac.app" in the path, so for example when I entered `"/Applications/RimWorld.app"` the error was `RimSort is unable to get the game version at the expected path: [/Applications/RimWorld.app/RimWorldMac.app/Version.txt]  Is your game path [/Applications/RimWorld.app/RimWorldMac.app] set correctly?`, which I (not so descriptively) would described as appending the default name in the path.

Maybe checking if Version.txt exists before appending the default `RimWorldMac.app` name in the path could also be a solution.

I could try a fix and posting a pull request, but sadly python is a language I almost never use.
If I have time later today however I may give it a go.

Ant-V:
[...]

Issue #158, opened 2023-09-01
Implement performance improvements


jefferyharrell:
**Describe the bug**
There's about a three-second delay between clicking the "Sort active mods" button and getting the results. This seems to be independent of the number of active mods.

**To Reproduce**
1. Clear the active mod list by clicking the "Clear active mods" button. Only Core and the DLCs (if installed) should be in the active mod list.
2. Click "Sort active mods."
3. Observe that it takes about three seconds (roughly) for the active mods list to populate.

**Expected behavior**
Sorting an essentially empty active mod list should take imperceptible time. Sorting a longer active mod list might take longer in proportion to the number of mods in the list, but an empty list should sort instantaneously.

**Please complete the following information:**
 - OS: Windows, macOS, Linux
 - RimSort version: 6bc03c2
 - Are you running from Python interpreter or from a compiled RimSort release? Compiled.

[RimSort.log](https://github.com/RimSort/RimSort/files/12491698/RimSort.log)

twstagg:
I have a series of performance improvements that I am working on planning that should help address this. Unfortunately, this is going to involve doing all of the things I have previously spoken to you about with the way XML is parsed and how information from the overall cache is used in relation to these activities. I will comment here once this has been addressed.

twstagg:
This should be significantly faster with the UUID changes I have made recently - closing issue.

PR #172, opened 2023-09-29
MacOS: Changed Settings: user now selects .app

Ant-V:
Fixes issue [#171](https://github.com/RimSort/RimSort/issues/171) by having users select their .app instead of asking for a folder and then assuming the app name, which causes crashes if the app was not named "RimWorldMac.app"

**May cause crashes to users on MacOS with old version settings**
If it does: delete ~/Library/Application Support/RimSort/settings.json or run:
`rm ~/Library/Application\ Support/RimSort/settings.json` in terminal

### DIFF
--- a/util/generic.py
+++ b/util/generic.py
@@ -78,9 +78,8 @@ def launch_game_process(game_install_path: str, args: list) -> None:
     if game_install_path:
         system_name = platform.system()
         if system_name == "Darwin":
-            executable_path = str(
-                Path(os.path.join(game_install_path, "RimWorldMac.app")).resolve()
-            )
+            # MacOS
+            executable_path = str(game_install_path)
         elif system_name == "Linux":
             # Linux
             executable_path = str(
--- a/util/mods.py
+++ b/util/mods.py
@@ -714,9 +714,6 @@ def get_game_version(game_path: str) -> str:
     """
     logger.info(f"Getting game version from Game Folder: {game_path}")
     version = ""
-    if platform.system() == "Darwin" and game_path:
-        game_path = str(Path(os.path.join(game_path, "RimWorldMac.app")).resolve())
-        logger.debug(f"Running on MacOS, generating new game path: {game_path}")
     version_file_path = str(Path(os.path.join(game_path, "Version.txt")).resolve())
     logger.debug(f"Generated Version.txt path: {version_file_path}")
     if os.path.exists(version_file_path):
@@ -790,11 +787,6 @@ def get_installed_expansions(game_path: str, game_version: str) -> Dict[str, Any
     mod_data = {}
     if game_path != "":
         logger.info(f"Getting installed expansions with game folder path: {game_path}")
-        # RimWorld folder on mac contains RimWorldMac.app which
-        # is actually a folder itself
-        if platform.system() == "Darwin" and game_path:
-            game_path = str(Path(os.path.join(game_path, "RimWorldMac.app")).resolve())
-            logger.info(f"Running on MacOS, generating new game path: {game_path}")
 
         # Get mod data
         data_path = str(Path(os.path.join(game_path, "Data")).resolve())
@@ -883,7 +875,7 @@ def get_local_mods(
             and local_path == game_path
         ):
             local_path = str(
-                Path(os.path.join(local_path, "RimWorldMac.app", "Mods")).resolve()
+                Path(os.path.join(local_path, "Mods")).resolve()
             )
             logger.info(
                 f"Running on MacOS, generating new local mods path: {local_path}"
--- a/util/mods.py
+++ b/util/mods.py
@@ -864,23 +864,6 @@ def get_local_mods(
         if game_path:
             logger.info(f"Supplementing call with game folder path: {game_path}")
 
-        # If local mods path is same as game path and we're running on a Mac,
-        # that means use the default local mods folder
-
-        system_name = platform.system()
-        if (
-            system_name == "Darwin"
-            and local_path
-            and game_path
-            and local_path == game_path
-        ):
-            local_path = str(
-                Path(os.path.join(local_path, "Mods")).resolve()
-            )
-            logger.info(
-                f"Running on MacOS, generating new local mods path: {local_path}"
-            )
-
         # Get mod data
         logger.info(f"Getting local mods from path: {local_path}")
         mod_data = parse_mod_data(local_path, "local", steam_db)
--- a/view/game_configuration_panel.py
+++ b/view/game_configuration_panel.py
@@ -1032,7 +1032,7 @@ class GameConfiguration(QObject):
         rimworld_mods_path = ""
         if system_name == "Darwin":
             rimworld_mods_path = str(
-                Path(os.path.join(os_paths[0], "RimWorldMac.app", "Mods")).resolve()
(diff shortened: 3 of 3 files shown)


twstagg (commented):
This looks good to me! I've asked @jefferyharrell to give it a whirl on his Mac that he has RimWorld installed on. Should be good to go once we get his confirmation as well.

jefferyharrell:
I've encountered some problems with this change.

Firstly, it does in fact hard-crash if a previously valid `settings.json` is present. It shows this GUI error:

image

and then bombs with this stack trace: https://gist.github.com/jefferyharrell/4bb2f09db0054a683afe009be9870006.

So okay, we delete `settings.json` and restart the app. We get this error dialog on startup, as expected:

image

If we use "autodetect paths" and hit "refresh mods," we get exactly the same crash as before, starting with this error dialog:

image

and ending with this stack trace: https://gist.github.com/jefferyharrell/522c2f2d88c165dd40c90f637479d841

If we then relaunch the app, we get the same error and hard crash, 'cause the autodetected paths got written to `settings.json`.

Delete `settings.json` again, and this time set "Game folder" and "Local Mods" both manually to `~/Library/Application Support/Steam/steamapps/common/RimWorld/RimWorldMac.app`. The other two paths can either be set manually (tedious) or via the autodetect button. Hit the refresh mods button and it works. Quitting and re-running the app works too, indicating that the paths are getting saved correctly.

So we have four (what I would consider) major issues that need addressing:

- The hard crash that can only be rectified by deleting `settings.json` is a showstopper in my opinion. If the app stayed up allowing the user to "clear cfg paths" through the settings dialog box, that'd be better, but better still would be to incorporate appropriate checks in the logic. Maybe check at startup if the path to Version.txt is valid and if it's not, null that path variable out and ask the user please to set it correctly.
- "Autodetect paths" is working in the latest builds, but is not working with this change. Using it results in a hard crash. I consider this a regression that should be fixed before merging.
- In the GUI the path fields are labeled "Game folder" and "Local mods," when on the Mac both of them are _really supposed to_ point to the RimWorld app package. This is at least confusing, and arguably a UI bug.
- There is no visual indication on macOS Ventura that the user should set the aforementioned paths to the app package. This is what the file-selection dialog looks like:

image

It should be noted that the `show_dialogue_file` function call's `caption` argument has no effect on the Mac; the file selection dialog has no title bar or any other window chrome.

So to sum up, I think the hard crash is a showstopper, autodetect paths is a regression, and at the very least the GUI needs some tweaking if we're going to go this way with the Mac version of the app. Until these things are addressed, I can't recommend merging this PR just yet.

Please don't take these comments as anything other than friendly, constructive feedback. I was smiling when I wrote this. 😄

Ant-V:
Thanks for taking the time and looking into it and do not worry, I do find your feedback extremely constructive, positive and to the point!
So, let me address those issues.

Regarding the crash: it is an issue that kept happening in previous versions as well, and is the fact why I decided to get into fixing the path, as I could not make it start in any way possible. I believe it is an issue with the main process, but so far have not looked really into it and as I do not really use python, cannot promise I can fix.

Regarding the auto-detect: it is a feature that does not work for me at all, as I do not use steam, so I could not test it. I see what is going wrong and will come back with a fix for that and maybe add some more detection for non-steam.

The issues regarding the GUI should be easy to fix and I agree 100%. I changed the captions mostly as a form of documentation.

jefferyharrell:
I've figured out how to consistently reproduce the hard crash we were talking about.

Start with a valid `settings.json` from a previous version of RimSort. I used 9464e56 to create mine.

Next, switch to @Ant-V's c1ecfcc. Launch RimSort, observe the error message…

image

…click "OK", then boom, crash.

This behavior is the same on Intel and Apple Silicon Macs, with the exception that on Intel the app crashes due to SEGV, while on Apple Silicon it crashes due to an uncaught `NSInternalInconsistencyException` with the messsage `NSWindow should only be instantiated on the main thread!`.

Attached are my `settings.json`s and my `RimSort.log`s from both my Macs.

[settings_arm64.json.txt](https://github.com/RimSort/RimSort/files/12784952/settings_arm64.json.txt)
[RimSort_arm64.log.txt](https://github.com/RimSort/RimSort/files/12784950/RimSort_arm64.log.txt)
[settings_x86.json.txt](https://github.com/RimSort/RimSort/files/12784953/settings_x86.json.txt)
[RimSort_x86.log.txt](https://github.com/RimSort/RimSort/files/12784951/RimSort_x86.log.txt)

jefferyharrell:
I've been doing more looking into the crasher. To review, here are the steps to reproduce that work for me:

1. Start with a good `settings.json` from the main repo version of the app.
2. Check out @Ant-V's main branch.
3. Do the necessary things (symlink steamworks, create and activate venv, etc.)
4. Run `python RimSort.py`
5. Observe the "Issue Getting Game Version" warning dialog.
6. Observe the crash.

Attached is a copy of the complete stack trace I get when the app crashes. If you'll notice, the actual uncaught exception is `NSInternalInconsistencyException` with the error `NSWindow should only be instantiated on the main thread!`.

[stacktrace.txt](https://github.com/RimSort/RimSort/files/12795779/stacktrace.txt)

I've managed to track this exception down to `model/animations.py` line 77, `self.thread.start()`. As soon as that line gets hit, the crash happens.

Now, I'm still too rusty in Python or QT to understand where in this thread start call the exception is coming from, but it _looks_ like something in the thread is trying to instantiate an NSWindow?

Some insight from @twstagg would be greatly appreciated at this point. I feel like I've gone about as far as I can with what I have right now.

jefferyharrell:
> You can reproduce the crash without the patch. All you have to do is set the game path to something that is not the actual game path.

Yeah, confirmed, but in a subtle kind of way. If you point `game_folder` to a path that _does not exist,_ the app handles it correctly in my opinion: It nulls out the text input in the UI and throws up an "essential paths not set" warning as seen here:

image

But if you set it to a path that _exists_ but that points to a file, or that points to a directory where there's no `Version.txt`, then you get the crash. Something to do with failing to find `Version.txt` leads to the crash, I believe.

Ant-V:
Yes, if you set the path from the GUI, it checks that there is no such path and reverts to nothing.
If you change it through the settings, or simpler yet, rename something along the actual path, you get a crash.

I did run the whole trace, it leads to (util/mods.py#804)[https://github.com/RimSort/RimSort/blob/9464e5685d935f42e98c54565493d38f144dff2f/util/mods.py#L804C58-L804C58]. Try just renaming version.txt and the app will go through normally (minus all the warnings about incompatible mod and game versions).
The version alert is the last thing that happens before crashing.

Ant-V:
So, the true trigger of the crash is [util/mods.py#1351](https://github.com/RimSort/RimSort/blob/9464e5685d935f42e98c54565493d38f144dff2f/util/mods.py#L1351). 
For a reason I cannot (yet) understand it cannot show the window, although it does log all the window popping up.

I need a better understanding of python and why it is, and I quote:
`Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow should only be instantiated on the main thread!'`

twstagg:
Thanks guys. I will resolve the issues with the unsafe thread operations with my efforts in #158

twstagg:
Please reopen PR with branch from origin master

PR #175, opened 2023-10-03
MacOS: Changed Settings: user now selects .app

Ant-V:
Fixes [issue #171](https://github.com/RimSort/RimSort/issues/171) by having users select their .app instead of asking for a folder and then assuming the app name, which causes crashes if the app was not named "RimWorldMac.app"

##May cause crashes to users on MacOS with old version settings##
If it does: delete `~/Library/Application Support/RimSort/settings.json` or run:
`rm ~/Library/Application\ Support/RimSort/settings.json in terminal`


Ant-V:
Closing to edit and merge directly with `metadata-manager` branch

PR #176, opened 2023-10-06
Add icon caching to speed up drawing mod lists

jefferyharrell:
I did some profiling on the app this morning and noticed that `ModListItemInner.get_icon` runs for about 900 milliseconds during app startup.

image

I looked at the code and saw that `ModListItemInner.get_icon` calls `PySide6.QtGui.QIcon` to load an icon from a file on disk once for every item in the mod lists.

So I added a static class called `ModListIcons`, the sole purpose of which is to cache mod list icons. It has class methods that instantiate `QIcon`s for each icon and cache them for the duration of the program.

I also changed `get_icon` to use the new caching static methods.

The result was reducing startup time by about a second with my mod list. The performance improvement will be proportional to how many inactive and active mods the user has.

image

Note that this also speeds up mod list refresh by roughly the same amount. It's actually _slightly_ more beneficial on refresh than on startup because the icons are already cached.


twstagg (approved):
Great work!

PR #178, opened 2023-10-08
macOS: Game App button now opens containing folder

Ant-V:
Also included some compatibility to keep user selecting .app


Ant-V:
Its a mess, screw it, I'll do another push to head to fix all the conflicts

PR #192, opened 2023-10-13
Beautify update.sh

tuxiqae:


Chunnyluny:
Please provide information why you have done these changes, that will make it easier to understand for others what you have done and with what reasons. Are you in the RimSort discord ?

jefferyharrell:
I get the following error in Bash 3.2 on macOS 14.

```sh
chmod +x ./update.sh && ./update.sh
./update.sh: line 9: Darwin: unbound variable
```

This appears to be because you (wisely) called `set -u`, but it's barfing on the string literals.

tuxiqae:
> I get the following error in Bash 3.2 on macOS 14.
> 
> ```shell
> chmod +x ./update.sh && ./update.sh
> ./update.sh: line 9: Darwin: unbound variable
> ```
> 
> This appears to be because you (wisely) called `set -u`, but it's barfing on the string literals.

I think that it's due to the old version of bash on macOS, let me see if I can somehow replicate it

jefferyharrell:
> I think that maybe associative arrays might not be supported on this old version of bash on macOS, let me see if I can somehow replicate it

Yup, I can confirm the problem does not happen under Bash 5.2 on the same computer. But 3.2 is what comes with the OS, so we're kind of stuck there.

tuxiqae:
> > I think that maybe associative arrays might not be supported on this old version of bash on macOS, let me see if I can somehow replicate it
> 
> Yup, I can confirm the problem does not happen under Bash 5.2 on the same computer. But 3.2 is what comes with the OS, so we're kind of stuck there.

Yeah, I know that, fortunately I was able to run it on my PC using the Docker image `bash:3.2` in order to reproduce the issue.

tuxiqae:
@jefferyharrell Replaced the associative array with a normal array, seems to be fixed in my Docker container, let me know if it works :)

update.sh
```diff
@@ -1,39 +1,60 @@
...
+}
+
+for supported_os in "${SUPPORTED_OSES[@]}"; do
+    if [[ "$supported_os" == "$os" ]]; then
+        os_is_supported=true
+    fi
+done
+
+if [[ ! $os_is_supported ]]; then
+    echo "The operating system in use \"${os}\" is unsupported, please use once of the following ($(pretty_print_array_values "${SUPPORTED_OSES[@]}"))"
```
tuxiqae, line 23: Currently looks like that:
```
The operating system in use Linux is unsupported, please use once of the following (Darwin,Linux)
```
Let me know if you would like me to replace `Darwin` with `macOS`, I thought about doing that but it'll force me to use a solution which won't look as tidy as the one that is implemented

jefferyharrell: Speaking as one of two resident Mac guys, the name "Darwin" doesn't mean anything to most people. It should be macOS. But that's just my take. I'd like to know what @Ant-V thinks as well.

As for the other part, can we get away with just saying "Linux"? I'm _not_ a resident Linux guy, but aren't there several Linux-based operating systems out there? Just a question.

tuxiqae: > Speaking as one of two resident Mac guys, the name "Darwin" doesn't mean anything to most people. It should be macOS. But that's just my take. I'd like to know what @Ant-V thinks as well.
> 
> As for the other part, can we get away with just saying "Linux"? I'm _not_ a resident Linux guy, but aren't there several Linux-based operating systems out there? Just a question.

Regarding macOS, I'll let you guys decide
Regarding Linux, there are a ton of different distros, but currently we don't distinguish between them, and support them all.

jefferyharrell (approved):
Looks good to me, but I'd like to reiterate my suggestion that we get rid of `update.sh` entirely. Seeing `rm -rf` in a shell script makes me very uncomfortable. I think when a RimSort update is detected we should just pass the download URL to the user's default browser for it to download the file.

twstagg:
@tuxiqae 

I checked out on your `beautify_update.sh` branch and did the following:
```
(RimSort) tstagg@TSTAGG-U:~/src/RimSort$ mkdir ../RimSort_testing
(RimSort) tstagg@TSTAGG-U:~/src/RimSort$ cp update.sh  ..//RimSort_testing/
(RimSort) tstagg@TSTAGG-U:~/src/RimSort$ cd ../RimSort_testing/
(RimSort) tstagg@TSTAGG-U:~/src/RimSort_testing$ cp -r ../RimSort/dist/RimSort.dist/ /tmp/RimSort
(RimSort) tstagg@TSTAGG-U:~/src/RimSort_testing$ ./update.sh 
Updating RimSort in 5 seconds. Press Enter to continue.(RimSort) tstagg@TSTAGG-U:~/src/RimSort_testing$ 
```
The behavior is still disabled.. I understand we talked about contextual issues where bad things could happen if someone ran it out of context. What if modify it to take commandline arguments directly from the app instead? All it really needs is source dir (which is always the os tmpdir that RimSort downloads and extracts the update to, and the destination dir (which should resolve to the directory that the frozen application is running from.

@jefferyharrell 

> Looks good to me, but I'd like to reiterate my suggestion that we get rid of `update.sh` entirely. Seeing `rm -rf` in a shell script makes me very uncomfortable. I think when a RimSort update is detected we should just pass the download URL to the user's default browser for it to download the file.

I don't think this is done yet, so I am not approving at this time. I personally think that the app should be able to update itself. Apologies for the size of the MacOS app, PySide6 is indeed gigantic. More discussion can be had on a more safe approach to updating the app, as to avoid mistakes.

twstagg (changes requested):
Could you change it to take arguments for update source/destination paths? I can add the change in behavior to RimSort in the python end of things if so.

PR #206, opened 2023-10-31
Clean up RimSort.py

jefferyharrell:
- added an app_data_folder property to AppInfo that points to the `data` folder location
- refactored `MainWindow` into its own Python module and out of `RimSort.py`
- renamed some previously private methods in `MainWindow` to public methods because they're being called outside the class
- `RimSort.py` and `main_window.py` now both pass mypy, which the exception of `logger_tt` lacking type stubs

Tested on macOS 14 and Windows 11.

### DIFF
--- a/RimSort.py
+++ b/RimSort.py
@@ -1,40 +1,32 @@
 import sys
 
-from multiprocessing import current_process, freeze_support, set_start_method
+from multiprocessing import freeze_support, set_start_method
 
 import os
 from pathlib import Path
 import platform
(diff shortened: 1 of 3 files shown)


twstagg (changes requested):
Left you some comments in the Discord #developers

- Please restore the conditional import based on platform for PollingObserver on Windows, and Observer only on unix for now.
- Import sorting nitpicks. Great work on minimizing the broad imports we had before!

twstagg (approved):

PR #214, opened 2023-11-05
Add "Open…" buttons to Locations tab

jefferyharrell:
Per user request, the Locations tab now has "Open…" buttons which open their corresponding locations in the operating system's file manager.

image


twstagg (approved):
Thank you!

PR #224, opened 2023-11-24
Refactored application source code into a top-level RimSort package

jefferyharrell:
The structure of the application's source code has been preserved; if a class was previously named `foo.bar.MyClass` then that class is now named `RimSort.foo.bar.MyClass`. Note that this also preserves some idiosyncrasies in the code that probably want attention in the future, such as the `ImageLabel` class being located at `RimSort.models.image_label`.

### Why?

1. To clean up the source code overall and improve project organization
2. To make it possible to conveniently use documentation tools like `pyreverse`


twstagg (approved):

twstagg:
Verified w/ [Edge c781e46](https://github.com/RimSort/RimSort/releases/tag/Edge)

PR #232, opened 2023-11-26
Unify active/inactive panels, get rid of sub_views

twstagg:
Also fixes issues with

- delete_files_except_extension
- An issue on MacOS, where using debug mode caused issue with app bundle because i put it in the `AppInfo.application_folder` mistakenly 
- Steamworks issue stemming from libs path being passed from subprocesses, which were attempting to access `AppInfo` from another process


LionelColaso (approved):

twstagg:
> `app/views/mods_panel.py` is currently kind of a sprawling megafile with several different classes in it. I suggest refactoring those classes out into their own files. It'll make the code much easier to read and understand, making onboarding new developers a lot simpler.

`apps/views/mods_panel.py` provides a single view, consisting of several custom widgets that said view is composed of. It makes sense for all of this to be in a single file.

Keeping tightly-coupled things together is sensible practice. Avoiding inappropriate coupling is also good practice. Striking the right balance isn't a matter of strict rules, but of, well, striking a balance between different concerns. Yes, it is larger sized file now, but if you actually troll through the code, I think that you will see why it makes sense to keep it all together here. The things I moved out of `util.models` alone was a good justification i think, and also, this file encapsulates everything that is behind the "mods panel" view.

PR #233, opened 2023-11-26
Cleanup rentry

LionelColaso:


app/utils/rentry/wrapper.py
```diff
@@ -1,93 +1,98 @@
...
-            client.post("https://rentry.co/api/new", payload, headers=_headers).data
-        )
+        # Perform the POST request to create a new entry
+        return json_loads(client.post(API_NEW_ENDPOINT, payload, headers=_HEADERS).data)
 
 
 class RentryImport(QDialog):
     def __init__(self):
         super().__init__()
-        self.package_ids: list[str] = []  # Initialize an empty list to store package_ids
```
twstagg, line 87: Why remove typing?

app/utils/rentry/wrapper.py
```diff
@@ -1,93 +1,98 @@
...
-                logger.error("RentryUpload failed!")
-            except:
-                logger.error("RentryUpload failed!")
-        else:
-            self.upload_success = True
-            self.url = response["url"]
-            logger.debug("RentryUpload successfully uploaded data!")
-            logger.debug("Url: {}".format(self.url))
-            logger.debug("Edit code: {}".format(response["edit_code"]))
+        self.upload_success = False
```
twstagg, line 48: Nice work here. Much simpler, fewer dependencies.

app/utils/rentry/wrapper.py
```diff
@@ -118,11 +124,11 @@ def import_rentry_link(self):
             return
 
         try:
-            raw_url = rentry_link + "/raw"
-            response = urllib.request.urlopen(raw_url)
+            raw_url = f"{rentry_link}/raw"
+            response = requests.get(raw_url)
 
-            if response.getcode() == 200:
-                page_content = response.read().decode("utf-8")
```
twstagg, line 125: Why remove this? Particularly, `decode("utf-8")` could be useful

twstagg (changes requested):

twstagg (changes requested):
Additionally, I'm getting the following issue when I try to import:

`[INFO][2023-11-26 19:45:09][1140747][MainThread][dialogue][show_fatal_error][269] : Showing fatal error box with input: [RimSort crashed], [The RimSort application crashed! Sorry for the inconvenience!], [Please contact us on the Discord/Github to report the issue.], [Traceback (most recent call last):
  File "/home/tristin/src/RimSort/app/views/main_content_panel.py", line 1420, in _do_import_list_rentry
    self.progress_dialog.close()
    ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MainContent' object has no attribute 'progress_dialog'
]`

app/utils/rentry/wrapper.py
LionelColaso: got rid of list[str] being and extra step and importing modlist without it works as well can add it back if its
so now its
self.package_ids = []
before it was 
self.package_ids: list[str] = []

LionelColaso:
> Additionally, I'm getting the following issue when I try to import:
> 
> `[INFO][2023-11-26 19:45:09][1140747][MainThread][dialogue][show_fatal_error][269] : Showing fatal error box with input: [RimSort crashed], [The RimSort application crashed! Sorry for the inconvenience!], [Please contact us on the Discord/Github to report the issue.], [Traceback (most recent call last): File "/home/tristin/src/RimSort/app/views/main_content_panel.py", line 1420, in _do_import_list_rentry self.progress_dialog.close() ^^^^^^^^^^^^^^^^^^^^ AttributeError: 'MainContent' object has no attribute 'progress_dialog' ]`

fixed had missed that somehow

app/utils/rentry/wrapper.py
LionelColaso: i added it back now

twstagg: Noice

twstagg (approved):

app/utils/rentry/wrapper.py
```diff
@@ -1,85 +1,88 @@
...
-        response.status_code = response.getcode()
-        response.data = response.read().decode("utf-8")
+        # Initialize a session for making HTTP requests
+        self.session = requests.Session()
+
+    def get(self, url, headers=None):
+        # Perform a GET request and return the response
+        headers = headers or {}
+        response = self.session.get(url, headers=headers)
+        response.status_code = response.status_code
```
jefferyharrell, line 29: This line appears to do nothing.

app/utils/rentry/wrapper.py
```diff
@@ -1,85 +1,88 @@
...
+        response.data = response.text
         return response
 
+    def post(self, url, data=None, headers=None):
+        # Perform a POST request and return the response
+        headers = headers or {}
+        response = self.session.post(url, data=data, headers=headers)
+        response.status_code = response.status_code
```
jefferyharrell, line 37: This line also appears to do nothing.

app/utils/rentry/wrapper.py
```diff
@@ -1,85 +1,88 @@
...
 from loguru import logger
 
-_headers = {"Referer": "https://rentry.co"}
+# Constants
+BASE_URL = "https://rentry.co"
+API_NEW_ENDPOINT = f"{BASE_URL}/api/new"
 
+_HEADERS = {"Referer": BASE_URL}
 
 class UrllibClient:
```
jefferyharrell, line 20: I suggest renaming this class to something more descriptive, since `urllib` is not actually being used any more.

jefferyharrell (changes requested):
Generally looks good, I suggest some small changes and some larger-scale refactoring to improve overall.

app/utils/rentry/wrapper.py
LionelColaso: Fixed

app/utils/rentry/wrapper.py
LionelColaso: Fixed

app/utils/rentry/wrapper.py
LionelColaso: Fixed

app/utils/rentry/wrapper.py
```diff
@@ -1,85 +1,88 @@
...
         }
 
-        return json_loads(
-            client.post("https://rentry.co/api/new", payload, headers=_headers).data
-        )
-
+        # Perform the POST request to create a new entry
+        return json_loads(client.post(API_NEW_ENDPOINT, payload, headers=_HEADERS).data)
 
 class RentryImport(QDialog):
```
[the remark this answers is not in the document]
LionelColaso: why ? right now its just a basic QDialog
from PySide6.QtWidgets import QDialog

app/utils/rentry/wrapper.py
```diff
@@ -131,15 +137,11 @@ def import_rentry_link(self):
...
-            )
+            logger.error(f"An error occurred while fetching rentry.co content: {str(e)}")
 
-        # Close the dialog after processing
+        # Close the dialog after processing the link
         self.accept()
 
-
 if __name__ == "__main__":
     sys.exit()
```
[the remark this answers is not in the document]
LionelColaso: ?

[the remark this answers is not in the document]
jefferyharrell: These lines don't do anything, and they suggest that this file should be run as a script, which should never happen.

app/utils/rentry/wrapper.py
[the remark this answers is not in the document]
jefferyharrell: There are two principles of object-oriented programming that we need to take into consideration in this situation.

1. The single-responsibility principle.
2. Composition versus inheritance.

The single-responsibility principle, which is fundamental to object-oriented programming, says that classes should have just one job or area of responsibility. Here you have a class that is responsible for a GUI element (the dialog box) and that is also responsible for talking to the Rentry web site, downloading content and parsing it.

This class should be separated into two classes: One class that is responsible for downloading from the Rentry site and parsing the data, and a second class that is responsible for the GUI element.

As for composition versus inheritance, we usually talk in terms of _has a_ versus _is a._ Inheritance is an _is a_ relationship: Apple _is a_ Fruit; Employee _is a_ Person. Where an _is a_ relationship exists, it's correct to use inheritance.

Composition represents a _has a_ relationship. It would be a programming error for the Employee class to inherit from the EmailAddress class, because an employee _is not an_ email address. Instead an employee _has an_ email address, so the Employee class should contain a reference to an EmailAddress object as an attribute.

Here you're inheriting from QDialog, saying that this class _is a_ QDialog. But it isn't. It downloads content from Rentry; that's its job. This class can _have a_ QDialog (or more properly, a QDialog subclass) but it's wrong to say it _is a_ QDialog. That's a programming error.

app/utils/rentry/wrapper.py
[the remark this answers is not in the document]
LionelColaso: There is a really nice use case for the __name__ variable, whether you want a file that can be run as the main program or imported by other modules. We can use an if __name__ == "__main__" block to allow or prevent parts of code from being run when the modules are imported.

When the Python interpreter reads a file, the __name__ variable is set as __main__ if the module being run, or as the module's name if it is imported. Reading the file executes all top level code, but not functions and classes (since they will only get imported).

https://www.freecodecamp.org/news/if-name-main-python-example/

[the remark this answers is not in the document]
jefferyharrell: Heh. Thank you, I'm familiar with the `__name__ == __main__` idiom. I'm saying it shouldn't be in this file. This file should never be run as a script, so we don't need a `__name__ == __main__` block, especially not one that does nothing.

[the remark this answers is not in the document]
LionelColaso: thats right i suppose i can remove it not sure if it was a failsafe implemented by twsta
fixed

app/utils/rentry/wrapper.py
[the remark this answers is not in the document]
LionelColaso: yeah i see your point but unless there are more places that we would be needing QDialog entirely as a class i dont want to make it so maybe you can do it after the merge and show me what you mean i am happy to learn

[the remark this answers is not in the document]
jefferyharrell: Yeah, all right.

Issue #142, opened 2023-08-19
Implement ability to add a Steam collection to active, unless collection is not downloaded


jefferyharrell:
An action needs implemented to do the following:

- [x] Ensure that this option is apart of the "Import mod list" submenu
- [x] Assemble a list of mods that are only from a respective collection (publishedfileids should be parsed for Workshop mods in internal metadata)
- [x] Force insert them all into the active mod list

Original post:

I saw this on a Discord this morning:

<img width="735" alt="image" src="https://github.com/RimSort/RimSort/assets/8659114/738dfb83-659a-43ba-8ae8-2277186ffd0b">

Made me wonder if Steam collections metadata is exposed via API anywhere, and if RimSort could access it, either to provide information in the GUI or _maaaaaybe_ to provide enable/disable collection functionality for those people who are into that sort of thing.

This is, I think, a very obscure use case, so I'm just asking the question to ask.

sparr:
https://steamapi.xpaw.me/#ISteamRemoteStorage/GetCollectionDetails

This API call will tell you all the mods in a collection

twstagg:
@jefferyharrell @sparr This is an interesting idea. We already have the ability to download Steam collections via https://github.com/RimSort/RimSort/blob/main/util/steam/webapi/wrapper.py#L485

We could probably implement a mechanism that would allow the user to "import a mod list from a Steam Workshop collection" but since the defacto sorting of most RimWorld workshop collections is pretty bunk, we probably don't want to rely on that default order.

However, that doesn't mean that we can't assemble a list of mods that are only from that collection, force insert them all into the active mod list, and then trigger a sort.

LionelColaso:
Implemented

Issue #26, opened 2023-02-21
Implement "Import save" feature


twstagg:
This would allow us to import a modlist from the data parsed from a Rimworld save

twstagg:
This has been completed.

PR #236, opened 2023-12-03
Add ability to import mod list from .rws files (RimWorld save)

twstagg:
* Also implements xmltodict as primary xml parser, with bs4 as a fallback for trickier chars/encodings


twstagg:
Fixes #26

twstagg:
Fixes #142

Issue #137, opened 2023-08-14
Implement a way to filter out/highlight version incompatible mods


glencoe2004:
**Describe the solution you'd like**
While the current system of throwing a "Mod and Game Version Mismatch" warning works, it is too tedious for assembling large mod lists. A much better system would be something like RimPy's highlighting version incompatible mods in red, or even better, allowing the Inactive list to be filtered to remove old version mods.

Chunnyluny:
I understand the situation better now. Here's a suggestion for your response:

"I completely understand your concern. Dealing with a 'Mod and Game Version Mismatch' warning for extensive mod lists can indeed be tedious. Your suggestion to enhance the system sounds very practical.

Highlighting version-incompatible mods in red, similar to how RimPy does it, would be an excellent approach. This visual cue would make it much easier to identify outdated mods at a glance. Additionally, I see the value in allowing users to filter the 'Inactive' list to remove old version mods directly. This would save a significant amount of time when managing large mod collections.

I appreciate your input, and I believe these improvements would not only enhance the user experience but also streamline the process of assembling mod lists. This could be a valuable addition for users who frequently manage extensive mod libraries."

lisanhu:
I have to say this is more useful at this time when the 1.5 is rolling out! I wish to see some parts of my installed mods turn white again (while the rest remains red)!

lisanhu:
I have just implemented this functionality. Any suggestions before it's merged?
![image](https://github.com/RimSort/RimSort/assets/3086096/b4ea484b-c462-41b1-a2fd-7b094cc0f6e6)

Daenyth:
I see there's been merges but is this released? It doesn't seem to be working in alpha-v1.0.6.2-hf. I don't get red highlights (darwin-arm release)

Daenyth:
I tried that first but when I tried to use the darwin-arm release it told me it that it didn't work

> You can’t use this version of the application “RimSort” with this version of macOS.

I'm on an M2 mac

LionelColaso:
Reopen issue if still exists on current RimSort

PR #275, opened 2024-03-15
Merging temp branch to main

lisanhu:
This is a temp branch I created to merge my fork for #137 
image

I just realized I don't need to create a seperate branch for this.

### DIFF
--- a/app/views/mods_panel.py
+++ b/app/views/mods_panel.py
@@ -335,15 +335,21 @@ class ModListItemInner(QWidget):
 
         # Adding checks for game version and mod supported game versions
         # Use red to highlight the mod that is not compatible with the game version
-        # TODO
         game_version = self.metadata_manager.game_version
         uuid = self.uuid
-        data_source = self.metadata_manager.internal_local_metadata[uuid].get(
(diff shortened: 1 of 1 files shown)


lisanhu:
I'm having a random crashing issue running this. 

Sorry I lost the bug report, but it's happening at line mods_panel.py:320, that after running the program for hours, the uuid might not in the metadata database. I've created a fix for that, still testing whether it's working.

lisanhu:
I might figure out what's causing the crash (not 100% sure).

After refreshing and upgrading some of my mods, I realized the upgraded mods were marked red (when they are supporting 1.5 and should be white). I checked steamworkshop and I'm sure those were supporting 1.5. I refreshed again and the mods were colored to white (which we expected).

So probably there's something wrong with the metadata after some mods updates? I'm not sure how metadata works in RimSort, so I'm just trying to give some clues.

The code didn't ever crash for me since I updated my branch (with a check whether uuid is in the metadata), so I'll push that commit to this temp branch (also probably rename this to make the branch name more meaningful).

LionelColaso (approved):

PR #284, opened 2024-04-11
[Hotfix] RimSort is not starting on MacOS

oceancabbage:
RimSort on MacOS is not starting because `upload_rimsort_old_log_action` is not properly instantiated upon MenuBar creation for MacOS.


LionelColaso (approved):
Looks good

PR #285, opened 2024-04-12
[QOL] Improving functionality clicking on mods

oceancabbage:
There is currently an edge case where if you click back and forth between the same two mods, one on the inactive mods panel and one on the active mods panel, the ModInfo panel does not update accordingly. This is because we're currently using the `currentItemChanged` hook and technically the item does not change in the context of a single mod panel. So, I'm adding the `itemClicked` hook to account for this edge case.

Technically, this means when you click an item the ModInfo panel is actually updated twice, but the work done during a single update is minimal so I think this is worth the QoL improvement.


twstagg (approved):
I pulled it down, tested it out. Cool!

PR #287, opened 2024-04-13
Minor Fixes

LionelColaso:


twstagg:
@LionelColaso Can you add a description detailing your changes? I don't see any issue after looking at the code, just wasn't clear what all is being done here and why

LionelColaso:
> @LionelColaso Can you add a description detailing your changes? I don't see any issue after looking at the code, just wasn't clear what all is being done here and why

Why is this an issue?
It is needlessly complex to invert the result of a boolean comparison. The opposite comparison should be made instead.

Noncompliant code example
if not a == 2:        # Noncompliant
    b = not i < 10    # Noncompliant

Compliant solution    
   if a != 2 :
    b = i >= 10

twstagg (approved):
Please squash into single commit

PR #290, opened 2024-04-14
Implement lazy loading for `ModListItemInner`

twstagg:
This introduces a functional "lazy-loading" for `ModListItemInner`

While unmeasured by me, you can observe a significant improvement in performance inside the app. Item widgets will not be created unless their parent items are visible.

Additional support has been added for search filter support as well


LionelColaso (approved):
Go Ahead

PR #291, opened 2024-04-14
display mod  SupportedVersions

LionelColaso:


twstagg (changes requested):
There are issues with this PR.

We need to properly parse + format in the tool tip, and you should also check how you are parsing the Expansions because that does not look correct in the info pane.

The value appears correct for actual mods in the info pane though.

LionelColaso:
> There are issues with this PR.
> 
> We need to properly parse + format in the tool tip, and you should also check how you are parsing the Expansions because that does not look correct in the info pane.
> 
> The value appears correct for actual mods in the info pane though.

Fixed 
image

app/views/mods_panel.py
```diff
@@ -272,23 +272,46 @@ def get_tool_tip_text(self) -> str:
...
+            ", ".join(supported_versions_list)
+            if isinstance(supported_versions_list, list)
+            else supported_versions_list or "Not specified"
+        )
+        supported_versions_line = f"Supported Versions: {supported_versions_text}\n"
+
+        path = metadata.get("path", "Not specified")
+        path_line = f"Path: {path}"
+
+        return "\n".join(
```
twstagg, line 305: This produces a different formatting than previous behavior (everything has an additional new line)

This effectively doubles the tooltip length and i am personally not a fan. Could we perhaps consider using an empty `""`?

twstagg (changes requested):
Everything else looks good though, good job!

LionelColaso:
> Everything else looks good though, good job!

It looks congested and cramped, without it, see screenshot below
so if u want i will use the one without it ?


image

LionelColaso:
done happy :-P

twstagg (approved):

PR #321, opened 2024-05-07
Update requirements.txt

LionelColaso:


LionelColaso:
builds and works on windows

twstagg (approved):
[10:38 PM]oceancabbage: The requirement update seems fine to me

Works on MacOS as well. I am able to run everything ok from interpreter on Linux with Python 3.12 as well.

PR #322, opened 2024-05-07
 Fixes

LionelColaso:


twstagg (changes requested):
Thank you for identifying these leftover things that still needed cleaned up. The only issue I have with this is that the purpose of the lines that you are removing were to close the settings dialog if it was open, before opening the rule editor, db builder, etc. This is because these options were able to be launched from the settings dialog.

Could you please add functionality here to replace what you are removing, so that maybe there is a signal that can be emitted when the settings dialog needs to be ensured that it is closed?

The reason for this is that the settings dialog has modality, at least last I checked, which means that it tries to stay in front of any open RimSort windows.

app/views/main_content_panel.py
```diff
@@ -2530,8 +2530,6 @@ def _do_notify_no_git(self) -> None:
     def _do_open_rule_editor(
         self, compact: bool, initial_mode=str, packageid=None
     ) -> None:
-        if GameConfiguration.instance().settings_panel.isVisible():
```
twstagg, line 2533: The new settings panel needs to close if it is visible here

LionelColaso:
1) settings_dialog is not an instance do u really want me to make it an instance ?
2) rule editor is not in settings anymore 
3) currently any action you do in settings it closes the itself
These are the reasons that i feel that there is no need to add the checks to see if settings has closed

twstagg:
> 1) settings_dialog is not an instance do u really want me to make it an instance ?
> 2) rule editor is not in settings anymore 
> 3) currently any action you do in settings it closes the itself
> These are the reasons that i feel that there is no need to add the checks to see if settings has closed

1) If you are asking whether or not you need to make the settings dialog a singleton, the answer is no. You can very easily implement what i am asking by using a signal/slot.

2) I was not aware of this. Rule editor should be in settings, but it should be configured to open right in to Rule Editor with Community Rules open, and no specific mod selected. This is another thing that @jefferyharrell seemingly removed from the app entirely without any explanation. 

3) If this one is true, then maybe isnt as important to handle it the way i am asking. 

However, i would like to test and confirm that the behavior is there one way or another. #2 is the more important thing to me as well, because a feature i added was removed for no reason. I can dig up the old code from the old settings panel and restore it.

LionelColaso:
> > 1. settings_dialog is not an instance do u really want me to make it an instance ?
> > 2. rule editor is not in settings anymore
> > 3. currently any action you do in settings it closes the itself
> >    These are the reasons that i feel that there is no need to add the checks to see if settings has closed
> 
>     1. If you are asking whether or not you need to make the settings dialog a singleton, the answer is no. You can very easily implement what i am asking by using a signal/slot.
> 
>     2. I was not aware of this. Rule editor should be in settings, but it should be configured to open right in to Rule Editor with Community Rules open, and no specific mod selected. This is another thing that @jefferyharrell seemingly removed from the app entirely without any explanation.
> 
>     3. If this one is true, then maybe isnt as important to handle it the way i am asking.
> 
> 
> However, i would like to test and confirm that the behavior is there one way or another. #2 is the more important thing to me as well, because a feature i added was removed for no reason. I can dig up the old code from the old settings panel and restore it.

its already does that from settings_controller for every EventBus() Signal
image

LionelColaso (commented):
ready for merge

twstagg (approved):

Issue #51, opened 2023-03-26
FIXME: All mods dissapear when clicking the "sort" button multiple times


Thekiborg:
OS: Windows 10 Pro 10.0.19045

1. Click the sort button multiple times, in my case, 3 times is enough for all the lists to dissapear
2. After doing that, no mods will be shown in either the inactive or the active lists, and the official expansions won't appear either
![image](https://user-images.githubusercontent.com/91976529/227784116-67f94ab2-8e15-41c3-9a22-441c0b4a23f7.png)
3. Clicking any button that updates the list, such as sort, clear, restore, etc, will make the mods visible again

twstagg:
@oceancabbage i can reproduce on Linux

largemoron:
I also got this while the list was refreshing after deleting local and empty folder UNKNOWN mods.

twstagg:
@largemoron shouldn't happen anymore now that i got rid of delayed refresh thread. The issue with the mods disappearing is still there though

Limfjorden:
Is this still present, or should we close this issue?

GitWolf22k:
Able to reproduce in latest edge (after about 10 clicks or so). However only my active list cleared.

PR #333, opened 2024-05-13
fix issue  #51

LionelColaso:


twstagg (approved):
Nice, this works for me! #51 and other side effects happen on main, but not with this branch. Great work!

PR #336, opened 2024-05-14
Fix object has no attribute 'pos'

LionelColaso:
  File "E:\Github\RimSort_Lion\build\__main__.dist\app\views\mods_panel.py", line 994, in eventFilter
AttributeError: 'PySide6.QtGui.QInputEvent' object has no attribute 'pos'


twstagg (approved):
Took me a second to realize what this did. Good catch! This definitely looks more correct than it was initially after reading QEvent docs.

Issue #339, opened 2024-05-16
Suggestion: General UI Readability Improvements


hk21702:
**Is your feature request related to a problem? Please describe.**
There are some significant shortcomings in terms of readability in certain parts of the app that manifest at least in windows (and I presume in other os's since it relates to the pyside stylesheet mostly). 

- [ ] 1. Radio button text is currently dark on dark.  (No styling exists for radio buttons currently)
- [ ] 2. Some items such as buttons have padding issues where the text is very close to the left and right edges (There's only 1px of padding around buttons. It really should be more. As an example, material design common buttons have left right padding of 24px, though even a padding of just 5px would be better.)
- [ ] 3. The menu bar of the app in terms of styling is rather inconsistent with the rest of the app on windows. I don't think any styling exists for it. My understanding is that it defaults to system default for mac/linux, but not on windows which is rather ugly without styling.

**Describe the solution you'd like**
I'm willing to help make these changes in a pull request, I'm just opening this issue first for purposes of discussion.

PR #341, opened 2024-05-16
UI Tweaks #339

hk21702:
Some UI tweaks related to #339

Increased padding for combo boxes.
Added styling for radio buttons for consistency with the rest of the program.
Added styling for menubar for consistency with the rest of the program on Windows. 

As a side note, should seriously consider moving away from pure white for font colors. There are guidelines for having too low contrast, but too high contrast can cause accessibility issues for some users, particularly for astigmatism and dyslexia. It's the main reason why some people use light mode when dark mode exists. I also personally find that the current colors show up kind of pixelated with such high contrast and small font size. As an example, Google uses #fafafa and GitHub uses #e6edf3.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,4 +17,5 @@
     "python.terminal.activateEnvInCurrentTerminal": true,
     "python.languageServer": "Jedi",
     "mypy.runUsingActiveInterpreter": true,
+    "autoDocstring.docstringFormat": "sphinx"
 }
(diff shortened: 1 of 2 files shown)


hk21702:
image
image
image

twstagg:
I like it! Style away!

A ThemeManager will be built eventually as well.

twstagg (approved):

hk21702:
Before merging, what are thoughts on changing the text color to be a slight off-white? I haven't included it in the PR but can do so quickly and provide screenshots. Goal of the change is to maintain readability, but make it a little easier on the eyes.

hk21702:
@twstagg Should be good to merge once you look at the final commit.

twstagg (approved):
I like all of this. Good work! Adding @LionelColaso for review as well.

LionelColaso (approved):
Buttons are much  better now

PR #343, opened 2024-05-17
(Minor) Rule editor fixes

hk21702:
Work in progress currently.

General improvements and fixes for the rule editor.

- [x] Refactor loading of user and community rules
  - [x] Shared function between user and community rules. Ensures identical behavior.
  - [x] Handling both string and list representations of rule names and comments to fix a bug where only the first char of a name or comment showed up
  - [x] Fixed force bottom rule commenting
  - [x] Fix bottom rule assuming rule is true when the rule exists instead of checking value
- [x] Fix input dialogue
  - [x] Marked input dialogue function as deprecated since It's now just a wrapper around a single line with no other functionality
- [x] Factored out comment dialogue to its method for readability and maintainability reasons 

- [x] Additional refactors
~~- [ ] Speed up loading and saving rules (May move this to a new branch/PR if it is a large endeavor so that the smaller fixes can be merged)~~ Implemented some speedups related to loading. Saving requires a deeper dive into how metadata is handled.


hk21702:
This should be good for review now. There are some unsatisfactory things that I'd like to change, but the scope would likely bleed into an entire refactoring of metadata.py which potentially has significant consequences for the rest of the code base. 

Otherwise, as far as I can tell, the major functions of the rule editor have been fixed. There are some more optimizations that could be made, but it would bleed into how databases are handled as well as metadata stuff.

LionelColaso:
wouldnt it be best to move `def _show_comment_input` to dialogue.py so it could be reused ?

hk21702:
For comments specifically, only the rule editor uses it, and I don't think anything similar to that is or could be used by anything else. Show comment input is mostly just to make sure the text setup of the window is set up consistently and is reused.

LionelColaso:
yeah you are right i had not noticed it calls `show_dialogue_input  `my bad

hk21702:
To be honest, show_dialogue_input could be depreciated since it's now just a wrapper around a single pyside6 method and doesn't do any processing. For now, I'm keeping it there with a warning so it doesn't break anything.

LionelColaso (approved):
Looks goods to merge

PR #350, opened 2024-05-20
add cconstants and fix key error

LionelColaso:
image


app/utils/steam/browser.py
hk21702, line 1: Constants in python should be all caps and defined at the module level per PEP8.

Also, I'm not sure if the constants are even necessary, since they seem to be just used as shorthand for namespace changes? Eliminated the entire namespace in usage is not the best idea because it can cause confusion as to where things are from. If you'd like to shorthand stuff, import the second to last part of the object like AlignmentFlag so useage is ALignmentFlag.AlignCenter instead of QT.AlignmentFlag...

app/windows/missing_mods_panel.py
hk21702, line 1: See comment for browser.py file

app/utils/watchdog.py
```diff
@@ -1,8 +1,7 @@
-from functools import partial
-import json
-from pathlib import Path
 import os
+from functools import partial
 from threading import Timer
+from typing import Optional
```
hk21702, line 4: Optional typing in python3.10+ is recommended to use the union op as Type | None instead of Optional[Type]

app/windows/workshop_mod_updater_panel.py
hk21702, line 1: See comment for browser.py file

hk21702 (changes requested):
See comments for individual files.

Constants in python should be all caps and defined at the module level per PEP8.

Also, I'm not sure if the constants are even necessary, since they seem to be just used as shorthand for namespace changes? Eliminated the entire namespace in usage is not the best idea because it can cause confusion as to where things are from. If you'd like to shorthand stuff, import the second to last part of the object like AlignmentFlag so useage is ALignmentFlag.AlignCenter instead of QT.AlignmentFlag...

app/windows/workshop_mod_updater_panel.py
LionelColaso: fixed

app/utils/watchdog.py
LionelColaso: I  just added the missing import i am not going to modify watchdog since i am sure i will break it hope twsta or you can look into it

hk21702 (commented):
Seems okay to me.

hk21702 (approved):

PR #356, opened 2024-05-22
Minor fixes

LionelColaso:


app/views/mods_panel.py
```diff
@@ -1331,9 +1331,9 @@ def eventFilter(self, source_object: QObject, event: QEvent) -> None:
                                 mod_metadata = (
                                     self.metadata_manager.internal_local_metadata[uuid]
                                 )
-                                if not mod_metadata[
+                                if mod_metadata[
```
hk21702, line 1334: Is there a potential key error for these?

LionelColaso: None at the moment
I am only fixing whatever PEP8 issues Pycharm is complaining about at the moment

LionelColaso: if you want i will undo it ?

LionelColaso: The difference is in the way the first condition is written:
mod_metadata["data_source"] != "expansion" is a more direct way of checking if the value is not equal to "expansion".
not mod_metadata["data_source"] == "expansion" first checks if the value is equal to "expansion", and then negates the result using the not operator.

app/controllers/main_window_controller.py
```diff
@@ -1,7 +1,6 @@
-from loguru import logger
-
 from PySide6.QtCore import QObject, Slot
 from PySide6.QtWidgets import QPushButton
+from loguru import logger
```
twstagg, line 3: I agree that the whitespace should be removed, but import order groups should be sorted alphabetically.

> Imports should be grouped in the following order:
> 
>     Standard library imports.
>     Related third party imports.
>     Local application/library specific imports.

Your grouping choice is correct, but loguru should be imported before PySide6 imports (L comes before P)

app/models/dialogue.py
```diff
@@ -1,7 +1,7 @@
 import os
+import warnings
```
twstagg, line 2: Same as above. Import groups should be sorted alphabetically.

app/views/main_window.py
```diff
@@ -1,7 +1,6 @@
+import os
```
twstagg, line 1: Same as above. Import groups should be sorted alphabetically.

app/views/mods_panel.py
twstagg: I agree that this definitely reads better with Lion's change.. It's really a matter of semantics though, and if it helps us be able to read the code better, I am all for it!

app/windows/rule_editor_panel.py
```diff
@@ -19,6 +18,7 @@
     QVBoxLayout,
     QWidget,
 )
+from loguru import logger
```
twstagg, line 21: Same as above. Import groups should be sorted alphabetically.

twstagg (changes requested):
Some minor nitpicks.

All and all, great work on updating some of those namespace changes! I think that a lot of that is just left over from PySide2 (which is what we initially had) Thank you for taking the time to catch all of these!

app/controllers/main_window_controller.py
LionelColaso: I just Used Optimize Imports feature in PyCharm and it arranged it

app/models/dialogue.py
LionelColaso: I just Used Optimize Imports feature in PyCharm and it arranged it

app/views/main_window.py
LionelColaso: I just Used Optimize Imports feature in PyCharm and it arranged it

app/windows/rule_editor_panel.py
LionelColaso: I just Used Optimize Imports feature in PyCharm and it arranged it

app/models/dialogue.py
twstagg: `import warnings` should be moved to where it was.

app/views/main_window.py
twstagg: `import os` should be moved back to where it was.

app/windows/runner_panel.py
```diff
@@ -3,19 +3,19 @@
 from re import compile, search
 
 import psutil
```
twstagg, line 5: `import psutil` should be below loguru alphabetically.

twstagg: `from re import ...` should be below PySide6.

app/models/dialogue.py
LionelColaso: Currently all "Import" and " from Import" are grouped and sorted

app/views/main_window.py
LionelColaso: same as above

twstagg (approved):

PR #365, opened 2024-05-25
submodule wiki

hk21702:
- add the wiki as a submodule
- remove old duplicated files

The idea here is so that the actual files are synchronized instead of needing to keep one set of documentation in the true wiki repo, and the other in the main repo.

Also allows people to submit pull requests for the documentation more easily


twstagg (approved):
Very cool. Thanks for teaching me this!

PR #369, opened 2024-05-25
(minor) Enable Theme Support

LionelColaso:
- [x] Option to disable stylesheet to use OS/System Theme (current theme "fusion")
- [x] Added Theme Controller
- [x] Added Button to open theme folder location
- [x] Users can add new Themes (stylesheet)
- [x] Added New Themes

How to add your own Themes (stylesheet)
1) Create a new folder in "themes" folder in your RimSort Config Folder
2) Using the default RimPy Theme As Refrence create your own style.qss file in the folder you created
3) Start RimSort and Select your theme from the Dropdown
4) Save settings to apply the Theme
"Note: The Name of Folder will be used as Name of the Theme / Invalid Themes will be ignored

- [x]  Users can contribute to add new Themes by just adding it to "themes" folder in  "application_folder"  and create a pull request 

image

### DIFF
--- a/app/controllers/app_controller.py
+++ b/app/controllers/app_controller.py
@@ -6,6 +6,7 @@ from PySide6.QtWidgets import QApplication
 
 from app.controllers.main_window_controller import MainWindowController
 from app.controllers.settings_controller import SettingsController
+from app.controllers.theme_controller import ThemeController
 from app.models.settings import Settings
 from app.utils.app_info import AppInfo
 from app.utils.constants import DEFAULT_USER_RULES
@@ -21,31 +22,43 @@ class AppController(QObject):
         super().__init__()
 
         self.app = QApplication(sys.argv)
-
-        self.app.setStyle("Fusion")
-
-        self.app.setStyleSheet(  # Add style sheet for styling layouts and widgets
-            (
-                AppInfo().application_folder / "themes" / "RimPy" / "style.qss"
-            ).read_text()
-        )
         self.app.setWindowIcon(GUIInfo().app_icon)
 
-        # One-time initialization of userRules.json
+        self.initialize_user_rules()
+        self.initialize_settings()
+        self.initialize_theme_controller()
+        self.initialize_steamcmd_interface()
+        self.initialize_metadata_manager()
+        self.initialize_main_window()
+
+        self.app.setStyle("Fusion")
+        self.theme_controller.apply_selected_theme(
+            self.settings.enable_themes,
+            self.settings.theme_name,
+        )
+
+    def initialize_user_rules(self) -> None:
+        """Initializes userRules.json if it does not exist."""
         user_rules_path = AppInfo().databases_folder / "userRules.json"
         if not user_rules_path.exists():
             initial_rules_db = DEFAULT_USER_RULES
             with open(user_rules_path, "w", encoding="utf-8") as output:
                 json.dump(initial_rules_db, output, indent=4)
 
-        # Instantiate the settings model, view and controller
+    def initialize_settings(self) -> None:
+        """Initializes the settings model, view, and controller."""
         self.settings = Settings()
         self.settings_dialog = SettingsDialog()
         self.settings_controller = SettingsController(
             model=self.settings, view=self.settings_dialog
         )
 
-        # Initialize SteamcmdInterface
+    def initialize_theme_controller(self) -> None:
+        """Initializes the ThemeController."""
+        self.theme_controller = ThemeController()
+
+    def initialize_steamcmd_interface(self) -> None:
+        """Initializes the SteamcmdInterface."""
         self.steamcmd_wrapper = SteamcmdInterface.instance(
             self.settings_controller.settings.instances[
                 self.settings_controller.settings.current_instance
@@ -53,22 +66,27 @@ class AppController(QObject):
             self.settings_controller.settings.steamcmd_validate_downloads,
         )
 
-        # Initialize the MetadataManager
+    def initialize_metadata_manager(self) -> None:
+        """Initializes the MetadataManager."""
         self.metadata_manager = MetadataManager.instance(
             settings_controller=self.settings_controller
         )
 
-        # Instantiate the main window and its controller
+    def initialize_main_window(self) -> None:
+        """Initializes the main window and its controller."""
         self.main_window = MainWindow(settings_controller=self.settings_controller)
         self.main_window_controller = MainWindowController(self.main_window)
 
     def run(self) -> int:
+        """Runs the main application loop after initializing the main window."""
         self.main_window.show()
         self.main_window.initialize_content(is_initial=True)
         return self.app.exec()
 
     def shutdown_watchdog(self) -> None:
+        """Initiates the shutdown procedure for the watchdog."""
         self.main_window.shutdown_watchdog()
 
     def quit(self) -> None:
+        """Exits the application."""
         self.app.quit()
--- a/app/controllers/settings_controller.py
+++ b/app/controllers/settings_controller.py
@@ -7,6 +7,7 @@ from loguru import logger
 from PySide6.QtCore import QObject, Slot
 from PySide6.QtWidgets import QApplication, QLineEdit, QMessageBox
 
+from app.controllers.theme_controller import ThemeController
 from app.models.settings import Instance, Settings
 from app.utils.constants import SortMethod
 from app.utils.event_bus import EventBus
@@ -55,6 +56,10 @@ class SettingsController(QObject):
         self.settings_dialog = view
 
         self._last_file_dialog_path = str(Path.home())
+
+        self.theme_controller = ThemeController()
+
+        self.app_instance = QApplication.instance()
 
         # Initialize the settings dialog from the settings model
 
@@ -209,6 +214,11 @@ class SettingsController(QObject):
         )
         self.settings_dialog.steamcmd_install_button.clicked.connect(
             self._on_steamcmd_install_button_clicked
+        )
+
+        # Theme tab
+        self.settings_dialog.theme_location_open_button.clicked.connect(
+            self._on_theme_location_open_button_clicked
         )
 
         # Connect signals from dialogs
@@ -504,6 +514,15 @@ class SettingsController(QObject):
             self.settings.todds_overwrite
         )
 
+        # Themes tab
+        self.settings_dialog.enable_themes_checkbox.setChecked(
+            self.settings.enable_themes
+        )
+        self.theme_controller.populate_themes_combobox(
+            self.settings_dialog.themes_combobox
+        )
+        self.theme_controller.setup_theme_dialog(self.settings_dialog, self.settings)
+
         # Advanced tab
         self.settings_dialog.debug_logging_checkbox.setChecked(
             self.settings.debug_logging_enabled
@@ -651,6 +670,12 @@ class SettingsController(QObject):
             self.settings_dialog.todds_overwrite_checkbox.isChecked()
         )
 
+        # Themes tab
+        self.settings.enable_themes = (
+            self.settings_dialog.enable_themes_checkbox.isChecked()
+        )
+        self.settings.theme_name = self.settings_dialog.themes_combobox.currentText()
+
         # Advanced tab
         self.settings.debug_logging_enabled = (
             self.settings_dialog.debug_logging_checkbox.isChecked()
@@ -720,6 +745,10 @@ class SettingsController(QObject):
         self.settings_dialog.close()
         self._update_model_from_view()
         self.settings.save()
+        self.theme_controller.apply_selected_theme(
+            self.settings.enable_themes,
+            self.settings.theme_name,
+        )
 
     @Slot()
     def _on_game_location_text_changed(self) -> None:
@@ -1044,9 +1073,8 @@ class SettingsController(QObject):
             self.settings_dialog.community_rules_db_local_file_choose_button.setEnabled(
                 False
             )
-            app_instance = QApplication.instance()
-            if isinstance(app_instance, QApplication):
-                focused_widget = app_instance.focusWidget()
+            if isinstance(self.app_instance, QApplication):
+                focused_widget = self.app_instance.focusWidget()
                 if focused_widget is not None:
                     focused_widget.clearFocus()
             return
@@ -1117,9 +1145,8 @@ class SettingsController(QObject):
             self.settings_dialog.steam_workshop_db_local_file_choose_button.setEnabled(
                 False
             )
-            app_instance = QApplication.instance()
-            if isinstance(app_instance, QApplication):
-                focused_widget = app_instance.focusWidget()
+            if isinstance(self.app_instance, QApplication):
+                focused_widget = self.app_instance.focusWidget()
                 if focused_widget is not None:
                     focused_widget.clearFocus()
             return
@@ -1307,7 +1334,26 @@ class SettingsController(QObject):
         self.settings.instances[self.settings.current_instance].run_args = run_args_list
         self.settings.save()
 
+    @Slot()
     def _do_reset_settings_file(self) -> None:
         logger.info("Resetting settings file and retrying load")
         self.settings.save()
         self._load_settings()
+
+    @Slot()
+    def _on_theme_location_open_button_clicked(self) -> None:
+        """
+        Open the location of the selected theme.
+        """
+        selected_theme_name = self.settings_dialog.themes_combobox.currentText()
+        logger.info(f"Opening theme location: {selected_theme_name}")
+        stylesheet_path = self.theme_controller.get_theme_stylesheet_path(
+            selected_theme_name
+        )
+
+        if stylesheet_path and stylesheet_path.exists():
+            platform_specific_open(stylesheet_path.parent)
+        else:
+            logger.warning(
+                f"Failed to open theme location: {stylesheet_path} not found or does not exist"
+            )
--- a/app/controllers/theme_controller.py
+++ b/app/controllers/theme_controller.py
@@ -1,0 +1,207 @@
+from pathlib import Path
+from typing import Optional
+
+from loguru import logger
+from PySide6.QtWidgets import QApplication, QComboBox
+
+from app.models.settings import Settings
+from app.utils.app_info import AppInfo
+from app.views import dialogue
+from app.views.settings_dialog import SettingsDialog
+
+
+class ThemeController:
+    def __init__(self, default_theme: str = "RimPy") -> None:
+        """Initialize the ThemeController with a default theme.
+
+        Args:
+            default_theme (str): The name of the default theme. Defaults to "RimPy".
+        """
+        self.app_info = AppInfo()
+        self.app_instance = QApplication.instance()
+        logger.info("Initializing ThemeController")
+        self.default_theme = default_theme
+        self.themes = self._get_supported_themes()
+        logger.info(f"Supported themes: {self.themes}")
+        self.theme_stylesheets_folder = [
+            self.app_info.theme_storage_folder,
+            self.app_info.theme_data_folder,
+        ]
+
+    def _get_supported_themes(self) -> set[str]:
+        """Retrieves a set of supported themes from the theme data and storage folders."""
+        theme_data_folders = self._get_theme_names_from_folder(
+            self.app_info.theme_data_folder
+        )
+        theme_storage_folders = self._get_theme_names_from_folder(
+            self.app_info.theme_storage_folder
+        )
+        supported_themes = theme_data_folders | theme_storage_folders
+        logger.info(f"Supported themes retrieved: {supported_themes}")
+        logger.debug(
+            f"Checking theme folders: {[self.app_info.theme_data_folder, self.app_info.theme_storage_folder]}"
+        )
+        return supported_themes
+
+    def _get_theme_names_from_folder(self, folder: Path) -> set[str]:
+        """Helper method to get theme names from a specified folder.
+
+        Args:
+            folder (Path): The folder to search for themes.
+
+        Returns:
+            set[str]: A set of theme names found in the folder.
+        """
+        supported_themes = set()
+        if folder.exists():
+            for subfolder in folder.iterdir():
+                if subfolder.is_dir() and (subfolder / "style.qss").exists():
+                    supported_themes.add(subfolder.name)
+                    logger.info(
+                        f"Found theme with stylesheet in {folder}: {subfolder.name}"
+                    )
+        else:
+            logger.warning(f"Folder does not exist: {folder}")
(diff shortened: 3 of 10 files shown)


hk21702:
Added (minor) to the pr name for semver reasons. (Is a feature addition).

app/controllers/settings_controller.py
hk21702, line 1: I wonder if it would be better to handle themes using an (index, name) tuple/named tuple. Index is a bit more reliable. 

Also, I question if it is a good idea to force the themes combobox to index 0 when it is at index -1. Index -1 is for when the selection is blank. What if there are no available themes to choose from or it is intentional to choose blank? This can cause a bad selection in the best case and crash in worst case if index 0 does not exist.

app/models/settings.py
```diff
@@ -44,6 +44,8 @@ def __init__(self) -> None:
         self._todds_active_mods_target: bool = False
         self._todds_dry_run: bool = False
         self._todds_overwrite: bool = False
+        self._enable_themes: bool = False
```
hk21702, line 47: Why have the defaults here be different from that in apply default settings when you are just gonna set it in apply default settings right after?

app/utils/theme.py
```diff
@@ -0,0 +1,89 @@
...
+    def get_default_theme_name(self) -> str:
+        """
+        Get the default theme name by scanning the theme data folder.
+        """
+        for folder in theme_data_folder.iterdir():
+            if folder.is_dir():
+                return folder.name
+        return "RimPy"  # Fallback to "RimPy" if no theme folders are found
+
+    def validate_theme(self):
```
hk21702, line 26: needs return type

app/controllers/settings_controller.py
```diff
@@ -444,6 +445,19 @@ def _update_view_from_model(self) -> None:
...
 
+        # themes tab
+        if self.settings.enable_themes:
+            self.settings_dialog.enable_themes_checkbox.setChecked(True)
+        # get theme names
+        current_theme_name = self.settings.theme
+        current_index = self.settings_dialog.themes_combobox.findText(
+            current_theme_name
+        )
+        if current_index != -1:
```
hk21702, line 456: Why force the index to be 0 if the current_index is -1?

Worst case this causes a crash if there is no index 0, best case it picks a theme that the user did not intend to pick. IMO for debugging purposes and just user clairity, it'd just be better to leave it as -1 and down the line handle it and understand that -1 means the selection is blank.

app/utils/theme.py
```diff
@@ -0,0 +1,89 @@
...
+        for folder in theme_storage_folder.iterdir():
+            if folder.is_dir():
+                stylesheet_path = folder / "style.qss"
+                if stylesheet_path.exists():
+                    available_themes.append(folder)
+                else:
+                    logger.warning(
+                        f"Skipping folder '{folder.name}' in theme_folder as it doesn't contain a valid stylesheet."
+                    )
+
```
hk21702, line 88: This is duplicated code and should be condensed for consitency and maintainability. 
```suggestion
        folders = [theme_data_folder, theme_storage_folder]
        available_themes = []

        for dir in folders:
            for folder in dir.iterdir():
                if folder.is_dir():
                    stylesheet_path = folder / "style.qss"
                    if stylesheet_path.exists():
                        available_themes.append(folder)
                    else:
                        logger.warning(
                            f"Skipping folder '{folder.name}' in `{dir}' as it doesn't contain a valid stylesheet."
                        )
```

app/controllers/settings_controller.py
```diff
@@ -577,6 +591,23 @@ def _update_model_from_view(self) -> None:
...
+        selected_theme = self.settings_dialog.themes_combobox.currentText()
+        self.settings.theme = selected_theme
+        available_themes = [folder.name for folder in Themes.get_available_themes()]
+
+        if selected_theme in available_themes:
+            self.settings.theme = selected_theme
+        else:
+            # Handle invalid theme selection
+            logger.warning(f"Invalid theme selected: {selected_theme}")
+            self.settings.theme = "RimPy"  # Set a default theme
```
hk21702, line 609: What if there is no theme RimPy for some reason?

Also, you are modifying the internally stored theme without modifying what is actually shown to the user. Its logged as a warning but it can be potentially confusing to debug for the user if the combo box says one thing, but the theme stays as RimPy.

app/controllers/settings_controller.py
```diff
@@ -577,6 +591,23 @@ def _update_model_from_view(self) -> None:
...
 
+        # themes tab
+        if self.settings_dialog.enable_themes_checkbox.isChecked():
+            self.settings.enable_themes = True
+        else:
+            self.settings.enable_themes = False
+        # get theme names
+        selected_theme = self.settings_dialog.themes_combobox.currentText()
+        self.settings.theme = selected_theme
+        available_themes = [folder.name for folder in Themes.get_available_themes()]
```
hk21702, line 602: I wonder if it would be better to use an index based approach to store the themes. Or at least a tuple pair storing the index and the name. That way you avoid issues with duplicate names, string formatting, etc.

themes/RimPy/style.qss
hk21702, line 1: For consistency in naming convention, I'd keep RunnerPaneltext named as RunnerPanelText and just fix where it shows up to be consistent.

app/utils/theme.py
```diff
@@ -0,0 +1,89 @@
+from pathlib import Path
+from typing import Optional
+
+from loguru import logger
+
+from app.utils.app_info import AppInfo
+
+theme_data_folder = AppInfo().theme_data_folder
+theme_storage_folder = AppInfo().theme_storage_folder
```
hk21702, line 9: I'm not a fan of doing global constants for these. 

I'd have these just be class attributes passed at init that you can override, and then decouple the entire class from AppInfo.

app/utils/theme.py
hk21702, line 1: This file is prob a controller within MVC and should be treated as such within the folder structure instead of just being thrown under 'utils'

hk21702 (changes requested):
See comments. Especially about handling of defaults, -1 index and invalid themes.

app/utils/theme.py
LionelColaso: Done

app/controllers/settings_controller.py
LionelColaso: Done

app/controllers/settings_controller.py
LionelColaso: Not Everyone is Going To use the same set of themes also most people install less then 10 themes max not 100 themes so i think its ok for now

app/utils/theme.py
LionelColaso: i am still not sure about if it should be on its own or a class in app_info

app/utils/theme.py
LionelColaso: i have updated hope u agree with current changes

app/utils/theme.py
LionelColaso: Thankyou i indeed missed that i was tempted to just add the suggestion to the current code but i ended up slightly modifying your idea

themes/RimPy/style.qss
LionelColaso: isnt it already 
RunnerPanelText

app/controllers/settings_controller.py
LionelColaso: This is for a reason if a theme other then Default is set and for some reason it goes missing then instead of showing blank selection it alsways points to Rimpy which is default theme
this has also been addressed in multiple places as a failsafe to avoid the app from launching or crashing just beause of theme 
i could not crash the app due to themes in my testings feel free to crash it :-)

app/controllers/settings_controller.py
LionelColaso: see below comment

app/controllers/app_controller.py
```diff
@@ -21,14 +22,6 @@ def __init__(self):
```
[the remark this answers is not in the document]
LionelColaso: done

app/models/settings.py
LionelColaso: done

app/utils/theme.py
hk21702: Make it so that its on its own. app_info is its weird little thing. I'd avoid messing with that since there are some issues related to it in terms of code structure that causes down stream issues like weird import orders that need to be fixed.

app/controllers/settings_controller.py
hk21702: It does crash if you delete the RimPy theme from the themes folder, albeit the error happens upstream before this even has a chance to error.

Though the user is instructed to put their personal themes under the config folder, I think it's reasonable to say that the app shouldn't crash if the user clears out their themes folder accidentally, especially since it is just named "themes" and a lot of other apps have users just place themes into a themes.

For this behavior, it should default to no theme selected, and fall back to system theme instead of using a qss stylesheet.

themes/RimPy/style.qss
hk21702: Think it got changed at some point in main branch.

hk21702:
Can you move the instructions on how to add your own themes into either a popup or a tooltip? For consistency reasons mostly since it is a lot of text.

Also, the drop-down for selecting QSS themes should be labeled.

I'd also make the checkbox for switching between QSS and System theme to instead be labeled as "Default to Stylesheet Theme instead of System Theme". Or even better, invert the operation and label it as, "Use system theme instead of stylesheet".

I didn't test switching between stylesheets, but ideally the application should switch to and from the system theme without needing a restart. Or at least, the user should be informed that "RimSort needs to restart for some settings to take effect."

hk21702:
> Can you move the instructions on how to add your own themes into either a popup or a tooltip? For consistency reasons mostly since it is a lot of text. I personally think a tooltip would be the most consistent with how things are right now.

Actually, it might be fine to keep it out since its so much text. I'd make it live in another sub-box though for separation purposes.

app/controllers/settings_controller.py
LionelColaso: ok foxed

app/utils/theme.py
LionelColaso: is it ok now

app/views/settings_dialog.py
```diff
@@ -686,6 +688,50 @@ def _do_todds_tab(self) -> None:
...
+        group_layout.addWidget(self.themes_combobox)
+
+        # Add Theme info label
+        self.themes_info_label = QLabel(
+            "How to add your own Themes (stylesheet) \n"
+            "1) Create a new folder in 'themes' folder in your RimSort Config Folder \n"
+            "2) Using the default 'RimPy' Theme As Refrence create your own 'style.qss' file in the folder you created \n"
+            "3) Start RimSort and Select your theme from the Dropdown \n"
+            "4) Restart RimSort to apply the Theme \n\n"
+            "Note: The Name of Folder will be used as Name of the Theme / Invalid Themes will be ignored"
```
hk21702, line 730: Fix spelling mistakes and inconsistent capitalization.

Also, I think this info would be better in its own dialogue instead of a QLabel. Then you get rich text formatting and don't have everything stuck as QLabel stylization which is just bold.

app/views/settings_dialog.py
```diff
@@ -686,6 +688,50 @@ def _do_todds_tab(self) -> None:
...
+        group_layout = QVBoxLayout()
+        group_box.setLayout(group_layout)
+
+        self.themes_label = QLabel("Themes Support")
+        self.themes_label.setFont(GUIInfo().emphasis_font)
+        group_layout.addWidget(self.themes_label)
+
+        self.enable_themes_checkbox = QCheckBox(
+            "Enable to use .qss Themes (stylesheet) / Disable to use OS/System Theme"
+        )
```
hk21702, line 710: Compress the info for the label. Add a tooltip for additional context and info.

app/views/settings_dialog.py
hk21702: Additionally, should add a button to open the theme directory in explorer.

app/views/settings_dialog.py
```diff
@@ -686,6 +688,50 @@ def _do_todds_tab(self) -> None:
...
+        group_layout.addWidget(self.enable_themes_checkbox)
+
+        self.themes_combobox = QComboBox()
```
hk21702, line 713: Needs tooltip. Additionally, should be disabled whenever the themes checkbox is not checked.

app/views/settings_dialog.py
hk21702: E.g.,
Label: Use stylesheet instead of system theme

Tooltip: Utilize the selected .qss stylesheet for the theme instead of the system theme

app/controllers/theme_controller.py
```diff
@@ -0,0 +1,123 @@
...
+                if folder.is_dir():
+                    stylesheet_path = folder / "style.qss"
+                    if stylesheet_path.exists():
+                        available_themes.append(folder)
+                    else:
+                        logger.warning(
+                            f"Skipping folder '{folder.name}' in `{theme_path}' as it doesn't contain a valid stylesheet."
+                        )
+
+        return available_themes
```
hk21702, line 123: I'm not sure if I agree with this method storing stylesheets. A lot of the themes you see for QT comes as a single file. The differences are differentiated via the filename. The corresponding icons are put in a folder with the same name as the filename.

It'd be a themes folder with all the stylesheets for different themes living at root of the themes folder. The only subfolders in there would be for icons.

As an example, see how MO2 handles it.

app/views/settings_dialog.py
```diff
@@ -686,6 +688,50 @@ def _do_todds_tab(self) -> None:
...
+        tab_layout = QVBoxLayout(tab)
+        tab_layout.setAlignment(Qt.AlignmentFlag.AlignTop)
+
+        group_box = QGroupBox()
+        tab_layout.addWidget(group_box)
+
+        group_layout = QVBoxLayout()
+        group_box.setLayout(group_layout)
+
+        self.themes_label = QLabel("Themes Support")
```
hk21702, line 704: Is this label necessary, there is only one section and the tab title of Themes should be enough as a descriptor? Additionally, if chosen to be kept, support should be lowercase for consistency.

app/views/settings_dialog.py
```diff
@@ -686,6 +688,50 @@ def _do_todds_tab(self) -> None:
         )
         group_layout.addWidget(self.todds_overwrite_checkbox)
 
+    def _do_themes_tab(self) -> None:
+        tab = QWidget()
+        self.tab_widget.addTab(tab, "Themes")
```
hk21702, line 693: Should probably not be plural

app/controllers/app_controller.py
```diff
@@ -62,6 +56,13 @@ def __init__(self) -> None:
         self.main_window = MainWindow(settings_controller=self.settings_controller)
         self.main_window_controller = MainWindowController(self.main_window)
 
+        # Initialize themes
+        if self.settings.enable_themes is True:
+            self.app.setStyle("Fusion")
+            theme_name = self.settings.theme_name
+            theme = Themes(theme_name)
+            self.app.setStyleSheet(theme.style_sheet())
```
hk21702, line 64: How does this handle where the provided stylesheet is not a valid format?

hk21702 (changes requested):
See comments. In particular handling of the settings dialog, and process of installing themes. As a good example that is pretty standard for QT applications, see MO2

Additionally, you can likely remove the requirement to restart RimSort by just checking for when the selected theme changes, emitting a signal, and then setting it.

app/views/settings_dialog.py
LionelColaso: Done

app/views/settings_dialog.py
LionelColaso: Done

app/views/settings_dialog.py
LionelColaso: removed Done

app/views/settings_dialog.py
LionelColaso: Done

app/views/settings_dialog.py
LionelColaso: Done

app/controllers/app_controller.py
LionelColaso: if the stylesheet in not valid format it is ignored there is a small validation at the moment which only allows .qss file

hk21702: What if it is an invalid qss file?

app/controllers/theme_controller.py
hk21702: For structural things that affect how the user would interact with the system, I'd really rather do it now. That avoids potentially breaking user space in the future.

app/controllers/app_controller.py
LionelColaso: its loaded but since its loaded on top of fusion it doesnt affect much

app/controllers/theme_controller.py
LionelColaso: all issues related to themes breaking anything are now taken care of

app/controllers/theme_controller.py
```diff
@@ -0,0 +1,158 @@
...
+    def apply_theme(self) -> None:
+        """Applies the current theme by setting the application stylesheet."""
+        stylesheet = self.load_theme(self.theme_name)
+        if stylesheet:
+            logger.info(
+                f"Applying theme: {self.theme_name} with stylesheet: {stylesheet}"
+            )
+            # Implement theme application logic here
+        else:
+            logger.error(f"Failed to apply theme: {self.theme_name}")
```
hk21702, line 158: I'm not really sure what this method is for.

app/controllers/theme_controller.py
```diff
@@ -0,0 +1,158 @@
...
+        """Returns the path to the stylesheet for the specified theme."""
+        logger.info(f"Searching for stylesheet for theme: {theme_name}")
+
+        # Initialize a variable to hold the path
+        stylesheet_path = None
+
+        for folder in self.theme_stylesheets_folder:
+            # Construct the stylesheet path
+            potential_path = folder / theme_name / "style.qss"
+            logger.debug(f"Checking for stylesheet at: {potential_path}")
```
hk21702, line 132: I still don't really agree with this method of storing themes. Whats the reasoning for getting rid of the the middle "theme_name" folder and using the .qss file name as the theme name? That seems to be standard in other QT apps.

app/views/settings_dialog.py
```diff
@@ -686,6 +688,51 @@ def _do_todds_tab(self) -> None:
...
         group_layout.addWidget(self.todds_overwrite_checkbox)
 
+    def _do_themes_tab(self) -> None:
+        tab = QWidget()
+        self.tab_widget.addTab(tab, "Theme")
+
+        tab_layout = QVBoxLayout(tab)
+        tab_layout.setAlignment(Qt.AlignmentFlag.AlignTop)
+
+        theme_group_box = QGroupBox("Themes")  # Added a title for the group box
```
hk21702, line 698: Doesn't have the correct styling compared to other group boxes

app/views/settings_dialog.py
```diff
@@ -686,6 +688,51 @@ def _do_todds_tab(self) -> None:
...
+        theme_controller = ThemeController()
+        available_themes = list(
+            theme_controller.get_supported_themes()
+        )  # Ensure it's a list
+        self.themes_combobox.addItems(available_themes)
+        theme_group_layout.addWidget(self.themes_combobox)
+
+        self.theme_location_open_button = QToolButton()
+        self.theme_location_open_button.setText("Open Theme Location")
+        theme_group_layout.addWidget(self.theme_location_open_button)
```
hk21702, line 734: Move this to the same line as the drop down menu. Also, it should really be opening the user themes folder, not the theme currently being used. The idea is that you click it to go to the directory to install new themes or remove them.

app/utils/app_info.py
```diff
@@ -175,6 +177,13 @@ def theme_data_folder(self) -> Path:
         """
         return self._theme_data_folder
 
+    @property
+    def theme_storage_folder(self) -> Path:
+        """
+        Get the path to the folder where application-specific data is stored.
+        """
+        return self._theme_storage_folder
```
hk21702, line 185: These need better doc string.

app/controllers/theme_controller.py
```diff
@@ -0,0 +1,158 @@
...
+        self.app_info = AppInfo()
+        logger.info("Initializing ThemeController")
+        self.default_theme = default_theme
+        self._theme_name = self.load_theme_name() or self.default_theme
+        logger.info(f"Current theme name: {self._theme_name}")
+        self.themes = self.get_supported_themes()
+        logger.info(f"Supported themes: {self.themes}")
+        self.theme_stylesheets_folder = [
+            self.app_info.theme_data_folder,
+            self.app_info.theme_storage_folder,
```
hk21702, line 21: Flip the order to prioritize storage when looking for themes for the case where there are duplicate names.

hk21702 (changes requested):
See comments.

In particular, things related to settings should not be handled by theme_controller, especially the reading of the settings file. Additionally, I still don't agree with how you're storing the themes and I'm still not sure why you're doing it this way. There are also some minor changes that should be made in terms of doc strings and formatting of the settings dialog.

app/controllers/theme_controller.py
LionelColaso: yeah right done Thanks

app/views/settings_dialog.py
LionelColaso: ok done

app/views/settings_dialog.py
LionelColaso: ok done

app/controllers/theme_controller.py
LionelColaso: Removed 
This was for something i was trying a left over

app/utils/app_info.py
LionelColaso: Yeah Done

app/controllers/theme_controller.py
```diff
@@ -0,0 +1,158 @@
...
+                    saved_theme = json.load(file)
+                    theme_name = saved_theme.get("theme_name")
+                    logger.info(f"Loaded theme name {theme_name} from userRules.json")
+                    return theme_name
+            except Exception as e:
+                logger.error(f"Error loading theme name from userRules.json: {e}")
+                return ""
+        else:
+            logger.warning(f"userRules.json does not exist: {stored_settings}")
+            return ""
```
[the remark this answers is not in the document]
LionelColaso: for some reason i couldnt get the theme_name reliably from settings
would often get the theme_name correctly 
so till then this was the most reliable way of getting the theme_name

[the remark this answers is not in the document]
LionelColaso: now it loads the themes from settings correctly

twstagg (approved):

PR #370, opened 2024-05-25
Non python lint fixes

hk21702:
Linting fixes for non-python related things


LionelColaso (approved):
looks Good

hk21702:
Will hold off on merging so @twstagg can take one last look over

twstagg (changes requested):
Let's skip any changes to the update.sh, we are going to be removing this soon in favor of a dedicated updater module/program.

twstagg (approved):

PR #373, opened 2024-05-26
fix error while creating new instance

LionelColaso:
fix new instance creation
image

### DIFF
--- a/app/views/main_window.py
+++ b/app/views/main_window.py
@@ -201,23 +201,23 @@ class MainWindow(QMainWindow):
             # Setup watchdog
             self.initialize_watchdog()
 
-    def __ask_for_new_instance_name(self) -> str:
-        instance_name, cancelled = show_dialogue_input(
+    def __ask_for_new_instance_name(self) -> str | None:
+        instance_name, ok = show_dialogue_input(
             title="Create new instance",
-            text="Input a unique name of new instance that is not already used:",
+            label="Input a unique name of new instance that is not already used:",
         )
-        return instance_name if not cancelled else None
+        return instance_name.strip() if ok else None
 
-    def __ask_for_non_default_instance_name(self) -> str:
+    def __ask_for_non_default_instance_name(self) -> str | None:
         while True:
-            instance_name, cancelled = show_dialogue_input(
+            instance_name, ok = show_dialogue_input(
                 title="Provide instance name",
-                text='Input a unique name for the backed up instance that is not "Default"',
+                label='Input a unique name for the backed up instance that is not "Default"',
             )
-            if cancelled:
-                return None
-            if instance_name.lower() != "default":
+            if ok and instance_name.lower() != "default":
                 return instance_name
+            else:
+                return None
 
     def __ask_how_to_workshop_mods(
         self, existing_instance_name: str, existing_instance_workshop_folder: str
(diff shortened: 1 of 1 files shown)


app/views/main_window.py
```diff
@@ -569,7 +572,7 @@ def clone_essential_paths(
             )
         )
         # Sanitize the input so that it does not produce any KeyError down the road
-        new_instance_name = self.__ask_for_new_instance_name().strip()
```
hk21702, line 572: You shouldn't be just removing the .strip(), then you're not sanitizing it

Fix the underlying issue.

app/views/main_window.py
```diff
@@ -202,22 +202,25 @@ def initialize_content(self, is_initial: bool = True) -> None:
...
+            return instance_name.strip()
+        else:
+            return None
 
     def __ask_for_non_default_instance_name(self) -> str:
         while True:
-            instance_name, cancelled = show_dialogue_input(
+            instance_name, ok = show_dialogue_input(
                 title="Provide instance name",
                 text='Input a unique name for the backed up instance that is not "Default"',
```
hk21702, line 218: Change to label

app/views/main_window.py
LionelColaso: it is 
def __ask_for_new_instance_name(self) -> str:
        instance_name, ok = show_dialogue_input(
            title="Create new instance",
            text="Input a unique name of new instance that is not already used:",
        )
        if ok:
            return instance_name.strip()
        else:
            return None

app/views/main_window.py
```diff
@@ -202,29 +202,32 @@ def initialize_content(self, is_initial: bool = True) -> None:
             self.initialize_watchdog()
 
     def __ask_for_new_instance_name(self) -> str:
```
hk21702, line 204: return type should be str | None

app/views/main_window.py
```diff
@@ -202,29 +202,32 @@ def initialize_content(self, is_initial: bool = True) -> None:
...
-            text="Input a unique name of new instance that is not already used:",
+            label="Input a unique name of new instance that is not already used:",
         )
-        return instance_name if not cancelled else None
+        if ok:
+            return instance_name.strip()
+        else:
+            return None
 
     def __ask_for_non_default_instance_name(self) -> str:
```
hk21702, line 214: return type should be str | None

hk21702:
@twstagg can you take a look since you're more familiar with how instancing is supposed to work?

app/views/main_window.py
```diff
@@ -201,23 +201,26 @@ def initialize_content(self, is_initial: bool = True) -> None:
...
-    def __ask_for_new_instance_name(self) -> str:
-        instance_name, cancelled = show_dialogue_input(
+    def __ask_for_new_instance_name(self) -> str | None:
+        instance_name, ok = show_dialogue_input(
             title="Create new instance",
-            text="Input a unique name of new instance that is not already used:",
+            label="Input a unique name of new instance that is not already used:",
         )
-        return instance_name if not cancelled else None
+        if ok:
```
twstagg, line 209: This could still be a single line.

twstagg (changes requested):
Everything seems to work fine. i see that the strip is still happening. Looks good other than my nitpick.

twstagg (approved):
Thanks!

PR #378, opened 2024-05-27
Some ruff fixes

LionelColaso:


hk21702:
There appears to be a build failure, though it seems to be part of attestation stuff. I'm not entirely sure why it’s failing, since the same action just worked fine like 3 mins ago on main and on a different branch. Will rerun action for your pr before approving.

hk21702:
Oh, its because you're working off of your fork. Yeah, that'll cause issues. Though I can see that the issue is just with attestations and not making the build, so I will approve for now.

hk21702 (approved):

PR #386, opened 2024-06-03
Fix crash on opening settings with a new instance

hk21702:
Fixes an incorrect type crash observed when opening settings in a new instance/setting up a new instance

This is mostly just a quick patch. Should seriously consider reworking how settings are handled so that these kinds of bugs are rarer/impossible.


LionelColaso (approved):
I cant Check this Since i Dont own The Game on Steam but if it fixes for you then great go ahead

PR #388, opened 2024-06-03
Add warning to restart the app on git install

hk21702:
Adds to the popup dialogue as well as the log warning that the app needs to be restarted after installing Git in order for the path changes to take effect.


hk21702:
As a side note, need to get rid of those star imports in main_content_panel at some point

LionelColaso (approved):
Great

PR #390, opened 2024-06-04
Format with ruff

LionelColaso:
eliminate ruff related complaints


hk21702 (approved):
Checks out for me.

Added lxml-stubs to dev requirements since I noticed it was an error

PR #402, opened 2024-06-10
fix repo is not defined crash

hk21702:
Patch to fix 'Repo' is not defined crash when Git is not installed.


LionelColaso (approved):
good to go

Issue #380, opened 2024-05-30
Suggestion: Don't delete the About folders/xml files when recursively deleting/batch redownloading mods


Ookaishi:
**Is your feature request related to a problem? Please describe.**
I have recently swapped to RimSort and I wanted to do a full redownload of all of my mods, sadly the process broke halfway through and now I'm left with a ton of Invalid.Item pseudo-mods which can't be used to redownload the actual mods.
The folders, mod steam workshop IDs are all there, there's just no way to simply attempt a redownload using that.
Having a backup of the modlist.xml doesn't help in this case.
**Describe the solution you'd like**
1) Not deleting the About folders/xml files as to allow proper identification of mods in case something happens.
2) 2-3 Backups (not temp files) of the steamcmd.txt files used to batch download mods, that can be re-used in case of crashes.
3) Have the mods that are being redownloaded placed in a temporary location until they're validated and then moved over to the final location (Although this is a slower option and not always feasible, it's worth considering nonetheless)
**Describe alternatives you've considered**
An alternative would be allowing batch downloading using the modlist.xml, similar to the way RimPy handles missing mods.
Also a backup of the current/active mod list and inactive mods list would be useful as it can be used to properly track mods in situations such as this.
Another alternative would be caching the About folders of every mod somewhere else, if splitting those and the mod data into two different folders and then "merging" them back through sym/hardlinking isn't possible.
**Additional context**
A bit of extra info: Trying to redownload the mods manually results in a failed download, usually saying that the mod could not be found or that no publishedfield was provided.
For context, this is what over half of my mods list is atm:
![image](https://github.com/RimSort/RimSort/assets/22888160/aba695c5-34d0-40f8-b0eb-9bc35bfe5c80)

_Edited to add another alternative solution_

Ookaishi:
A little update: Managed to recover some of the mods through RimPy's downloader (ugh...oh well.) and through manually downloading them directly with steamCMD; Since the folders that were left still had the workshop IDs as names, all it took was a batch script to write the commands out to pass to steamCMD in a file and then running that in steamCMD.
If interested I can post the script here.

Chunnyluny:
please do so others can maybe use it if they think it is useful for their own cases! thanks for finding your own solution and share it so others can recover their fieles back if they are originally rimpy users!!

Ookaishi:
[...]

PR #403, opened 2024-06-11
(minor) Metadata rework backend

hk21702:
A rework of how metadata is handled. Utilize proper design patterns as well as static type checking to help eliminate a lot of the current issues around metadata. Utilize msgspec for faster json parsing. Ensure multi-threading support. Ensure there are unit tests where relevant.

Goals:
- [x]  Add PyTests
- [x] Represent metadata as classes/data classes
- [x] Schema based rules DB parsing
- [x] Schema based steam DB parsing
- [x] Separation between structure and views
- [x] Ensure function parity
  - [x] About.xml
  - [x] Scenarios
  - [x] Mod List
    - [x] Reading
    - [x] Writing 
  - [x] User Rules
  - [x] Community Rules
  - [x] Steam DB
  - [x] SteamCMD handling
  - [x] Add external rules to a mod's metadata
 - [x] Implement local auxiliary metadata using SQLAlchemy with SQLite as backend
 - [x] Implement new metadata controller
 - [x] Implement metadata removal/deletion
 - [ ] ~~Integrate new metadata system~~ To be done in other PR


hk21702:
To handle additional meta data for future features like mod colors, user notes, custom urls etc, I am thinking of doing a metadata file that is generated by RimSort.
There's two options of handling this. Having a big universal file that maps to packageids, or having an individual file per folder, somewhat like the meta.ini of mod organizer 2. 

The advantage of the former is that it gets around any potential issues with adding files in the mod folder. However, I think in 99% of cases its fine to add files. Steam verify shouldn't pick up and delete new files. Am not sure about steamcmd. Git updates is fine as long as the pull is forced assuming that the metadata file generated by RimSort is not tracked by the git repository (which it should not unless the author of the repo messed up).

The real downside of the former is that it is not true individualized metadata. If two folders share the same packageid, then doing a mapping method would mean that they would share custom metadata. As such, I think only the later is a viable solution.

This file could also potentially act as a backup cache to help prevent issues like described in #380, giving a little bit more info than just "invalid mod" when possible.

For that use case, when loading all metadata, we should first check for the about file. If it exists, then force the metadata file to match the about file.  If the about file does not exist, fall back to the metadata file for very basic string information like mod name, package id, and maybe version. For sorting and loading purposes, a mod without the about file should still be considered invalid, just that the metadata helps with identifying what it is.

hk21702:
As a continuation of the previous comment, having the metadata being stored in the mod file itself could potentially have issues with pulling from git mods. The file would be considered an unstaged change and may get overwritten or throw an error.

It may be best to utilize a database library with folder path keys instead? Multithreaded read write access would be a must, and document based would make the most sense for schema. Though, something like SQLite would likely also be fine. Using a proper db instead of a file does sacrifice some user operability to go in and manually edit things, but in 99% of cases they shouldn't be doing that anyways.

LionelColaso (approved):
Everything seems to be working except 
re_steamcmd_action (not sure if resubscribing with steam is also broken as well since i can't test this)
not related to metadata rework and can be fixed separately

hk21702:
> Everything seems to be working except re_steamcmd_action (not sure if resubscribing with steam is also broken as well since i can't test this) not related to metadata rework and can be fixed separately since its broken on main branch as well

This PR shouldn't change any functionality at all, since it is not implemented or integrated. It is just backend code that hasn't been hooked up to anything.

PR #405, opened 2024-06-12
Instance and "Check Flow" Improvements 

hk21702:
- Convert instances to classes with struct support
- Fix local mods monitoring
- General Typing fixes
- Make InstanceController to fix MVC violations
- Utilize msgspec for better instance to json handling and structure
- Minor improvements to user flow
- Prevent windows junctions from being zipped on instance backup
- Warning on restore if an instance already exists and extraction will replace the existing folder
- Validate paths on instance restore

- Check Flow Improvements
  - No more steam_cmd check on "ok" event. 
    - This means no double dialogue and attempt to double install when manually clicking install steam cmd, and avoiding the potentially faulty state that can happen with the ok triggered dialogue
  - Initial steam cmd check is done after the refresh event.
  - Added a spin lock of sorts on the refresh event if the settings page was opened, waiting for it to be closed.
    -  This is so the steam cmd check is truly done after the user has a chance to set the path settings and the pop up doesn't happen while they just got into the settings page.


hk21702:
Instance backup restoration seems to be broken in the case where the instance backup is named something that does not already exist (or if the instance local files are not pre-existing). During restoration, the symlink check fails, and a crash occurs due to the path not existing. This bug seems to originate from when instances were originally introduced.

LionelColaso:
personally i still think user should be asked if he wants to reuse current game paths and steamcmd path or he wants to set them up when creating an instance
this could be added later as well
except this i have no issues

LionelColaso (approved):
Good Work

twstagg:
> personally i still think user should be asked if he wants to reuse current game paths and steamcmd path or he wants to set them up when creating an instance this could be added later as well except this i have no issues

It should do this through the normal check of those paths. So if you create a fresh blank instance (not cloned) then it will be empty, and then the essential paths check will fail, which prompts the user if they want to set them up. If they choose yes in that dialog, it will take them to Settings -> Locations to configure their path. This check is instance agnostic, and will happen for whatever your "current instance" is set as, as you load into it.

twstagg (approved):

PR #412, opened 2024-06-17
Graceful fail if exception when validating mods list

hk21702:
Graceful fail if exception, such as if the mods_config_data is None for some reason, when validating mod lists. 

Gives a warning to the user instead of a vague hard crash that is difficult to self diagnose.


LionelColaso (approved):
Good

PR #419, opened 2024-06-19
[417] toggle warnings icon fixes

oceancabbage:
Small improvements for the "toggle warning icon" feature:

- Adding debug statement when a toggle happens
- Toggling "off" the last warning while errors exist (or the last error while warnings exist) does not remove the summary text for warnings (or errors) due to the if/else logic. Fixing so this is behaving as expected.
- Toggling "off" version mismatch errors does not remove the summary text due to missing a condition. Fixing so this is behaving as expected.

Testing: manually tested on my Windows machine


hk21702 (approved):
Looks good. 

I'd consider maybe also turning the "Toggle Warnings" context action into a checkable action while you're at it.

PR #426, opened 2024-06-19
Refactored settings model

hk21702:
Refactored settings model. Goal of this is to make it much easier to modify the model. There is only one place that needs to be modified in 9/10 instances where you want to add or remove or modify an attribute. And in the 1/10 instances, the process is relatively easy and non-breaking for other attributes.

Get rid of the apply_default_settings method which was just duplicate code when you can just make a new instance
Apply the default values from the apply_default_settings method to init default values.
Override default __setattr__ method to emulate the behavior of the setters and getters. Remove setters and getters which are now redundant.
Change relevant attributes that no longer have setters and getters from private to public.
Refactor _from_dict and _to_dict such that the attributes they target are based on the method's targets.

PR and branch is based on #405 

#405 should be merged before this.


LionelColaso (approved):

app/models/settings.py
```diff
@@ -21,400 +21,61 @@ def __init__(self) -> None:
...
+        self.duplicate_mods_warning: bool = False
+        self.steam_mods_update_check: bool = False
+        self.try_download_missing_mods: bool = False
+        self.steamcmd_validate_downloads: bool = True
+        self.todds_preset: str = "optimized"
+        self.todds_active_mods_target: bool = True
+        self.todds_dry_run: bool = False
+        self.todds_overwrite: bool = False
+        self.current_instance: str = "Default"
+        self.instances: dict[str, Instance] = {"Default": Instance()}
```
LionelColaso, line 58: just move 
self.instances: dict[str, Instance] = {"Default": Instance()}
below
self.steam_apikey: str = ""
since all instances are added below that in the json

LionelColaso (changes requested):

LionelColaso (approved):

LionelColaso (changes requested):
check suggesstion

app/models/settings.py
hk21702: I don't think the order should matter? The only thing it affects is the order of writing when dumping to json. Reading has no effect since the json file is just converted to a dictionary anyways.

LionelColaso: agreed but it looks nice in settings.json

LionelColaso (approved):

oceancabbage (approved):
This is clean. Great stuff

Issue #371, opened 2024-05-26
FIXME: RimSort crashes when dragging active mods into an inactive list


windicevista:
**Describe the bug**
RimSort crashes when dragging active mods into an inactive list

**To Reproduce**
1.Dragging nearly 15+ mods from "Active" to "Inactive"

**Expected behavior**
Drag the module from "Active" to "Inactive" and Save

**Screenshots**
![image](https://github.com/RimSort/RimSort/assets/7019719/be3d3ce7-8e80-435b-8779-9d65a5ccd1be)

**Please complete the following information:**
 - OS: Windows 10
 - RimSort version: v1.0.7
 - Please provide a `RimSort.log`:[RimSort.log](https://github.com/RimSort/RimSort/files/15445468/RimSort.log)
 - Running from a compiled release

Limfjorden:
Isn't this the same issue as https://github.com/RimSort/RimSort/issues/327

Issue #327, opened 2024-05-10
RimSort crashes when dragging inactive mods into an empty list


lacqui:
**Describe the bug**
When I drag mods to the right in order to activate them, RimSort crashes if the active mods list only contains the official content.

**To Reproduce**
1. Clear the mods list
2. Drag mods from "Inactive" to "Active"

**Expected behavior**
Dragged mods are activated

**Please complete the following information:**
 - OS: Windows 11
 - RimSort version: Edge 978b59e and af0eae1
- [RimSort.log](https://github.com/RimSort/RimSort/files/15280318/RimSort.log)
- Running from a compiled release

lacqui:
It works when I drag a few at a time, so it may be the same issue.

PR #429, opened 2024-06-20
Fixes occasional crash when moving lots of mods

hk21702:
Quick and dirty fix for the crash that happens when moving many mods from inactive to active or vice versa.

I'm still not really sure why the data ends up being None, but just skipping the item seems to have no negative effects, and all the mods show up as needed when moved.

Fixes: #371, #327


LionelColaso (approved):
Lol Works
Wonder How its Like to move 30000+ mods at once with this one

PR #437, opened 2024-06-22
Catch non-empty directory error when deleting mod

Madrigalist:
Fix for #438


hk21702 (changes requested):
Thanks for the PR.

Can you make it such that it catches all OSErrors? Any OS error shouldn't be a fatal error that causes RimSort to crash. Also, utilize logger.warning instead of debug and also use a show_warning dialogue to show a visible dialogue on catch. The user message should include e.winerror if the os is windows, e.errno otherwise, and also show e.strerror and e.filename.

Madrigalist:
> Can you make it such that it catches all OSErrors? Any OS error shouldn't be a fatal error that causes RimSort to crash.

See changes. Is catching all OSErrors this way a good idea? I feel it should log them at least

Madrigalist:
@hk21702 Have a look at the latest commit

oceancabbage (approved):

LionelColaso:
> @hk21702 Have a look at the latest commit

Go ahead and Merge

PR #440, opened 2024-06-23
Add arch, duplicate issue and sort method question

hk21702:
Modifies the issue bug report template to add a question about the processor architecture, what sort method they are using if applicable, and also asks the user to first check if their issue had already been reported.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -30,6 +30,7 @@ If applicable, add screenshots to help explain your problem.
 - RimSort version:
 - Please provide a `RimSort.log`:
 - Are you running from the Python interpreter or from a compiled RimSort release?
+- If this is a sorting issue, what sort mode are you using?: [Alphabetical/Topological]
 
 **Additional context:**
 Add any other context about the problem here.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -25,12 +25,17 @@ If applicable, add screenshots to help explain your problem.
 
 **Please complete the following information:**
 
-- OS: [e.g. iOS]
+- Operating System: [e.g. Windows 11]
```
LionelColaso, line 28: Os Naming Can be such as 
Windows (Version) , Mac (Version) , Linux (Name)(Version)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -25,12 +25,17 @@ If applicable, add screenshots to help explain your problem.
...
+- Processor Architecture: [Arm/x86/Unknown]
```
LionelColaso, line 29: you could add more if you want 
https://en.wikipedia.org/wiki/Comparison_of_instruction_set_architectures

LionelColaso (approved):
all good modify and merge if u like

.github/ISSUE_TEMPLATE/bug_report.md
hk21702: The only ones we support are Arm and x86 so I think its fine lol. Even though I've worked with MIPs assembly before, I doubt we will ever support RimSort MIPs.

PR #441, opened 2024-06-23
Implement Sort Controller

hk21702:
Implements a controller for sorting functions.

This solves some MVC violations in main_content_panel.
Internal handling of what sort method also now uses an Enum to better prevent key errors and the like.

Also fixes some typing issues related to sorting in main_content_panel

Note that there are still some violations, but it requires the metadata rework first. 
This PR is based on #426 and requires it to be merged first.


divineDerivative:
Doesn't using an Enum preclude users making their own sorting algorithms? I'm not familiar with Python but in C# you can't add to an enum. Or is it intended that people making their own algorithms need to maintain their own fork of RimSort?

hk21702:
> Doesn't using an Enum preclude users making their own sorting algorithms? I'm not familiar with Python but in C# you can't add to an enum. Or is it intended that people making their own algorithms need to maintain their own fork of RimSort?

I'm not entirely sure what you mean. It's not intended for users to be able to plug and play their sorting algorithms. If they have knowledge to develop and implement their own algorithm, then they have enough Python knowledge to add to the enum where the enum is defined. There shouldn't be a need to dynamically add onto the enum object. The reason for using an enum instead of say, a dictionary or just strings, is to prevent potential sources of key errors. It can still be extended easily enough if need be, but there shouldn't be a need to dynamically extend the object. 

If someone is making their own sorting algorithm, then they can do it in a fork or a branch. I don't think we'd ever really support just dropping in a python file with a sort function, if that is what you're asking about.

divineDerivative:
[This](https://github.com/RimSort/RimSort/wiki/User-Guide#writing-your-own-algorithm) reads to me like you can indeed just drop in a file. If modifying the actual application is required I would not mention this feature in the User Guide, but in the Development Guide.

hk21702:
> [This](https://github.com/RimSort/RimSort/wiki/User-Guide#writing-your-own-algorithm) reads to me like you can indeed just drop in a file. If modifying the actual application is required I would not mention this feature in the User Guide, but in the Development Guide.

I see what you mean. You're right, the wiki does need to be changed. Currently, sorting algorithms aren't like what the wiki is describing, either. You can't just drop in a function that sorts directed graphs and expect it to work. I personally don't think that It's something that will be supported either.

LionelColaso (approved):
looks good

app/controllers/sort_controller.py
```diff
@@ -0,0 +1,93 @@
...
+            for i, graph in enumerate(dependency_graphs):
+                logger.info(f"Sorting tier {i + 1}")
+                sorted_mods = self.sort_method(graph, self.active_uuids)
+                logger.info(f"Tier {i + 1} sorted: {len(sorted_mods)}")
+                sorted_uuids += sorted_mods
+        except CircularDependencyError:
+            logger.info("Circular dependency detected, abandoning sort")
+            return False, []
+
+        return True, list(dict.fromkeys(sorted_uuids))
```
oceancabbage, line 93: Nice!

oceancabbage (approved):
LGTM, thanks for this

PR #476, opened 2024-07-08
[475] adding log hints for circular dependency

oceancabbage:
In some mod lists toposort may find circular dependencies. This is almost always because of an issue with how mods have specified dependencies or load orders (or because of inaccurate community rules).

In any case we want to easily figure out which mods are conflicting with which. Currently it's very difficult to do because toposort spits out a ton of mods with no direction. This PR adds `find_circular_dependencies` which will beautifully log circular dependencies so we can easily see which mods are causing issues, e.g.:

`harmony -> abc -> xyz`
^ in this example, harmony depends on xyz but xyz also depends on harmony

Another addition in this PR is a small bugfix to `get_dependencies_recursive` where if there are circular dependencies in tier one mods an infinite loop happens as we try to recurse infinitely based on the circle. The fix is just to keep track of what mods we've already processed.

Reviewers-- please pull and test to make sure this isn't breaking anything :)


oceancabbage:
Don't approve yet I'm going to make a second revision

oceancabbage:
Alright the latest revision makes it so that circular dependencies create a warning modal and clicking OK just abandons sort (instead of quitting the application).

LionelColaso (approved):

PR #480, opened 2024-07-13
Move Workshop database expiry to Database Tab

LionelColaso:
Moving This since it makes more sense here

Screenshot 2024-07-13 131728

### DIFF
--- a/app/controllers/settings_controller.py
+++ b/app/controllers/settings_controller.py
@@ -422,6 +422,7 @@ class SettingsController(QObject):
             self.settings.external_steam_metadata_repo
         )
         self.settings_dialog.steam_workshop_db_github_url.setCursorPosition(0)
+        self.settings_dialog.database_expiry.setText(str(self.settings.database_expiry))
 
         # Sorting tab
         if self.settings.sorting_algorithm == SortMethod.ALPHABETICAL:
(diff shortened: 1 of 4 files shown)


hk21702 (changes requested):
Seems good. Do me a favor and update from main to get the new settings model, and then reorder the corresponding setting so that it matches the new organization.

LionelColaso:
> Seems good. Do me a favor and update from main to get the new settings model, and then reorder the corresponding setting so that it matches the new organization.

yeah sure , also would it be good idea to add a toogle to enable disable this notification time ?

app/controllers/settings_controller.py
```diff
@@ -422,6 +421,9 @@ def _update_view_from_model(self) -> None:
             self.settings.external_steam_metadata_repo
         )
         self.settings_dialog.steam_workshop_db_github_url.setCursorPosition(0)
+        self.settings_dialog.db_builder_database_expiry.setText(
```
hk21702, line 424: this should probably be renamed for clarity

hk21702 (changes requested):

app/controllers/settings_controller.py
LionelColaso: ok done should be good now

LionelColaso:
All should be fine now

oceancabbage (approved):

hk21702 (approved):

PR #482, opened 2024-07-14
Add Optional Settings to "Use Mods Dependency Rules For Sorting"

LionelColaso:
- [x] Add setting to use Mods Dependency in the About file as Sorting Rule

- [ ] when enabled uses dependencies (mod required) to load_before the mod having it as a dependency in about.xml

Screenshot 2024-07-14 055405

### DIFF
--- a/app/controllers/settings_controller.py
+++ b/app/controllers/settings_controller.py
@@ -599,6 +599,10 @@ class SettingsController(QObject):
         elif self.settings.sorting_algorithm == SortMethod.TOPOLOGICAL:
             self.settings_dialog.sorting_topological_radio.setChecked(True)
 
+        # Use dependencies for sorting checkbox
+        if self.settings.use_dependency_for_sorting:
+            (self.settings_dialog.use_dependency_for_sorting_checkbox.setChecked(True))
+
         # Set dependencies checkbox
         self.settings_dialog.check_deps_checkbox.setChecked(
             self.settings.check_dependencies_on_sort
@@ -800,6 +804,11 @@ class SettingsController(QObject):
             self.settings.sorting_algorithm = SortMethod.ALPHABETICAL
         elif self.settings_dialog.sorting_topological_radio.isChecked():
             self.settings.sorting_algorithm = SortMethod.TOPOLOGICAL
+
+        # Use dependencies for sorting checkbox
+        self.settings.use_dependency_for_sorting = (
+            self.settings_dialog.use_dependency_for_sorting_checkbox.isChecked()
+        )
 
         # Set dependencies checkbox
         self.settings.check_dependencies_on_sort = (
--- a/app/views/settings_dialog.py
+++ b/app/views/settings_dialog.py
@@ -468,6 +468,15 @@ class SettingsDialog(QDialog):
         self.sorting_topological_radio = QRadioButton("Topologically")
         sort_group_box_layout.addWidget(self.sorting_topological_radio)
 
+        # Use dependencies for sorting checkbox
+        self.use_dependency_for_sorting_checkbox = QCheckBox(
+            "Use dependency rules for sorting."
+        )
+        self.use_dependency_for_sorting_checkbox.setToolTip(
+            "If enabled, Mods will be sorted such that dependencies are loaded before the dependent mod."
+        )
+        sort_group_box_layout.addWidget(self.use_dependency_for_sorting_checkbox)
+
         # Dependencies group
         deps_group_box = QGroupBox("Sort Dependencies")
         tab_layout.addWidget(deps_group_box)
--- a/app/models/settings.py
+++ b/app/models/settings.py
@@ -67,6 +67,7 @@ class Settings(QObject):
         self.check_dependencies_on_sort: bool = (
             True  # Whether to check for missing dependencies when sorting
         )
+        self.use_dependency_for_sorting: bool = False
 
         # DB Builder
         self.db_builder_include: str = "all_mods"


app/controllers/settings_controller.py
```diff
@@ -570,6 +571,10 @@ def _update_model_from_view(self) -> None:
             self.settings.sorting_algorithm = "Alphabetical"
         elif self.settings_dialog.sorting_topological_radio.isChecked():
             self.settings.sorting_algorithm = "Topological"
+        if self.settings_dialog.db_use_dependency_for_sorting_checkbox.isChecked():
+            self.settings.dependency_for_sorting = "True"
+        else:
+            self.settings.dependency_for_sorting = "False"
```
hk21702, line 577: I'd just do:

```suggestion
self.settings.dependency_for_sorting = self.settings_dialog.db_use_dependency_for_sorting_checkbox.isChecked()
```
instead since you're storing the setting as a bool

app/controllers/settings_controller.py
```diff
@@ -428,6 +427,8 @@ def _update_view_from_model(self) -> None:
             self.settings_dialog.sorting_alphabetical_radio.setChecked(True)
         elif self.settings.sorting_algorithm == "Topological":
             self.settings_dialog.sorting_topological_radio.setChecked(True)
+        if self.settings.dependency_for_sorting == "True":
```
hk21702, line 430: Just do
```suggestion
        if self.settings.dependency_for_sorting:
```
since its a bool in the settings model

app/utils/metadata.py
```diff
@@ -612,6 +612,33 @@ def compile_metadata(self, uuids: list[str] = []) -> None:
                         dependencies,
                         self.internal_local_metadata,
                     )
+                    # Cleanup Dependencies so that it can be used as load order rules
+                    if (
+                        self.settings_controller.settings.dependency_for_sorting
+                        == "True"
+                    ):
```
hk21702, line 619:
```suggestion
                    if self.settings_controller.settings.dependency_for_sorting:
```

app/utils/metadata.py
```diff
@@ -635,6 +662,33 @@ def compile_metadata(self, uuids: list[str] = []) -> None:
...
+                                )
+                                logger.debug(load_dependencies_by_ver)
+                                add_load_rule_to_mod(
+                                    self.internal_local_metadata[uuid],
+                                    load_dependencies_by_ver,
+                                    "loadTheseBefore",
+                                    "loadTheseAfter",
+                                    self.internal_local_metadata,
+                                    self.packageid_to_uuids,
+                                )
```
hk21702, line 691: Duplicated code. Extract to a private method

app/views/settings_dialog.py
```diff
@@ -487,6 +487,12 @@ def _do_sorting_tab(self) -> None:
         explanatory_label.setWordWrap(True)
         tab_layout.addWidget(explanatory_label)
 
+        # Checkboxes
+        self.db_use_dependency_for_sorting_checkbox = QCheckBox(
+            "Use Mods Dependency Rules For Sorting"
```
hk21702, line 492: A little unclear. I'd add some more context to ensure that it is clear.

Something like, "Use mod dependency rules for sorting such that dependencies are loaded before the dependent mod."

hk21702 (changes requested):
See the specific comments.

Namely, better use of bool nature of the settings model, a private method to avoid duplicate code, and a better description of what the setting does.

app/utils/metadata.py
LionelColaso: Thanks for Suggesstions
This is the actual part of code that can be improved and i am clueless to how
thats the reason the code is duplicated

LionelColaso: ok now can you check

app/utils/metadata.py
```diff
@@ -568,506 +568,280 @@ def __update_from_settings(self) -> None:
...
+                    self.packageid_to_uuids,
+                )
+
+        for uuid in uuids:
+            mod_metadata = self.internal_local_metadata[uuid]
+            logger.debug(f"UUID: {uuid} packageid: {mod_metadata.get('packageid')}")
+
+            # Process moddependencies
+            dependencies = mod_metadata.get("moddependencies")
+            if dependencies:
```
oceancabbage, line 639: Minor thing, and I know we didn't have this before, but as an overall improvement to RimSort we should start being explicit with conditionals involving `None`s. Since `dependencies` could be anything here, including an empty string, empty list, etc, `if dependencies...` could resolve to `True` even though we got something erroneous back.

This is common across RimSort but we should try to change it

oceancabbage (changes requested):

app/utils/metadata.py
```diff
@@ -568,506 +568,280 @@ def __update_from_settings(self) -> None:
...
+                        .split(",")[0]
+                        .replace("{'packageid': '", "")
+                        .replace("[{'packageid' :", "")
+                        .replace("displayname': '", "")
+                        .replace("[", "")
+                        .replace("'", "")
+                        .replace("{", "")
+                        .replace("}", "")
+                        .replace("li:", "")
                     )
```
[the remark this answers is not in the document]
LionelColaso: i think i fixed it now with rewrite

app/utils/metadata.py
LionelColaso: i think i fixed it now with rewrite

LionelColaso:
Hope I addressed all the issues please review

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
-                    if match(version_regex, version):
-                        if (
-                            dependencies_by_ver
-                            and isinstance(dependencies_by_ver, dict)
-                            and dependencies_by_ver.get("li")
-                        ):
-                            logger.debug(
-                                f"Current mod requires these mods by version to work: {dependencies_by_ver['li']}"
-                            )
+        def process_dependencies(uuid, dependencies):
```
hk21702, line 579: This needs to be typed

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
+            """
+            Process dependencies for a given mod UUID.
+            """
+            if dependencies and isinstance(dependencies, list):
```
hk21702, line 583: If the method is only supposed to take a list, I would not use a check like this. It can cause weird bugs and behavior and makes the method more difficult to use. Just set the param types and rely on static type checking.

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
-                    if load_these_after:
-                        logger.debug(
-                            f"Current mod should load before these mods: {load_these_after}"
+                    else:
+                        logger.warning(
+                            f"Illegal dependency format for mod {uuid}: {dependency}"
                         )
-                        add_load_rule_to_mod(
+
+        def process_incompatibilities(uuid, incompatibilities):
```
hk21702, line 611: Needs to be typed.

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
+                            f"Illegal incompatibility format for mod {uuid}: {incompatibility}"
                         )
-                except Exception as e:
-                    mod_metadata_path = self.internal_local_metadata[uuid][
-                        "metadata_file_path"
-                    ]
-                    logger.warning(
-                        f"About.xml syntax error. Unable to read <forceloadbefore> tag from XML: {mod_metadata_path}"
+
+        def process_load_order(uuid, load_these, rule_type, opposite_rule_type):
```
hk21702, line 628: Needs to be typed

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
+                        logger.warning(
+                            f"Illegal dependency format for mod {uuid}: {dependency}"
                         )
-                        add_load_rule_to_mod(
+
+        def process_incompatibilities(uuid, incompatibilities):
+            """
+            Process incompatibilities for a given mod UUID.
+            """
+            if incompatibilities and isinstance(incompatibilities, list):
```
hk21702, line 615: Just use static type checking if the param is supposed to only be one type

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
-                        "metadata_file_path"
-                    ]
-                    logger.warning(
-                        f"About.xml syntax error. Unable to read <forceloadbefore> tag from XML: {mod_metadata_path}"
+
+        def process_load_order(uuid, load_these, rule_type, opposite_rule_type):
+            """
+            Process load order rules for a given mod UUID.
+            """
+            if load_these and isinstance(load_these, list):
```
hk21702, line 632: Just use static type checking if the param is supposed to only be one type

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
+            force_load_before_list = (
+                force_load_before.get("li") if force_load_before else None
+            )
+            process_load_order(
+                uuid, force_load_before_list, "loadTheseAfter", "loadTheseBefore"
+            )
+
+            # Process loadbeforebyversion
+            load_before_by_version = mod_metadata.get("loadbeforebyversion", {})
+            if load_before_by_version and isinstance(load_before_by_version, dict):
```
hk21702, line 737: Just use static type checking if the param is supposed to only be one type

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
             logger.info("No Steam database supplied from external metadata. skipping.")
-        # Add load order to installed mods based on dependencies from community rules
-        if self.external_community_rules:
-            logger.info("Started compiling metadata from configured Community Rules")
-            for package_id in self.external_community_rules:
-                # Note: requiring the package be in self.internal_local_metadata should be fine, as
-                # if the mod doesn't exist self.internal_local_metadata, then either mod_data or dependency_id
-                # will be None, and then we don't insert a dependency
+
+        def process_external_rules(rules, rule_type, opposite_rule_type):
```
hk21702, line 809: typing

app/utils/metadata.py
```diff
@@ -569,506 +569,309 @@ def __update_from_settings(self) -> None:
...
-                        "loadAfter"
-                    )
-                    if load_these_before:
-                        logger.debug(
-                            f"Current mod should load after these mods: {load_these_before}"
-                        )
-                        # In Alphabetical, load_these_before is at least an empty dict
-                        for load_this_before in load_these_before:
+                    load_these = rule_data.get(rule_type)
+                    if load_these and isinstance(load_these, list):
```
hk21702, line 816: Just use static type checking if the param is supposed to only be one type

app/views/settings_dialog.py
```diff
@@ -497,6 +497,12 @@ def _do_sorting_tab(self) -> None:
         explanatory_label.setWordWrap(True)
         tab_layout.addWidget(explanatory_label)
 
+        # Checkboxes
+        self.db_use_dependency_for_sorting_checkbox = QCheckBox(
+            "Use mod dependency rules for sorting such that dependencies are loaded before the dependent mod"
```
hk21702, line 502: "Use dependency rules for sorting."

Then add a tooltip for more context: "Loads defined dependencies before the dependent mod."

app/utils/metadata.py
hk21702, line 1: You are using add_load_rule_to_mod incorrectly somewhere. Particularly for the dependency_or_dependency_ids arg. Sometimes, dicts which are not dependencies get passed to it. For example, "{'comment': 'RocketMan should always be loaded last. It is known.', 'value': True}" and "{'value': True}" and an error is logged.

hk21702 (changes requested):
See comments.

Also, your RImSort.wiki module commit is old and older than main as far as I can tell.

app/utils/metadata.py
LionelColaso: Fixed

app/utils/metadata.py
LionelColaso: Fixed

app/utils/metadata.py
LionelColaso: Fixed

app/utils/metadata.py
LionelColaso: Fixed

app/utils/metadata.py
LionelColaso: Fixed

app/utils/metadata.py
LionelColaso: Fixed

app/utils/metadata.py
LionelColaso: Fixed

app/views/settings_dialog.py
LionelColaso: Fixed

app/utils/metadata.py
LionelColaso: Sorry I dont understand what u want to convey can you elobarate a bit 
since i tested the changes with some initial 20 mods which had dependencies and later with 450+ mods and everything seems to be working as expected 
Roctetman which has to be loaded last is always on the end of the list
so i am confused to what exactly is wwrong ?

Silly of me for missing typing (i am still getting used to it)

Everything else should now be fixed as per your suggesstions
Thankyou

PR #487, opened 2024-07-15
Refactor menu bar

oceancabbage:
Refactored `menu_bar.py`
- individual menus are separated into their own method
- single menu creation method for both mac and non-mac
- settings and quit actions are on the App menu for mac and on the File menu for non-mac
- on mac, moved update stuff to Help menu to be consistent with non-mac

Tested on Mac and Windows

### DIFF
--- a/app/controllers/menu_bar_controller.py
+++ b/app/controllers/menu_bar_controller.py
@@ -21,17 +21,18 @@ class MenuBarController(QObject):
 
         self.menu_bar.quit_action.triggered.connect(QApplication.instance().quit)
 
-        self.menu_bar.check_for_updates_action.triggered.connect(
-            EventBus().do_check_for_application_update.emit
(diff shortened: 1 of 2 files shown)


hk21702 (approved):
Seems good on Windows. Will need someone to check for MacOS.

Out of curiosity though, what are the cut, copy, and paste buttons for anyways? Also, should settings be under edit instead of file?

oceancabbage:
Thanks for reviewing! No idea what the cut, copy, paste buttons do. They were present before the refactor and in case they do something I left them there for now. I don't mind where Settings goes, File or Edit both work if anyone has opinions.

oceancabbage:
Cut, copy, and paste do have attached functions, so leaving them there for now. Commented out update buttons as they are no-ops.

Issue #436, opened 2024-06-21
Suggestion: Add "open log file" button to menubar and settings


Limfjorden:
**Is your feature request related to a problem? Please describe.**
Currently I have to venture into the file explorer to find the log files. This takes time, and my only other choice is to upload them to the internet.

**Describe the solution you'd like**
It would be easy and convienient to have buttons in RimSort for opening `RimSort.log` and `Rimworld.log`. I think they should be in the "File" menu in the menu bar next to "Upload log". There should also be buttons in the settings menu called "locations" that opens the files.

**Describe alternatives you've considered**

**Additional context**
Tag: Accessibilty

Raizanad:
Call it shortcuts as a drop down menu.
Shortcuts included should be:
SteamCMD
Steam Workshop Mods folders
Local Mods folder
Game mod folder
Root directory folder for the game
Root directory for rimsort
Rimsort user files (as per my other feature request)
Rimworld appdata root folder.

oceancabbage:
Addressing through https://github.com/RimSort/RimSort/pull/493

PR #493, opened 2024-07-16
[436] adding shortcuts for opening useful directories

oceancabbage:
Changes for https://github.com/RimSort/RimSort/issues/436
Adds four new menu buttons, under `File -> Shortcuts`
- Open the app directory (where RimSort is)
- Open the directory RimSort keeps settings files in (and DB and instances)
- Open the RimSort logs directory
- Open the RimWorld logs directory

Did not add shortcuts for the following because they're easily reachable via settings panel:
- SteamCMD
- Steam Workshop Mods folders
- Local Mods folder
- Game mod folder

Tested on Windows and Mac

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -20,6 +20,9 @@ from zipfile import ZipFile
 
 from loguru import logger
 
+from app.utils.generic import platform_specific_open
+from app.utils.system_info import SystemInfo
+
 # GitPython depends on git executable being available in PATH
 try:
     from git import Repo
@@ -57,7 +60,6 @@ from app.utils.generic import (
     delete_files_except_extension,
     launch_game_process,
     open_url_browser,
-    platform_specific_open,
     upload_data_to_0x0_st,
 )
 from app.utils.metadata import MetadataManager, SettingsController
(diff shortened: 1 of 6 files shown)


app/views/main_content_panel.py
```diff
@@ -17,6 +17,9 @@
 from typing import Any, Callable, Self
 from urllib.parse import urlparse
 from zipfile import ZipFile
+from pathlib import Path
+from app.utils.generic import platform_specific_open
+from app.utils.system_info import SystemInfo
 
 from loguru import logger
```
hk21702, line 24: Needs to be auto-sorted with Ruff

app/views/main_content_panel.py
```diff
@@ -1604,6 +1619,43 @@ def _do_upload_list_rentry(self) -> None:
...
+
+    def _do_open_settings_directory(self) -> None:
+        settings_directory = AppInfo().app_storage_folder
+        logger.info(f"Opening settings directory: {settings_directory}")
+        platform_specific_open(settings_directory)
+
+    def _do_open_rimsort_logs_directory(self) -> None:
+        logs_directory = AppInfo().user_log_folder
+        logger.info(f"Opening RimSort logs directory: {logs_directory}")
+        platform_specific_open(logs_directory)
```
hk21702, line 1635: This is fine for now, but should move away from using AppInfo where possible. In its current state, it forces a weird non-pep approved import and init order of the modules due to weird dependency chains.

hk21702 (changes requested):
Tested on windows and seems to work. Just need to format that one import block in main_content_panel.py (Think menu_bar.py needs to be formatted too)

app/views/main_content_panel.py
oceancabbage: Oo, I thought `ruff format` would do this. Fixed using `ruff check --fix`

app/views/main_content_panel.py
oceancabbage: Thanks for pointing that out, will look into it going forward

hk21702 (approved):

Issue #483, opened 2024-07-14
Suggestion: Tooltips for Filter Mode Icons


hk21702:
Add tooltips for filter mode icons.

![image](https://github.com/user-attachments/assets/28e9e5f7-ec66-4f8f-9cb7-19ca6f887b6f)

1) RimSort icon(black) = All Mods
2) Orange = Official Content (Core and DLC)
3) Dark blue Folder Icon = Local Mods (Including SteamCMD and Git)
4) Git Icon = Git Mods
5) Cloud with Steam icon = SteamCMD Mods
6) Steam Icon = Steam Mods

oceancabbage:
Addressed in https://github.com/RimSort/RimSort/pull/494

PR #494, opened 2024-07-16
[483] tooltips for filter mode icons

oceancabbage:
Adding tooltips for active and inactive mod list filter mode icons
For https://github.com/RimSort/RimSort/issues/483


hk21702 (approved):
Looks good. There are some MyPy issues that could be cleaned up, but they're for the most part unrelated to the PR.

PR #497, opened 2024-07-17
Typing fixes and Cleanup

hk21702:
A bunch of code cleanup, particularly typing fixes related to MyPy. There are also some fixes for Ruff and jscpd errors.

There may be some things related to steam that breaks due to some changes so it needs to be tested. No behaviors are changed intentionally.


hk21702:
Down to 186 errors. We've passed 200!

hk21702:
Is multibutton.py and proxy_style.py used for anything? Can they be removed? They both seem potentially broken, and are also just not utilized as far as I can tell.

LionelColaso (approved):
Nice

PR #499, opened 2024-07-17
Minor Setting Changes

LionelColaso:
1) Rearranged some Buttons
2) Added Warning Texts to Build Database Tab

### DIFF
--- a/app/controllers/settings_controller.py
+++ b/app/controllers/settings_controller.py
@@ -133,7 +133,7 @@ class SettingsController(QObject):
         )
 
         # Wire up the Databases tab buttons
-
+        # Community Database Rules buttons
         self.settings_dialog.community_rules_db_none_radio.clicked.connect(
             self._on_community_rules_db_radio_clicked
         )
@@ -143,7 +143,6 @@ class SettingsController(QObject):
         self.settings_dialog.community_rules_db_local_file_radio.clicked.connect(
             self._on_community_rules_db_radio_clicked
         )
-
         self.settings_dialog.community_rules_db_local_file_choose_button.clicked.connect(
             self._on_community_rules_db_local_file_choose_button_clicked
         )
@@ -154,6 +153,7 @@ class SettingsController(QObject):
             EventBus().do_download_community_rules_db_from_github
         )
 
+        # Steam Workshop Database buttons
         self.settings_dialog.steam_workshop_db_none_radio.clicked.connect(
             self._on_steam_workshop_db_radio_clicked
         )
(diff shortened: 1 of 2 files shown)


app/controllers/settings_controller.py
```diff
@@ -1044,6 +1052,20 @@ def _on_community_rules_db_local_file_choose_button_clicked(self) -> None:
...
 
+    @Slot()
+    def _on_community_rules_db_location_open_button_clicked(self) -> None:
+        """
+        Open the community rules database Folder Location.
+        """
+        community_rules_db_location = (
+            self.settings.external_community_rules_file_path.replace(
+                "communityRules.json", "dbs/Community-Rules-Database"
+            )
```
oceancabbage, line 1063: For the DB location, it seems you are determining the parent folder via assuming there is a `communityRules.json`. But, isn't it possible for the user to choose a `.json` file that isn't named `communityRules`? In that case, there are community rules but this string replacement logic doesn't work as intended.

To find the db location I would use path logic instead using `os`.

app/controllers/settings_controller.py
```diff
@@ -1044,6 +1052,20 @@ def _on_community_rules_db_local_file_choose_button_clicked(self) -> None:
...
+        )
+        if not community_rules_db_location:
+            return
+        platform_specific_open(Path(community_rules_db_location))
```
oceancabbage, line 1067: If the user doesn't have community rules enabled, e.g. selected `None` in settings, this button is enabled and clicking it does nothing (except log to the console). Would we want to only enable the button when a local community rules is being used? Why or why not?

If no, should a warning pop up anyway if the user clicks the button but no folder could be opened? Just thinking that from a UX perspective a failure to open the folder is not shown in any way.

app/controllers/settings_controller.py
```diff
@@ -1117,6 +1139,33 @@ def _on_steam_workshop_db_local_file_choose_button_clicked(self) -> None:
...
+        Open the Steam workshop database Folder Location.
+        """
+        steam_workshop_db_location = (
+            self.settings.external_steam_metadata_file_path.replace(
+                "steamDB.json", "dbs/Steam-Workshop-Database"
+            )
+        )
+        if not steam_workshop_db_location:
+            return
+        print(steam_workshop_db_location)
```
oceancabbage, line 1154: Likely not needed

oceancabbage (changes requested):
Looks good but a few points of discussion I would like to have with you

app/controllers/settings_controller.py
LionelColaso: ok i will do a separate pull for this

LionelColaso:
Should be Good now

oceancabbage (approved):

PR #503, opened 2024-07-19
fix for TypeError: 'IterableList' object is not callable

LionelColaso:
image

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -105,9 +105,7 @@ class MainContent(QObject):
             cls._instance = super(MainContent, cls).__new__(cls)
         return cls._instance
 
-    def __init__(
-        self, settings_controller: SettingsController
-    ) -> None:
+    def __init__(self, settings_controller: SettingsController) -> None:
         """
         Initialize the main content panel.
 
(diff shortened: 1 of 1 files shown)


app/views/main_content_panel.py
```diff
@@ -2169,20 +2169,27 @@ def _check_git_repos_for_update(self, repo_paths: list[str]) -> None:
...
-                        refs = repo.refs()
-                        remote_ref = refs[f"origin/{local_ref.name}"]
+                        remote_ref = None
+                        for ref in repo.refs:
+                            if f"origin/{local_ref.name}" in ref.path:
+                                remote_ref = ref
+                                break
+
+                        if remote_ref is None:
+                            logger.warning("Remote reference not found for the local branch.")
```
oceancabbage, line 2179: Minor: thoughts on slightly more specific logging here? e.g. `logger.warning(f"Remote reference not found for the local branch '{local_ref.name}'.")`

app/views/main_content_panel.py
```diff
@@ -2169,20 +2169,27 @@ def _check_git_repos_for_update(self, repo_paths: list[str]) -> None:
 
                         # Get the local and remote refs
                         local_ref = repo.head.reference
-                        refs = repo.refs()
-                        remote_ref = refs[f"origin/{local_ref.name}"]
+                        remote_ref = None
+                        for ref in repo.refs:
+                            if f"origin/{local_ref.name}" in ref.path:
+                                remote_ref = ref
+                                break
```
oceancabbage, line 2176: (no changes needed) equivalent one-liner for (debatably) more readable code:
```
remote_ref = next((ref for ref in repo.refs if f"origin/{local_ref.name}" in ref.path), None)
```

oceancabbage (changes requested):
Looks good, perhaps a minor change to log more specifically when `remote_ref is None` as this will help with future debugging

app/views/main_content_panel.py
LionelColaso: Thanks Should be good now

app/views/main_content_panel.py
```diff
@@ -2168,8 +2168,14 @@ def _check_git_repos_for_update(self, repo_paths: list[str]) -> None:
...
+                            (
+                                ref
+                                for ref in repo.refs
+                                if f"origin/{local_ref.name}" in ref.path
+                            ),
+                            None,
+                        )
 
                         # Check if the local branch is behind the remote branch
                         if local_ref.commit != remote_ref.commit:
```
oceancabbage, line 2181: Why was the logging removed from here? I think it'd be useful to know when `remote_ref` is `None`.

Also, when `remote_ref` is `None`, the conditional `if local_ref.commit != remote_ref.commit:` will fail because `None` has no attribute `.commit`. I think you have to check the existence of `remote_ref` first, e.g. `if remote_ref is not None`.

oceancabbage (changes requested):

app/views/main_content_panel.py
LionelColaso: hopefully i did not miss anything now

oceancabbage:
Does this need to be updated with changes from main first? When I pull the PR to review locally the code is not the same as what is here

LionelColaso:
> Does this need to be updated with changes from main first? When I pull the PR to review locally the code is not the same as what is here

the below command shoulld fix the issue 
git pull --rebase -f

twstagg:
Can you please provide steps to reproduce this issue on `HEAD`?

PR #516, opened 2024-07-21
Set the default sorting algorithm to be Topological

hk21702:
Sets the default sorting algorithm to be Topological instead of Alphabetical. This change has been in discussion for a while.

The reason for making this change is twofold. Currently, Alphabetical is bugged in the sense that it does not respect certain rules when sorting. Additionally, topological as an algorithm makes more sense and is easier to maintain.

The downside is that alphabetical has been the norm for a long time. The result of this is that there may be alphabetical based sorting results that incidentally solved load order issues, but are not explicitly defined in about.xml or the community rules database. To be clear, this is not a flaw of the sorting algorithm itself. These are flaws in the rules defined that were missed.

To be clear, if someone has used RimSort already, and has a valid settings file, they will not experience any change. Only freshly new users as well as those that reset their settings to default will notice a change.

### DIFF
--- a/app/models/settings.py
+++ b/app/models/settings.py
@@ -44,7 +44,7 @@ class Settings(QObject):
         self.database_expiry: int = 604800  # 7 days
 
         # Sorting
-        self.sorting_algorithm: SortMethod = SortMethod.ALPHABETICAL
+        self.sorting_algorithm: SortMethod = SortMethod.TOPOLOGICAL
 
         # DB Builder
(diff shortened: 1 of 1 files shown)


oceancabbage (approved):
With inconsistencies in mod XMLs and the brittle code behind alphabetical, I'm not even sure how much value alphabetical provides anymore (if any); I wonder how many people have issues with mod load order solely because they use alphabetical rather than topological sort.

Anyway, thanks for switching it over

PR #517, opened 2024-07-21
Refactor app_info.py initialization

hk21702:
Refactors the app_info.py initialization process such that it no longer takes a main_file arg. Instead, it looks up the `__main__` module and gets the correct file path to later get the application folder. This makes it such that the app_info singleton can be initialized at any time and by any module. As a consequence, it does not need to have special handling in app.__main__, resolving the long-standing import order issues.

### DIFF
--- a/app/utils/app_info.py
+++ b/app/utils/app_info.py
@@ -1,3 +1,4 @@
+import sys
 from pathlib import Path
 from typing import Optional
 
@@ -13,14 +14,13 @@ class AppInfo:
     using the `platformdirs` package, ensuring platform-specific conventions are adhered to.
 
     Examples:
-        >>> app_info = AppInfo(__file__)
         >>> print(app_info.app_name)
         >>> print(app_info.app_storage_folder)
     """
 
     _instance: Optional["AppInfo"] = None
 
-    def __new__(cls, main_file: Optional[str] = None) -> "AppInfo":
+    def __new__(cls) -> "AppInfo":
         """
         Create a new instance or return the existing singleton instance of the `AppInfo` class.
         """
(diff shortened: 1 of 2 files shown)


app/utils/app_info.py
```diff
@@ -13,36 +16,34 @@ class AppInfo:
...
-            main_file (Optional[str]): Path to the main application file (i.e., __file__ from __main__).
-
         Raises:
-            ValueError: If `main_file` is not provided during the first initialization.
+            Exception: If the main file path cannot be determined.
         """
         if hasattr(self, "_is_initialized") and self._is_initialized:
             return
 
+        main_file = sys.modules["__main__"].__file__
```
oceancabbage, line 43: No changes needed: there are cases where `__main__` does not exist (e.g. interactive interpreter or embedded Python environment). Wrapping this with a try/except for `KeyError, AttributeError` would make safer

app/utils/app_info.py
```diff
@@ -55,8 +56,17 @@ def __init__(self, main_file: Optional[str] = None) -> None:
         # Application metadata
 
         self._app_name = "RimSort"
-        self._app_version = ""
         self._app_copyright = ""
+        
+        self._app_version = "Unknown version"
+        version_file = str(self._application_folder / "version.xml")
```
oceancabbage, line 62: No changes needed: if these strings are used elsewhere, consider having these as constants or moving config values to a separate configuration file or module.

oceancabbage (approved):
Looks great, thanks for solving this critical issue

app/utils/app_info.py
hk21702: So main here refers to the literal file that is called "__main__" that acts as the entry point for the entire app, not just this specific file. If that module doesn't exist, then the entire app isn't goimg to work anyways. I tested in interpreted mode albeit with the normal app entry point and it seems fine.

Issue #500, opened 2024-07-18
Suggestion: Improved Crash Dialogue


hk21702:
Improve the crash dialogue

- [ ] Improved formatting so that the dialogue is less squished
- [ ] Add button to open log folder
- [ ] Add button to attempt to upload the most recent log

PR #520, opened 2024-07-22
[500] Dialogue improvements

hk21702:
Moves dialogue to be under view instead of model.

Also implements #500 via a new custom dialog.
On fatal error, the user is presented with two additional options. Open the directory where RimSort logs live, and to upload the latest logs to 0x0.st. Spacing was between the elements was also improved. All previous functionality such as hidden details tabs was maintained.

Upload to 0x0.st is threaded and a separate dialog with a working bar shows up while waiting for a response from 0x0.st. 
Upon successful upload, the user is provided with the link, and the link is copied to their clipboard. If failure, the user is informed. 

After either action, opening log dir and upload, the fatal error dialog remains in case the user wishes to take additional actions.

image

image

image


hk21702:
Some of the upload code could go into a controller, but we aren't using a dialog controller currently

app/utils/generic.py
```diff
@@ -17,7 +17,7 @@
 )
 from requests import post as requests_post
 
-from app.models.dialogue import show_fatal_error, show_warning
+import app.views.dialogue as dialogue
```
oceancabbage, line 20: Minor: for consistency, is there a reason to not import the methods that are used here rather than `dialogue` itself? Other files import the methods

app/views/dialogue.py
```diff
@@ -223,45 +237,195 @@ def show_warning(
...
+    """Custom dialog to display fatal errors.
+
+    Has button to show more details, open the log directory, and upload the log file to 0x0.
+    """
+
+    def __init__(
+        self,
+        title: str = "Fatal Error",
+        text: str = "A fatal error has occurred!",
+        information: str = "Please report the error to the developers. Lorem ipsum asdlkfj asldkfj asd;lkj as;dlkfjlasdkjf asdlfkja sdflkasjdf sadl;kjas;ldf sa;dlkjslad fasl;dkfj as;dlfjasld;fjas",
```
oceancabbage, line 273: Forgot to take out the filler text? :)

oceancabbage (changes requested):
Tested on my system, looks awesome. Just two minor comments

app/utils/generic.py
hk21702: Was a little hack to prevent circular dependency I found. I forgot with what it was, I think generic. If both modules don't use each other on just loading, then you avoid circular dep error if you import the module itself instead of its components.

app/views/dialogue.py
hk21702: Woops

hk21702:
Added functionality to swap the show details btn text between "Show" and "Hide" depending on the state.
Added some comments to fix jscpd error that was triggering on docstring.
Fixed merge conflicts with main.

oceancabbage (approved):
Fantastic, thanks!

PR #525, opened 2024-07-23
Set application window icon for Windows

hk21702:
Sets the application window icon so that the windows don't use a placeholder icon.


oceancabbage (commented):
On macOS (ARM), the app icon is still the magnifying glass when building, but from interpreter it is the new icon

hk21702:
@oceancabbage Oh right. I only changed it for windows. Mostly I wanted to make sure the change doesn't crash other platforms. Should I swap to the alt version for mac too? Will need to make a new icns too since I think thats what mac uses.

oceancabbage:
> @oceancabbage Oh right. I only changed it for windows. Mostly I wanted to make sure the change doesn't crash other platforms. Should I swap to the alt version for mac too? Will need to make a new icns too since I think thats what mac uses.

It's okay let's just do windows for this PR

oceancabbage (approved):

PR #526, opened 2024-07-23
Temporarily force windows console as workaround Nuitka bug 3019

hk21702:
Temporarily tell Nuitka to force enable the console on Windows. This is a temporary workaround for Nuitka issue 3019 so we can avoid downgrading Nuitka and messing with the Nuitka action.


oceancabbage (approved):
I appreciate the thorough investigation and quick fix

PR #532, opened 2024-07-25
[501] handle malformed settings with dialog

oceancabbage:
If the settings file cannot be parsed, present a settings parse failure dialog, allowing the user to perform 1 of 4 actions:
1. Open the settings file
2. Open the settings file location
3. Reset the settings file
4. Close the application (by clicking on window X)

If the user resets the settings file, the startup sequence continues (this is done through a while loop). See image below.

Screenshot 2024-07-24 at 10 29 29 PM

Includes a bunch of ruff formats probably covered in https://github.com/RimSort/RimSort/pull/530 so we should merge that first and rebase this one.


LionelColaso:
Shouldnt this also tajke care of invalid steamcmd paths?

https://0x0.st/Xf-8.log

image

  File "D:\GitHub\RimSort\app\__main__.py", line 96, in main_thread
    app_controller = AppController()
                     ^^^^^^^^^^^^^^^
  File "D:\GitHub\RimSort\app\controllers\app_controller.py", line 49, in __init__
    self.steamcmd_wrapper = SteamcmdInterface.instance(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 99, in instance
    cls._instance = cls(*args, **kwargs)
                    ^^^^^^^^^^^^^^^^^^^^
  File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 44, in __init__
    self.initialize_prefix(steamcmd_prefix, validate)
  File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 77, in initialize_prefix
    os.makedirs(self.steamcmd_install_path)
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 225, in makedirs
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'E:\\'

oceancabbage:
> This does not take care of invalid steamcmd paths?
> 
> https://0x0.st/Xf-8.log
> 
> image
> 
> Traceback (most recent call last): File "D:\GitHub\RimSort\app__main__.py", line 96, in main_thread app_controller = AppController() ^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\controllers\app_controller.py", line 49, in **init** self.steamcmd_wrapper = SteamcmdInterface.instance( ^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 99, in instance cls._instance = cls(*args, **kwargs) ^^^^^^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 44, in **init** self.initialize_prefix(steamcmd_prefix, validate) File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 77, in initialize_prefix os.makedirs(self.steamcmd_install_path) File "", line 215, in makedirs File "", line 215, in makedirs File "", line 215, in makedirs File "", line 225, in makedirs FileNotFoundError: [WinError 3] The system cannot find the path specified: 'E:\'

It does not, in its current state this PR only makes it so that the dialog appears on JSON decode error. For cases where values for certain keys in the settings file are "invalid", I don't think it's possible to assume that the settings file is corrupt and present the same dialog.

In your example, what if the user does have an external E drive and that was the intended steamCMD install path, but then forgot to plug in the drive and ran RimSort? There's nothing malformed about the settings file, just an unexpected error with how the key/values are being used. In these cases I think a FatalError dialog is appropriate.

Thoughts, @hk21702 ?

LionelColaso:
> > This does not take care of invalid steamcmd paths?
> > https://0x0.st/Xf-8.log
> > image
> > Traceback (most recent call last): File "D:\GitHub\RimSort\app__main__.py", line 96, in main_thread app_controller = AppController() ^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\controllers\app_controller.py", line 49, in **init** self.steamcmd_wrapper = SteamcmdInterface.instance( ^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 99, in instance cls._instance = cls(*args, **kwargs) ^^^^^^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 44, in **init** self.initialize_prefix(steamcmd_prefix, validate) File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 77, in initialize_prefix os.makedirs(self.steamcmd_install_path) File "", line 215, in makedirs File "", line 215, in makedirs File "", line 215, in makedirs File "", line 225, in makedirs FileNotFoundError: [WinError 3] The system cannot find the path specified: 'E:'
> 
> It does not, in its current state this PR only makes it so that the dialog appears on JSON decode error. For cases where values for certain keys in the settings file are "invalid", I don't think it's possible to assume that the settings file is corrupt and present the same dialog.
> 
> In your example, what if the user does have an external E drive and that was the intended steamCMD install path, but then forgot to plug in the drive and ran RimSort? There's nothing malformed about the settings file, just an unexpected error with how the key/values are being used. In these cases I think a FatalError dialog is appropriate.
> 
> Thoughts, @hk21702 ?

yeah right i think this should be addressed just like how invalid game paths are handled i guess this is a separate issue
image

LionelColaso (approved):
Looks Good

hk21702:
> > This does not take care of invalid steamcmd paths?
> > https://0x0.st/Xf-8.log
> > image
> > Traceback (most recent call last): File "D:\GitHub\RimSort\app__main__.py", line 96, in main_thread app_controller = AppController() ^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\controllers\app_controller.py", line 49, in **init** self.steamcmd_wrapper = SteamcmdInterface.instance( ^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 99, in instance cls._instance = cls(*args, **kwargs) ^^^^^^^^^^^^^^^^^^^^ File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 44, in **init** self.initialize_prefix(steamcmd_prefix, validate) File "D:\GitHub\RimSort\app\utils\steam\steamcmd\wrapper.py", line 77, in initialize_prefix os.makedirs(self.steamcmd_install_path) File "", line 215, in makedirs File "", line 215, in makedirs File "", line 215, in makedirs File "", line 225, in makedirs FileNotFoundError: [WinError 3] The system cannot find the path specified: 'E:'
> 
> It does not, in its current state this PR only makes it so that the dialog appears on JSON decode error. For cases where values for certain keys in the settings file are "invalid", I don't think it's possible to assume that the settings file is corrupt and present the same dialog.
> 
> In your example, what if the user does have an external E drive and that was the intended steamCMD install path, but then forgot to plug in the drive and ran RimSort? There's nothing malformed about the settings file, just an unexpected error with how the key/values are being used. In these cases I think a FatalError dialog is appropriate.
> 
> Thoughts, @hk21702 ?

I'll look at the code itself later, but yeah, that should be the behavior we aim for. Path validation and fixing should be something we handle separate from this.

hk21702:
I'm going to update the branch for the sake of letting the linters catch anything new

app/utils/generic.py
```diff
@@ -216,7 +216,7 @@ def platform_specific_open(path: str | Path) -> None:
     path = str(path)
     if sys.platform == "darwin":
         logger.info(f"Opening {path} with subprocess open on MacOS")
-        if p.is_file() or (p.is_dir() and p.suffix == ".app"):
```
hk21702, line 219: I think this is likely better behavior, to actually open files instead of where they are located in finder. However, I just want to confirm that this doesn't break behavior in previous uses.

Also, the docstring needs to be updated to reflect this change. Granted, I don't think the doc string seems to be super accurate even without this change.

hk21702 (changes requested):
I'm not a super big fan of how the backend logic is implemented for this, primarily because of MVC violations.

I think that a dialogue controller isn't necessary here, but a lot of this logic should be moved into the settings controller.

Immediately, because to control what happens on parse failure may require user input from an interface (view), these controls should be entirely handled by a controller that communicates with a view, and manipulates the model as necessary. 

In addition to this, current stucture breaks the single responsibility of load(). .load()'s fundamental job is to populate and initialize the settings model object. To do this, its acting like a service, doing a bunch of IO operations, reading and in some cases, writing. That is its single responsibility. However, by handling the decode error within .load(), we grow outside of the scope of just trying to load settings from file. Now its also trying to fix it if it fails. Now there is a wrapper around the IO service in a way that the service can never be called alone. These two things should really be separate. 

The settings controller when its calling .load() on the settings model, should be the one monitoring for exceptions that .load() raises. Then, if it gets an exception, it triggers for the dialogue (view) to open. The view and settings controller then communicate with each other based on user input that the view receives. The view catches a user input, tells the controller what happened, and the controller takes that information and does whatever it needs to do with the model before telling the view what to do now.

The controller triggers the creation of the model, and controls the logic for what to do on failure, based on user input provided by the view. 

Additionally, we may need an explicit close application button. In the case where the error dialogue isn't the only window open, currently, pressing the x button would likely just quickly reopen the dialogue. Since the user can't interact with the parent until the child is closed, it'd be impossible for them to close out of RimSort unless there is a successful resolution to settings not being able to be loaded. There is also the potential case where the dialogue does close on clicking x. However, the rest of RimSort doesn't close, creating a dangerous situation where the application is running on potentially corrupted data.

Technically, currently the above situation doesn't happen because .load() is only ever called on the init of the settings controller which only happens at application start. However, there is no restrictions on where .load() can be called in the future. Additionally, having the explicit close application button may be more intuitive from a UX perspective.

app/models/settings.py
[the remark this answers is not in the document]
oceancabbage: Your feedback is greatly appreciated. I've revised the PR so (hopefully) it does not violate MVC so horrendously. The loop's been removed and the exception is handled by the settings controller (as well as any additional calls).

app/utils/generic.py
```diff
@@ -216,7 +216,7 @@ def platform_specific_open(path: str | Path) -> None:
...
+        if p.is_dir() and p.suffix == ".app":
```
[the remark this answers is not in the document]
oceancabbage: It does change the behavior from what is was prior, but I believe what it was prior is not ideal. On Mac (and in this context), `-R` makes it so that the directory is opened (or, if it is a file, the parent directory). This "special" case is only really necessary for `.app`s, where going without `-R` would try to run the application.

Normal directory and file paths are handled just fine with `open` without `-R`.

app/views/dialogue.py
```diff
@@ -23,6 +24,7 @@
 )
 
 import app.utils.generic as generic
+from app.models.settings import Settings
```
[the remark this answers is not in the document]
oceancabbage: Yeah.. my bad. I replaced the "need" for this with a signal.

app/controllers/settings_controller.py
```diff
@@ -205,6 +208,19 @@ def __init__(self, model: Settings, view: SettingsDialog) -> None:
...
+        EventBus().reset_settings_file.connect(self._do_reset_settings_file)
+
+        self._load_settings()
+
+    def _load_settings(self) -> None:
+        logger.info("Attempting to load settings from settings file")
+        try:
+            self.settings.load()
+        except JSONDecodeError:
+            logger.info("Unable to parse settings file")
```
hk21702, line 221: I'd change this to an error lvl log.

app/views/dialogue.py
```diff
@@ -466,3 +459,118 @@ def _setup_messagebox(title: str | None) -> QMessageBox:
...
+        # Dynamic sizing
+        self.setSizePolicy(
+            QSizePolicy(QSizePolicy.Policy.Expanding, QSizePolicy.Policy.Expanding)
+        )
+
+        # Add data
+        self.text = "Your RimSort settings file is corrupt.\nPlease choose one of the following options to proceed."
+
+        # Buttons
+        self.open_settings_file_btn = QPushButton("Open Settings File")
```
hk21702, line 501: Just a nickpick and up to your discretion, but since the dialogue is a little crowded, we may want to differentiate better between "Open Settings File" and "Open Settings Folder", especially because the stings are the same up to "Open Settings F". Maybe just "Open Settings"?

app/views/dialogue.py
```diff
@@ -466,3 +459,118 @@ def _setup_messagebox(title: str | None) -> QMessageBox:
...
+        main_layout.addLayout(r_layout)
+
+        layout.addLayout(main_layout)
+
+        self.setLayout(layout)
+        self.setFixedWidth(self.sizeHint().width())
+
+        def _open_settings_file() -> None:
+            generic.platform_specific_open(
+                AppInfo().app_storage_folder / "settings.json"
```
hk21702, line 544: Wonder if its worth moving this into AppInfo() as a constant. May make things easier, and also if we want to in the future, make it easier to change the path.

app/views/dialogue.py
```diff
@@ -466,3 +459,118 @@ def _setup_messagebox(title: str | None) -> QMessageBox:
...
+    to parse the settings file.
+
+    :param settings: settings model to allow resetting settings
+    """
+    logger.info("Showing settings failure box")
+    diag = SettingsFailureDialog()
+    diag.exec_()
+
+
+class SettingsFailureDialog(QDialog):
```
hk21702, line 477: Looks generally good.

I do wonder if we should look at making a base class for these dialogs and use inheritance to set custom dialogs up. Don't need to do it now in this PR.

hk21702 (changes requested):
Generally good. I made a few comments for nitpicks here and there.

Additionally, I wonder if there should be a secondary message dialog if the user choses to reset the settings. This dialog would be a child of the main settings one and simply pop up if the reset also failed. It'd have a single button to close and go back to the parent. This way, there is just a bit more clarity if the reset also failed for whatever reason.

app/views/dialogue.py
oceancabbage: Agree, was thinking that too when making these revisions. May take it up in a next PR

app/views/dialogue.py
oceancabbage: Done!

app/views/dialogue.py
oceancabbage: Revised, thanks for the suggestion

app/controllers/settings_controller.py
oceancabbage: Revised, thanks

hk21702 (approved):

PR #537, opened 2024-07-28
[538] Refactor path autodetect, add win reg based steam path finding

hk21702:
Refactors autodetection to be easier to maintain.
Adds registry-based detection of steam install path for Windows.

Needs some testing to ensure that nothing broke for Linux and Mac systems.


oceancabbage (approved):
Tested and works on Mac

PR #554, opened 2024-08-08
Db builder warning and partial dialogue refactor

hk21702:
Implements conditional warning dialogs that pop up before critical db builder actions. Warns the user that the action is most likely not something they should be doing.

Utilizes a new binary conditional dialog which inherits from new base classes to be used with future dialogs. There are two of these base dialogs, a more generic one (_BaseDialogue) and one specifically for dialogs based on QMessageBox (_BaseMessageBox). 

Marked show_dialogue_confirmation as deprecated in favor of creating a BinaryChoiceDialog object and then using one of its exec methods instead.

Implements some pytests for TestBinaryChoiceDialog which require the new development lib, pytestqt. Adds a new step in the pytest action to install required linux libraries for QT.

image

image

image


app/controllers/settings_controller.py
```diff
@@ -1179,6 +1180,20 @@ def _on_db_builder_download_all_mods_via_steamcmd_button_clicked(self) -> None:
         """
         Build the Steam Workshop database of all mods using steamcmd.
         """
+        confirm_diag = BinaryChoiceDialog(
+            "Confirm Build Database (SteamCMD)",
+            "Are you sure you want to download all mods via SteamCMD and build the Steam Workshop database?",
+            (
+                "For most users this is not necessary as the GitHub SteamDB is adequate. Building the database may take a long time. "
+                "This process download all mods (not just your own) from the Steam Workshop. "
```
oceancabbage, line 1188: Nit: downloads* or will download

app/views/dialogue.py
```diff
@@ -279,31 +280,216 @@ def show_fatal_error(
...
+        modal: bool = True,
+        parent: QWidget | None = None,
+    ) -> None:
+        """Initializes the binary choice dialog.
+        Used to display a binary choice message box.
+        Always has two buttons, one positive and one negative.
+        These buttons cannot be the same type.
+
+        :param title: The title of the message box
+        :type title: str, optional
```
oceancabbage, line 411: Curious if the added verboseness of having types in the docstrings are needed, seeing as we already have type annotations. For functions like these it seems like a lot

oceancabbage (approved):

app/views/dialogue.py
hk21702: The types in the doc string are a part of the sphinx standard which is what we currently are following (at least according to the wiki). It's probably required for the auto-generated html documentation that sphinx would generate, though we aren't using that right now.

app/controllers/settings_controller.py
hk21702: Good catch.

PR #560, opened 2024-08-12
Revert "Minor Setting Changes"

hk21702:
Reverts RimSort/RimSort#499 based on previous discussions on UX design.

DB warning to be replaced by #554


oceancabbage (approved):
Works on MacOS. Revert looks good

PR #571, opened 2024-08-20
[141] Mod Description Refactor and Unity Rich Text Rendering

hk21702:
- Refactored the mod description widget for simplicity, such that it is just a QTextBrowser.
- Fixed MVC violation by making the widget a view instead of a model
- Implemented dynamic conversion of mod descriptions from Unity Rich Text to HTML
- Added setting toggle to disable conversion of URT to HTML (Default enabled)

Note that images are **not** rendered. PySide doesn't seem to have the ability to grab the image directly from the internet. I tried implementing some memory caching stuff to get the images for local use, but it only worked for some image hosts and there would still be major formatting issues that can't be resolved easily due to the lack of info in the URT format. As such, anything wrapped in the img tag is just removed. Note that in certain circumstances, this may cause excess whitespace/line breaks, particularly if the image was nested within an object like a table.

There may also be some edge cases where line breaks don't render as expected, though this should be mostly taken care of.
Explicit line breaks ("\n" strings) are converted to `<br>`. Otherwise, line break behavior is determined by the inferred schema.

image
image
image
image

This PR is dependent on #560


twstagg (changes requested):
Could this be rebased with `main`? I'd love to get this merged. Great work!

twstagg (approved):
I was actually able to rebase this cleanly via the fancy button on the PR. I pulled the result down and everything seems to be OK. Approving, feel free to merge at your discretion!

PR #573, opened 2024-08-20
Add graceful error handling for sorting algorithm not implemented

hk21702:
Adds graceful error handling for when sorting algorithm not implemented error. Warns the user with a warning dialogue instead of a fatal crash when RimSort doesn't recognize the sorting method that has been selected. The dialogue describes some potential avenues of mitigation. Mitigation is not done automatically to provide greater user control.

This is particularly useful for improper migration between versions, or from other mod managers like RimPy that happen to use the same settings file and location.

image


oceancabbage (approved):
Looks good to me

PR #576, opened 2024-08-21
[451] Add steamCMD cache clear options

### LINKED ISSUE #451: FIXME: SteamCMD (Probably) tries to restore previously deleted mods.
This one is hard to describe, reproduce, and everything in between.

The situation:
Eventually, RimSort or SteamCMD, tries to restore deleted mods (but only their complete folder structure with no content inside, not even in the about folder).
I'm uncertain if it tries to restore mods deleted by the user in file explorer in the past, or deleted via right clicking within RimSort, since I do both actions.

What I do know is this:
Behavior does not happen on a fresh installation, and it takes a week or so, or so many amount of mod download/updating/deletions for it to even trigger but it eventually does happen. I did notice last time I had this issue that the DepotCache folder in SteamCMD had a lot of files by the time it started troubling my mod folder, and that cleaning it up did remove the issue albeit maybe cleaning it up does mess with how RimSort figures what mods need updating since workshop mods suffer a lot from no changelogs or version data in them.

This shouldn't cause big issues but it does cause an issue where you don't know if the invalid folder is due to this event, or an actual mod you use (or made) that was corrupted somehow (this happened to me recently when a locally modified mod was just "validated" I guess and everything within removed).

Also, this happens even if you have never modified any mod manually (as this has been a problem since before I started "fixing" broken mods locally).

Finally, the replication does happen after you run a SteamCMD action like updating mods. Last time it happened to me 4 mods where updated then 60 "invalid mods" popped and upon looking at one of them via their steam ID I noticed they were all deleted mods I no longer wanted.

Edit:
Just another clarification for something I know will pop as common sense to ask: Are there any .dds files inside as remnants of previous optimization?
No, the mod that gets repopulated is 0 bytes in size. Every folder inside is empty, but it has the complete folder tree structure of the working mod. I have checked many to be sure there is no phantom residues and there is none. I also have by default the show hidden files enabled in Windows, the file extension to make sure they are indeed folders with no file extension and this is on RimSort 1.0.8 (Intel i7 8750h processor if that matters).

hk21702:
Hopefully fixes #451

Adds options to manually clear steamCMD depot cache, or auto-clear it before any steamCMD download operation.
By default, the option to auto-clear it before steamCMD download operations is enabled. This can be disabled in the settings.

This PR also implements a helper function to more robustly handle rmtree calls. Eventually, everything should be migrated to use this as it has special handling for OSErrors which should not cause fatal crashes for the app in most cases, but instead warn the user.

This PR is dependent on #554 as it uses some of the dialogue classes implemented there.

As a side note, I wonder if the steamCMD wrapper should be re-implemented as a controller. Additionally, it might make sense to rework the settings controller so it is more easily extendable.


hk21702:
Setting this to draft since clearing the depot cache may not be the full solution. It sitll needs to be explored.

twstagg (approved):
I like this. Coupled with #623 i think these will be some good QoL features

hk21702:
@twstagg Please re-review since there were lots of merge conflicts with main I had to resolve. Just want to make sure nothing broke.

twstagg (approved):
Seems to still work as intended for me. Go nuts!

PR #577, opened 2024-08-22
Fix steamDB creation crash when expiry or life is 0

hk21702:
Fix steamDB creation crash when expiry or life is 0.

Modifies the init of the DynamicQuery class such that life must be an int (default 0). Thus, self.expiry is always set, preventing a potential crash.

Also adds docstrings to __expires


oceancabbage (approved):
Nice fix!

PR #578, opened 2024-08-23
[221] (minor) Pygit2 migration and git improvements

### LINKED ISSUE #221: Implement pygit2 module instead of GitPython module as a replacement for Git-related actions
RimSort currently utilizes GitPython module to interact with Git repositories. This is not a popular choice, as there are some individuals who do not wish to install Git. We can attempt to replace GitPython with pygit2 and hopefully replace current functionality as the underlying library used by pygit2 (libgit2) does not require Git executable to be installed/present.

- [] Replace ALL functionality that currently relies on GitPython, and replacing it with equivalent functionality with pygit2
- [] Testing of affected existing features is needed as well
- [] Update the Wiki to reflect this change (by removing references to GitPython and it's dependency in any pages)

hk21702:
Closes #221 

Migrates all git operations to utilize Pygit2 instead. This should make it such that users don't need to have git installed locally to use git operations.
Implements a new controller for main_content_panel.
Improves flow of some git operations. For example, when checking for git updates, we don't immediately pull. Instead, we ask for confirmation while displaying a list of updates that from the remote.

- [ ] Migrate to Pygit2
- [ ] Implement threading for git operations

Dependent on #576


hk21702:
Added minor keyword to PR name to trigger a minor version increment when merged

Issue #562, opened 2024-08-13
FIXME: In some instances "clear list" in the mod downloader does only clear the list in the UI but when clicking "download [...] CMD)" downloads everything from the former list anyways.


Hans-be:
-

PR #581, opened 2024-08-25
[562] Fixed clearing list in mod downloader

Patryk-Malinowski:
Addressing git issue #562 

From my understanding '.clear()' clears the list and all references to it while '= []' only initialized that specific list to an empty list, and did not touch any references to itself.

Issue seemed to be that here we create a reference to this list in the signal:
image

And it retained the old data even if the list was 'cleared'

While I did not look into it, I assume we want the same behavior for the **downloader_list_dupe_tracking** dict, so I changed it to use .clear() as well.

### DIFF
--- a/app/utils/steam/browser.py
+++ b/app/utils/steam/browser.py
@@ -290,8 +290,8 @@ class SteamBrowser(QWidget):
 
     def _clear_downloader_list(self) -> None:
         self.downloader_list.clear()
-        self.downloader_list_mods_tracking = []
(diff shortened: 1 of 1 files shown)


hk21702:
Good catch. I think your understanding of how clear works is correct.

I'll double check and test this later when I can.

hk21702 (approved):

hk21702:
Looks good. Feel free to merge.

Patryk-Malinowski:
Don't think I have write access so will have to let someone else merge.

PR #590, opened 2024-08-28
[586] Replace issue templates with forms

hk21702:
Replaced the issue templates with forms to make it easier for users to submit issues.

This should also make it such that it is much less likely that required information is omitted, force the user to acknowledge conduct requirements, look for potential issues, and that the wiki may have the solution to their issue. As a bonus, it should make it so that rudimentary bots won't be able to make spam issues. 

image

image

image


hk21702:
As a side note, this should be squashed when merged

hk21702:
The original screenshots are a little out of date. There have been a few minor formatting changes.

image
image

oceancabbage (approved):
This is awesome :D

Issue #502, opened 2024-07-19
FIXME: Mismatch between warning and errors


hk21702:
**Describe the bug:**
When a mod has an error such as a missing dependency detected, the icon next to it is the warning icon. The error count at the bottom is correctly bumped and the error is correctly added into the tooltip. The text of the mod is also not red/highlighted.

**Screenshots:**
![image](https://github.com/user-attachments/assets/87db77c8-2184-4207-8496-67ac2f523cfe)
![image](https://github.com/user-attachments/assets/effca558-2a24-4fbc-a875-e5ca2c016773)
![image](https://github.com/user-attachments/assets/7cb362aa-e0f7-4cb7-bfa4-32eb0e74339d)

**Please complete the following information:**

- Operating System: Windows
- RimSort version: 1.0.9-edge92

**Additional context:**

Patryk-Malinowski:
Quick question, should a mod with a warning be red?

Or should warnings only have the warning icon and errors are red with the error icon?

hk21702:
Current behavior is that warnings are also red. For example, version mismatch counts as a warning and results in red text, a warning icon, and a warning added to the bottom. I'd say just keep current behavior on that front.

Patryk-Malinowski:
Ok will do, cheers!

Patryk-Malinowski:
When a mod has a warning because of load order, should it be colored red? The current behavior is that its not colored.
![image](https://github.com/user-attachments/assets/94a34471-530c-4068-adde-c32b3ea46f17)

hk21702:
For consistency probably? Given that it counts as a warning, I'd presume it should be the same as other warnings.

Issue #459, opened 2024-06-27
Suggestion: Make "toggle warning" hide warnings for a mod completely


Limfjorden:
**Is your feature request related to a problem? Please describe.**

Currently, all the "toggle warning" button does i hide the yellow warning icon from the mod name, while keeping the red mod name and also listing the warning in the section below the mod list.

**Describe the solution you'd like**

- [ ] It should hide the icon, change the mod name away from red (maybe to a purple color to indicate that a manual override has been activated) and remove the warning from the section below the mod list. 
- [ ] There should also be a shortcut in the "Edit" menu to clear all manual warning toggles

**Additional context**

From Raizanad's QoL issue https://github.com/RimSort/RimSort/issues/406:
> Many mods for 1.4 work on 1.5, specially textures as an example. Would be great if it changed the about.xml too for the game not to trigger errors but this may be too much since you will need a db to remove the change (or changing about.xml to about.xml.old).
I'm mostly talking about version mismatch, but there are also mods that set something as incompatible when they are soft incompatibilities that could be desired to not be highlighted. As an example some mods that add races with animal ears have some trouble with "Select hats display" or something, but that mod has 2-3 things going for it and if you set the setting properly they are compatible. The warning by default is great since otherwise the pawns are invisible but once settings are configured the warning is annoying.

I'd like to try my hand at implementing this, though I don't know any python. Is this a good first issue?

PR #592, opened 2024-08-30
[502] [459] Improvements to mod warning/error display and toggle

Patryk-Malinowski:
Addressing git issue #502 

Slightly related to git issue #459 ?

Quick [demo](https://github.com/user-attachments/assets/b5dd9da9-9def-4d3e-b0c3-24f69b7946f6) of how it should work.

Forgot to show in the demo but it also colors mods red if missing dependency etc.
image

In it's current state the 'toggle warning' button will show/hide both the warnings and errors. Is this behavior something we want to keep? Maybe separate toggle for warning and error?

Also, if you 'toggle warning' on a mod with an error/warning it **stops** being red. Is this ok or do we want it to stay red? Or will I implement changing it to purple (or other color) to signify a manual override (suggested in git issue 459)?


hk21702:
I personally think it makes more sense to also make the text not red. Though, it is a bit harder to track when things are hidden though given the current state. Ideally, #459 is fully implemented with the clear all ignores button. Ofc, rn given that I still haven't gotten around to the metadata rework, toggle state isn't persistent, you can just restart rimsort to clear the ignores.

Additionally, the toggle button in the context menu should be changed to be an actual checkable qaction with changing icon instead of the current just default qaction.

hk21702 (changes requested):
Behavior for lazily loaded items also needs to be implemented (Around line 290). In the current implementation, the intended behavior only works if the item was already being rendered, i.e., items that you scroll into still have the old broken behavior such that `repolish` isn't ran.

Additional small nitpick. I think this issue existed before this PR as well, but can you strip the lazily loaded tooltips? That or just lstrip the tooltip before saving it in the item data.

Patryk-Malinowski:
> Behavior for lazily loaded items also needs to be implemented (Around line 290). In the current implementation, the intended behavior only works if the item was already being rendered, i.e., items that you scroll into still have the old broken behavior such that `repolish` isn't ran.
> 
> Additional small nitpick. I think this issue existed before this PR as well, but can you strip the lazily loaded tooltips? That or just lstrip the tooltip before saving it in the item data.

Good catch with the lazy loading behavior. That's impl. now as well as stripping the tooltips.

Patryk-Malinowski:
> I personally think it makes more sense to also make the text not red. Though, it is a bit harder to track when things are hidden though given the current state. Ideally, #459 is fully implemented with the clear all ignores button. Ofc, rn given that I still haven't gotten around to the metadata rework, toggle state isn't persistent, you can just restart rimsort to clear the ignores.
> 
> Additionally, the toggle button in the context menu should be changed to be an actual checkable qaction with changing icon instead of the current just default qaction.

Looking into adding that button to reset all ignores...

hk21702:
When you're ready, feel free to request a new review

Patryk-Malinowski:
Tried to make the toggle button a checkable QAction but after spending some time I don't think I will be able to implement it as easily as I thought. Can I make a separate git issue for this to be implemented?

Resetting warning/error toggles is now implemented.

hk21702:
Go ahead and make a new issue. Will review when I can.

hk21702 (changes requested):
See comments.

I'm a bit unsure of where _on_menu_bar_reset_warnings_triggered should go. Ideally, it should go into a dedicated controller for the main content panel or a dedicated controller for the mods panel. However, that doesn't exist right now. Thus I think it may be better to just stick it in the MainWindowController or a new dedicated controller for the mods panel or main content view. 

Right now the main content view does violate mvc a bit and a lot of it should be going into a controller. But imo it would make a little more sense for the modification of the mods panel to go under there than under the menu bar controller. The best solution would be to just have a new controller for this kind of stuff. I'll leave it up to your discretion though.

In addition to this, _on_menu_bar_reset_warnings_triggered is quite slow because of how its iterating through all of the mods, and also because how how reset_warning_signal works. For my mid sized mod list of about 200 mods, it takes a good 2 or 3 seconds for things to be updated. See the relevant comment for more context.

Patryk-Malinowski:
Oh that's on me, I forgot about MVC. Not experienced with MVC but I would rather do it right with a mods_panel controller. I will spend some time refactoring the code and learning a bit on MVC, and then come back here.

Patryk-Malinowski:
I wasn't too sure about having a controller inside another controller, but all menu_bar_controller does is emit a signal through mods_panel_controller. Is this acceptable?

I saw that menu_bar_controller already uses settings_controller hence I went with this approach.

Also, not sure what I was thinking with the logic of _on_menu_bar_reset_warnings_triggered... this should be improved now. Tried it on ~700 mods and it seems to be better.

hk21702:
Will take a peak when I can. It's perfectly fine to have nested controllers though! I think the rule is that they generally they should only be aware of their own scope though

Patryk-Malinowski:
That's good to know, thanks! I think I could also come up with some intermediate service layer between the two controllers if emitting that signal isn't the best approach.

app/views/mods_panel.py
```diff
@@ -1538,6 +1574,32 @@ def append_new_item(self, uuid: str) -> None:
         item.setData(Qt.ItemDataRole.UserRole, data)
         self.addItem(item)
 
+    def get_all_loaded_mod_list_items(self) -> list[ModListItemInner]:
+        # This gets all modlist items as ModListItemInner.
+        # Mods that have not been loaded
+        # or lazy loaded will not be returned.
```
hk21702, line 1580: Please format this as docstring. (Sphinx)

app/views/mods_panel.py
```diff
@@ -1538,6 +1574,32 @@ def append_new_item(self, uuid: str) -> None:
...
+            item = self.item(index)
+            widget = self.itemWidget(item)
+            if isinstance(widget, ModListItemInner):
+                mod_list_items.append(widget)
+        return mod_list_items
+
+    def get_all_loaded_and_toggled_mod_list_items(self) -> list[ModListItemInner]:
+        # This gets all modlist items that have their warnings toggled as ModListItemInner
+        # Mods that have not been loaded
+        # or lazy loaded will not be returned.
```
hk21702, line 1592: Please format this as docstring. (Sphinx)

app/views/mods_panel.py
```diff
@@ -1538,6 +1574,32 @@ def append_new_item(self, uuid: str) -> None:
...
+        mod_list_items = []
+        for index in range(self.count()):
+            item = self.item(index)
+            widget = self.itemWidget(item)
+            if isinstance(widget, ModListItemInner):
+                packageid = self.metadata_manager.internal_local_metadata[widget.uuid].get('packageid')
+                if packageid in self.ignore_warning_list:
+                    mod_list_items.append(widget)
+        return mod_list_items
```
hk21702, line 1601: Because this only gets non-lazy items, if the user refreshes and then tries to reset toggles, lazy items are not reset, even after loading them.

hk21702: To solve this, you may want to handle the toggle state in the QListWidgetItem's data. In relation to this, if you do this, because of the increasing complexity of the data, you may also want to change the data to be a class instead of a dict/map.

hk21702 (changes requested):
See review comments.

Overall, much better especially in performance and structure.
However, there is a bug where the reset function doesn't work if the item hasn't been widget hasn't been loaded and is lazy.

To solve this, you may want to handle the toggle state at the  QListWidgetItem data level. If you do so, I also recommend changing the data from a map to a class/dataclass because of the increased complexity and different datatypes being stored potentially.

app/views/mods_panel.py
```diff
@@ -2054,12 +2076,19 @@ def recreate_mod_list(self, list_type: str, uuids: List[str]) -> None:
...
         logger.debug(f"Toggled warning icon for: {packageid}")
+        current_mod_index = self.uuids.index(uuid)
+        item = self.item(current_mod_index)
+        item_data = item.data(Qt.ItemDataRole.UserRole)
         if packageid not in self.ignore_warning_list:
             self.ignore_warning_list.append(packageid)
+            item_data["warning_toggled"] = True
         else:
             self.ignore_warning_list.remove(packageid)
+            item_data = item.data(Qt.ItemDataRole.UserRole)
```
Patryk-Malinowski, line 2089: Duplicate line, need to remove

Patryk-Malinowski:
I had trouble trying to use a custom class to store data in combination with a wrapper for QWidgetListItem. Can I make a separate git issue? Then It could be revisited by me or someone else in the future.

The original behavior is that if you toggle a warning for a mod in the inactive list and move it to the active list, any errors will appear for the mod while its in the active list. It also works vice-versa. 

In it's current iteration the reset warnings button will only reset a mods warning toggle for that modlist. So when moved back into the inactive list it will still be ignored in that list.

Demo [Video](https://github.com/user-attachments/assets/aa4957d1-0b19-4f44-a0f2-e1bcdccb0aaf)

If we want to keep this behavior I could optionally change the reset button to reset both lists (I assume that would be the desired behavior in that case)

But I also have an implementation where the ignore and reset is applied to both lists.

Demo [Video](https://github.com/user-attachments/assets/9b323957-ea22-415e-9d32-c072bd0081d0)

Let me know which route you want to take! Also I joined the discord so I will dm someone to get added to the dev channel :)

Patryk-Malinowski:
Oh and for the mac build fail, did it just crash/timeout when uploading?

hk21702:
> I had trouble trying to use a custom class to store data in combination with a wrapper for QWidgetListItem. Can I make a separate git issue? Then It could be revisited by me or someone else in the future.

Go ahead and make that a separate git issue

app/controllers/mods_panel_controller.py
```diff
@@ -0,0 +1,33 @@
...
+            mod_data = mod.data(Qt.ItemDataRole.UserRole)
+            if mod_data["warning_toggled"]:
+                mod_data["warning_toggled"] = False
+                mod.setData(Qt.ItemDataRole.UserRole, mod_data)
+                widget = mod.listWidget()
+                if isinstance(widget, ModListWidget):
+                    package_id = widget.metadata_manager.internal_local_metadata[mod_data["uuid"]]["packageid"]
+                    widget.reset_warning_signal.emit(package_id)
+        self.mods_panel.active_mods_list.recalculate_warnings_signal.emit()
+        self.mods_panel.inactive_mods_list.recalculate_warnings_signal.emit()
```
hk21702, line 33: Should be auto-formatted with Ruff

hk21702 (changes requested):
Looks good overall. the controller needs to be ran through an auto-format with Ruff. 

Also, consider doing the lstrip on the error/warning tooltips at assignment instead of use. I can't think of a situation where we'd need the white space at the start anyways. In fact, since we add new items by putting the whitespace at the start at the time of concat, maybe we can just do a strip at assignment? There may be situations where performance costs increase by doing it this way, but it should be pretty minimal. It would make the code a lot easier to work with and cleaner though so imo its worth the tradeoff.

Patryk-Malinowski:
I changed the drop signal behavior so that if a drop occurs from one list to another it doesn't call ```recalculate_list_errors_warnings``` a third time.

hk21702 (approved):
Think there's a few redundant lstrips here and there but otherwise seems good.

PR #605, opened 2024-09-07
Symlink creation improvements and refactor

hk21702:
Moved bulk of symlink creation logic into its own module.

Improved error handling of symlink/junction creation.
Ensures that it is impossible for the user to accidentally delete files that they do not intend to be deleted when creating symlinks/junctions.
Fixed crash during symlink creation when the destination path parent dir did not exist.


twstagg (approved):
Slick!

PR #610, opened 2024-09-09
Remove pytz and types-Deprecated, replace outdated type libs with new typeshed libs

hk21702:
- Removed pytz from requirements as it was not being used anywhere and the library recommends that projects using 3.9 or later use the standard lib instead. https://pypi.org/project/pytz/
- Removed types-Deprecated as the Deprecated lib was not being used directly
- Replaced networkx-stubs and lxml-stubs with their typeshed equivalents
- Cleaned up and sorted requirements files
- Fixed any emerging type errors


twstagg (approved):
Thank you for your work!

PR #612, opened 2024-09-12
[603] Implement instance specific ability to ignore future steamcmd setup dialogs on startup

### LINKED ISSUE #603: [Suggestion]: Better SteamCMD Declined Install
### Origin of idea

Currently, if the SteamCMD is not detected on the start of RimSort or the switching to a new instance, either because it was manually removed or never installed in the first place, RimSort will prompt the user if they would like to install RimSort. Considering that SteamCMD is completly optional, this extra dialog can cause notification fatigue and be confusing for some users.

If the user declines to install SteamCMD on setup for a particular instance, RimSort should no longer ask at startup if they would like to install SteamCMD if it does not find steamCMD.

Note that it should still ask to install SteamCMD on the creation of a new instance.

For cloning if steamCMD is not found, we can also not have the dialog. This is as in current behavior of instance cloning, if SteamCMD is installed in the original isntance, that install is cloned into the new instance.

### Alternatives considered

_No response_

### Additional context

SteamCMD is an optional component for RimSort, providing an alternative way to download and update Steam workshop mods. It is only needed for this function.

Currently, if the user tries to download mods via SteamCMD, but RimSort does not find SteamCMD at the configured location, it will ask to install it. This behavior is fine. 

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

hk21702:
Closes #603 
The only way to reset is to reset this is to set settings to default.

SteamCMD can still be installed afterward. This only affects the check on startup of RimSort or startup of a RimSort instance.

Also added a new method in the settings controller to more easily access the active instance object.


hk21702:
Should not be merged as pytz is not in use and will be fully removed in #610

hk21702:
Somehow closed wrong pr >.>

twstagg (approved):
Much better. Thanks for implementing this!

PR #613, opened 2024-09-15
Sort the inactive mod list alphabetically

lisanhu:
Hi guys!

It seems the sorting functionality was gone since some big updates ago. I checked this part of code and tried to make it work again. 

It seems that there are some consideration for performance impacts of the recreate_mod_list function and it has been resolved. 

I did my own tests by adding a `print('sorting')` in the `sort_uuids` function and it looks like there's no abundant sorting anymore (I removed that print clause in this commit). However, I'm not 100% sure this is not creating any issues and I really wish someone would like to really look into this and test its impact on performance. 

For me, it's doing very well and not having any issues in performance right now.

Changes in this pull request: 
1. update the ModsPanelSortKey enum class to use numbers not functions as enum values
2. add a sort_uuids global function to sort the uuids based on the ModsPanelSortKey parameter
3. update the behaviour of `recreate_mod_list_and_sort` to use sorted uuids rather than the original uuids list

### DIFF
--- a/app/views/mods_panel.py
+++ b/app/views/mods_panel.py
@@ -87,9 +87,24 @@ class ModsPanelSortKey(Enum):
     """
     Enum class representing different sorting keys for mods.
     """
+    NOKEY = 0
+    MODNAME = 1
 
-    NOKEY = None
-    MODNAME = uuid_to_mod_name
+
+def sort_uuids(uuids: List[str], key: ModsPanelSortKey) -> List[str]:
+    """
+    Sort the list of UUIDs based on the provided key.
+    Args:
+        key (ModsPanelSortKey): The key to sort the list by.
+    Returns:
+        None
+    """
+    # Sort the list of UUIDs based on the provided key
+    if key == ModsPanelSortKey.MODNAME:
+        sort_function = uuid_to_mod_name
+    else:
+        sort_function = lambda uuid: uuid
+    return sorted(uuids, key=sort_function)
 
 
 class ModListItemInner(QWidget):
@@ -1917,9 +1932,10 @@ class ModListWidget(QListWidget):
             None
         """
         # TODO: Fix this or just get rid of it
-        sorted_uuids = uuids
+        # sorted_uuids = uuids
         # if key != ModsPanelSortKey.NOKEY:
         #    sorted_uuids = sorted(uuids, key=key)
+        sorted_uuids = sort_uuids(uuids, key = key)
         self.recreate_mod_list(list_type, sorted_uuids)
 
     def recreate_mod_list(self, list_type: str, uuids: List[str]) -> None:


hk21702:
Hi.

Will look at this more closely when I can, but you need to first fix the lint errors. Specifically you have a pep8 violation where you assigned a lambda function instead of a named function.

lisanhu:
Thank you for looking into this! I think this new commit has resolved the linting issue and the checks have passed!

I also updated a variable name in this commit.

app/views/mods_panel.py
```diff
@@ -1917,9 +1943,10 @@ def recreate_mod_list_and_sort(
             None
         """
         # TODO: Fix this or just get rid of it
-        sorted_uuids = uuids
+        # sorted_uuids = uuids
         # if key != ModsPanelSortKey.NOKEY:
         #    sorted_uuids = sorted(uuids, key=key)
```
hk21702, line 1948: Should be cleaned up and removed.

app/views/mods_panel.py
```diff
@@ -1917,9 +1943,10 @@ def recreate_mod_list_and_sort(
             None
```
hk21702, line 1943: Docstring needs to be fixed to include key

hk21702: Also needs improved description that is more accurate.

app/views/mods_panel.py
```diff
@@ -87,9 +98,24 @@ class ModsPanelSortKey(Enum):
...
+    Args:
+        key (ModsPanelSortKey): The key to sort the list by.
+    Returns:
+        None
+    """
+    # Sort the list of UUIDs based on the provided key
+    if key == ModsPanelSortKey.MODNAME:
+        key_function = uuid_to_mod_name
+    else:
+        key_function = uuid_no_key
```
hk21702, line 117: I think this should just be a lambda function, given how its just returning the uuid? Though it could be better to be more explicit for continuity reasons.

app/views/mods_panel.py
```diff
@@ -1917,9 +1943,10 @@ def recreate_mod_list_and_sort(
             None
         """
         # TODO: Fix this or just get rid of it
-        sorted_uuids = uuids
+        # sorted_uuids = uuids
         # if key != ModsPanelSortKey.NOKEY:
         #    sorted_uuids = sorted(uuids, key=key)
+        sorted_uuids = sort_uuids(uuids, key = key)
```
hk21702, line 1949: Needs to go through formatting via Ruff.

hk21702 (changes requested):
Some minor nitpicks here and there. See comments.

app/views/mods_panel.py
hk21702: Will leave this up to your discretion. Just a suggestion.

app/views/mods_panel.py
lisanhu: Thanks for looking into this! I'm working on addressing these issues.

app/views/mods_panel.py
lisanhu: It was a lambda function, and I changed it to named function because of the pep 8 violation. I don't know how to go through that linting check using a lambda function, do you know what is the correct way of doing it in this case? Thanks a lot!

lisanhu: Currently vs code / ai helper has told me to use a comment to skip that checking on that line, I'm not sure it's a good way to do this. What do you think?

app/views/mods_panel.py
lisanhu: Hi! I forgot to reply to this one. I used Ruff as formatter in vscode, I'm not sure whether this is the correct way of doing it. Could you please help check it with ruff for this commit?

app/views/mods_panel.py
hk21702: You should be able to instead do something like


```suggestion
        return sorted(uuids, key=lambda x: x)
```

Its not as clean in that you're repeating the uuids part of that func call. But, it gets around the pep error because you're not assigning the lambda function to a variable. 
As another alternative, use partial from functools to do something like

```suggestion
    pf = partial(sorted, uuids)
    if key == ModsPanelSortKey.MODNAME:
        return pf(key=uuid_to_mod_name)
    else:
        return pf(key=uuid_no_key)
```

app/views/mods_panel.py
hk21702: Looks correct in latest commit!

app/views/mods_panel.py
lisanhu: What do you think about the current docstring? Shall we resolve this?

hk21702 (approved):

PR #620, opened 2024-09-22
[619] Fix git url parsing, add improved git action dialogue contents, fix metadata deletion race condition crash

### LINKED ISSUE #619: [Bug]: Adding git mod replaces all local mod files if tail of the url is empty
### Release Type

Compiled Windows x86 64-bit

### Version

RimSort v1.0.9

### Operating System

Windows 10

### Relevant logs

https://0x0.st/XY8n.log

### Describe the bug

1. Select "Download" at ribbon.
2. From dropdown menu, choose "Browse Workshop"
3. Select any mod from Steam Workshop and download via steamcmd.
4. Refresh RimSort, the downloaded mod should show under Inactive list.
5. Select "Download" at ribbon.
6. Select "Add Git Mod".
7. Enter this link 
https://github.com/Maruf61/RimWorld-ResearchPal/
and press OK
8. Select "Clone new". Wait for the "Repo retrieved" prompt.
9. Close the prompt and refresh RimSort.
10. Now, all mods become "Invalid item" under Inactive list, including the successfully downloaded steam mods.

### Additional context

Video below shows the reported bug in action

https://github.com/user-attachments/assets/c7f9a69a-f481-47f4-94f8-3e574e181821

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

hk21702:
Fixes #619.

Makes it such that url parsing of http/s git urls should be much more robust. This includes the situation where the given url ends in a `/` with no other contents afterwards. Previously, that situation when parsing for the repo dir name would return an empty string as there was nothing after the last slash. These methods were factored out into generic with corresponding tests.

Additionally, the git clone action dialogue contents were improved. On bad url detection, the url is printed in the details section. On clone success, the display of the git url is not hyperlinked.

### DIFF
--- a/app/utils/generic.py
+++ b/app/utils/generic.py
@@ -291,3 +291,33 @@ def upload_data_to_0x0_st(path: str) -> tuple[bool, str]:
             f"Failed to upload data to http://0x0.st. Status code: {request.status_code}"
         )
         return False, f"Status code: {request.status_code}"
+
+
+def extract_git_dir_name(url: str) -> str:
+    """
+    Function to extract the directory name from a git url
+
+    :param url: a string url to a git repository
+    :return: a string that is the directory name of the git repository
+    """
+    return url.rstrip("/").rsplit("/", maxsplit=1)[-1].removesuffix(".git")
+
+
+def extract_git_user_or_org(url: str) -> str:
+    """
+    Function to extract the organization or user name from a git url
+
+    :param url: a string url to a git repository
+    :return: a string that is the organization name of the git repository
+    """
+    return url.rstrip("/").rsplit("/", maxsplit=2)[-2].removesuffix(".git")
+
+
+def check_valid_http_git_url(url: str) -> bool:
+    """
+    Function to check if a given url is a valid http/s git url
+
+    :param url: a string url to a git repository
+    :return: a boolean indicating whether the url is a valid git url
+    """
+    return url and url != "" and url.startswith("http://") or url.startswith("https://")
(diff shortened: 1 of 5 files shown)


Patryk-Malinowski (changes requested):
If you enter a non existing git url like: `https://github.com/Maruf61/RimWorld-ResearchPal5/`, it seems to display the appropriate warning followed by RimSort crashing.
image

image

You are then left with an invalid item on re-opening of RimSort:
image

Patryk-Malinowski:
Also no idea would this improve anything or add clutter to the message, but what do you think about adding something like: `latest commit -> commit` when updating a repo?

image

I guess if a user is updating a git mod they would probably connect the dots that Update Readme.md is the latest commit...

hk21702:
> If you enter a non existing git url like: `https://github.com/Maruf61/RimWorld-ResearchPal5/`, it seems to display the appropriate warning followed by RimSort crashing. image
> 
> image
> 
> You are then left with an invalid item on re-opening of RimSort: image

I think this is an existing bug? I think the crash is an existing issue related to the metadata where there's a race condition. I can try to fix that but I'm not going to touch the leaving of leftover files because I think there is some what I presume to be some weird intended behavior here for some case that makes it hard to do. I'm not entirely sure though since I didn't write this part. 

Thus I'm just going to see if I can fix the crash, and let the leftover files stuff exist until the pygit2 migration. #578

hk21702:
> Also no idea would this improve anything or add clutter to the message, but what do you think about adding something like: `latest commit -> commit` when updating a repo?
> 
> image
> 
> I guess if a user is updating a git mod they would probably connect the dots that Update Readme.md is the latest commit...

Something like this? 
image

Or did you have something else in mind?

Patryk-Malinowski (approved):
lgtm. Didn't realize the crash was caused by an existing bug but great job on handling it here.

PR #623, opened 2024-09-29
[380][451][635] Cleanup of SteamCMD mod manifest + acf metadata on deletion

### LINKED ISSUE #380: Suggestion: Don't delete the About folders/xml files when recursively deleting/batch redownloading mods

### LINKED ISSUE #451: FIXME: SteamCMD (Probably) tries to restore previously deleted mods.

### LINKED ISSUE #635: rimsort fails to update/download mods
### Release Type

Compiled (Windows x86-64)

### Version

RimSort-v1.0.10-Windows_x86_64

### Operating System

windows 11

### Relevant logs

_No response_

### Describe the bug

i'm getting I/O operation failed erros coming from the workshop update feature or whetever i try to download other contents from the workshop lately, i updated rimworld with the latest version that just came out today.

config file location is correct and local mods location is correct, i didn't touched anything other than deleting the old game version and installing the new one plus changing the locations

### Additional context

Captura de Tela (6)

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

twstagg:
Attempts to fix #380, fix #451, fix #635
- SteamCMD acf monitoring via Watchdog
- Pops the metadata for the SteamCMD mod's publishedfileid when deleting via RimSort.
- Removes the SteamCMD mod manifest file from the depotcache if found for the respective mod, after deleting.
- Fix several minor bugs, utilize `handle_remove_read_only` in rmtree usages
- Add support to `ModInfo` for `descriptionbyversion` tag
    - This is not documented on RimWorld wiki, but was tested and confirmed to be supported in-game mod manager.
- Convert watchdog.py to use Sphinx reST format docstrings
- Fix various MyPy errors
- Pep604 instead of Typing.Optional everywhere
- Replace `Typing.List` usages with `list`


hk21702 (changes requested):
Posting summary of discussions from Discord. 

Consider making steamcmd_purge_mod handle batches of mods instead of one at a time to reduce acf read and writes from n to 1 when doing deletions of multiple mods, maybe set self.workshop_acf_data to the post purge data at the end of steamcmd_purge_mod, on failure of reading acf in __refresh_acf_metadata, maybe set self.workshop_acf_data to the default value of an empty dict instead of leaving it as is, though I'm not entirely sure about this one on whats better tbh, and finally, consider doing is not None checks for your if statements involving .get() of parsed data for cases where the parsed data does exist, but the data just would happen to evaluate to False in Python like an empty string or 0

twstagg:
Additionally implemented:

- Watchdog acf file monitoring, refresh signal parses the files into `MetadataManager` on signal emission
  - `PollingObserver` is used for these due to probable bug on Windows with individual files being observed directly.

twstagg:
> Posting summary of discussions from Discord.
> 
> Consider making `steamcmd_purge_mod` handle batches of mods instead of one at a time to reduce acf read and writes from n to 1 when doing deletions of multiple mods, maybe set `self.workshop_acf_data` to the post purge data at the end of `steamcmd_purge_mod`, on failure of reading acf in `__refresh_acf_metadata`, maybe set `self.workshop_acf_data` to the default value of an empty dict instead of leaving it as is, though I'm not entirely sure about this one on whats better tbh, and finally, consider doing `is not None` checks for your if statements involving .get() of parsed data for cases where the parsed data does exist, but the data just would happen to evaluate to False in Python like an empty string or 0

Safer None-check has been implemented, and the overall process and flow is much more intuitive now.

app/utils/steam/steamcmd/wrapper.py
```diff
@@ -28,7 +28,7 @@ class SteamcmdInterface:
     Create SteamcmdInterface object to provide an interface for SteamCMD functionality
     """
 
-    _instance: Optional["SteamcmdInterface"] = None
+    _instance = None  # type: SteamcmdInterface
```
hk21702, line 32: Pylance doesn't like this.

To fix it, I'd do. Pylance and MyPy seem to be okay with it.

```suggestion
    _instance: "None | SteamcmdInterface" = None

```

app/utils/app_info.py
```diff
@@ -20,7 +19,7 @@ class AppInfo:
         >>> print(app_info.app_storage_folder)
     """
 
-    _instance: Optional["AppInfo"] = None
+    _instance = None  # type: "AppInfo" | None
```
hk21702, line 23: Try this:

```suggestion
    _instance: "None | AppInfo" = None
```

app/utils/gui_info.py
```diff
@@ -16,7 +15,7 @@ class GUIInfo:
     of the class exists throughout the application.
     """
 
-    _instance: Optional["GUIInfo"] = None
+    _instance = None  # type: GUIInfo | None
```
hk21702, line 19: Try this
```suggestion
    _instance: "None | GUIInfo" = None
```

app/utils/metadata.py
```diff
@@ -44,7 +38,7 @@
 
 
 class MetadataManager(QObject):
-    _instance: Optional["MetadataManager"] = None
+    _instance = None  # type: MetadataManager
     mod_created_signal = Signal(str)
```
hk21702, line 42: Pylance dislikes this. Try the following:

```suggestion
    _instance: "None | MetadataManager" = None
```

app/utils/system_info.py
```diff
@@ -18,9 +17,9 @@ class SystemInfo:
         >>> print(info.architecture)
     """
```
hk21702, line 19: Try the following.

```suggestion
    _instance: "None | SystemInfo" = None
    _operating_system: "None | SystemInfo.OperatingSystem" = None
    _architecture: "None | SystemInfo.Architecture" = None
```

app/utils/watchdog.py
```diff
@@ -180,72 +257,74 @@ def on_deleted(self, event: FileSystemEvent) -> None:
...
+                    "path": mod_path,
+                    "uuid": uuid,
+                }
+            )
+        else:
+            # logger.debug(
+            #     "UUID resolution failed for mod modified event: "
+            #     f"[event_scr_path_str: {event_scr_path_str}, uuid: {uuid}"
+            # )
+            return
```
hk21702, line 300: For cleanliness, I'd either just remove this, or uncomment out the logger

app/views/dialogue.py
```diff
@@ -33,11 +33,11 @@
...
-    title: Optional[str] = None,
-    text: Optional[str] = None,
-    information: Optional[str] = None,
-    details: Optional[str] = None,
-    button_text: Optional[str] = "Yes",
+    title: str | None = None,
+    text: str | None = None,
+    information: str | None = None,
+    details: str | None = None,
+    button_text: str | None = "Yes",
```
hk21702, line 40: Given that there is a default string for button_text, I don't think we need to make it | None, just make it str. Either that, make it default = None so that it defaults to the text from QMessageBox.StandardButton

Also, it doesn't seem like the param is used either way. However, realistically this method should be deprecated anyways since BinaryChoiceDialog should replace it and is much more powerful.

app/views/main_window.py
hk21702, line 1: For perfection, the logic here not related to the user viewing of data should be moved into the controller. Though that can be done at some other point I suppose.

app/views/mods_panel.py
```diff
@@ -7,7 +7,7 @@
 from pathlib import Path
 from shutil import copy2, copytree, rmtree
 from traceback import format_exc
-from typing import List, Optional
+from typing import List
```
hk21702, line 10: while you're at it, List from typing can be just replaced with list[]

hk21702 (changes requested):
There are a few improvements to how the typing is done that can be implemented.

Additionally, some of the logic in the main window view should be moved into the controller, especially if it doesn't have to do with the manipulation of what the user directly sees. This can prob be done at a later date but just a heads up.

hk21702:
There will prob be significant merge conflicts when #576 is merged into main

app/utils/system_info.py
twstagg: I will commit in batch

app/utils/watchdog.py
twstagg: @hk21702 Sounds good. It can easily be re-added.

app/views/dialogue.py
twstagg: @hk21702 I think that somebody sorta bastardized this along the way, because the function did work as intended when i made it. However, the new dialog is a better option, so I will implement that where it is needed and fully remove this function.

app/views/mods_panel.py
twstagg: Yeah i was just lazy. There was only a few left that i see. I took care of them.

app/views/main_window.py
twstagg: Definitely as a separate PR, but i am down to do the work. To be honest maybe a little guidance would get me on the right track.

hk21702 (approved):
Seems okay. May want to use the rmtree wrapper from utils.generic instead of directly from shutils but I will leave that up to you. Also may want to set depot cache clear setting to off by default as discussed.

PR #640, opened 2024-10-12
Minor metadata parsing bugfixes

hk21702:
Fixes some minor issues related to metadata for certain edge cases.

 - Add dict check for incompatiblewith entry before parsing for cases where it exists, but the list does not to prevent failed parsing.
 - Strip individual supported version strings to fix supported version parsing when the about.xml has extra white spaces and new lines such as in the following situation:
 
 
image


LionelColaso (approved):
Looks Good Thanks

twstagg (approved):
Good looks

Issue #608, opened 2024-09-09
[Suggestion]: Migrate Documentation to GitHub pages/Jekyll


hk21702:
### Origin of idea

Currently, GitHub does not support pull requests for a repository's wiki. This has been a long-standing issue with GitHub's wiki feature, and there doesn't seem to be any indications that it will change any time soon. The wiki counts as a separate repo from the main repo, but the wiki repo does not support PRs. Our current implementation is that the wiki repo is a submodule within the main repo. This gives us the advantage where local copies of the wiki can be up-to-date with the true wiki. However, it makes contributing, more difficult than it needs to be. Currently, those without write access to the wiki need to make a fork, make their changes, and then make an issue asking for a maintainer to take a look. The maintainer then, if they wish to merge the changes, has to look at the fork and do a merge manually. No PR also means that the change does not get put into the automated release notes, and the author doesn't get appropriate credit. 

Additionally, the submodule commit hash in the main repo has to be updated whenever the wiki is updated. This is done automatically by Dependabot, but is still rather annoying.

### Suggestion/proposed solution?

Migrate documentation to GitHub pages with a theme such as Just the Docs and with HTTPS enforcement on.

All documentation files would live within the docs folder of the repository.

This would allow for all documentation changes to be submitted via pull requests.
Additionally, pages provide for better flexibility and features extensions than just the GitHub wiki.

For instance, custom themes, custom HTML/CSS, custom build tools outside of Jekyll if necessary (shouldn't be needed), better search that isn't limited to page names, improved organization. Additionally, we can run Jekyll locally to get previews. The markdown of the wiki can also be rendered locally with some tools, but the GitHub-specific flare like the sidebar and style formatting is not visible until actually publishing it.

A potential theme is the very popular just the docs Jekyll theme. https://github.com/just-the-docs/just-the-docs
Migrating the existing content shouldn't be too hard as Jekyll and JTD support markdown, which is what is currently being used.

The main downside is that advertising of the wiki would be tricker. It would no longer be in the Wiki tab of the main repository. We can link to the page via the Read Me, and optionally via the old wiki as well. Additionally, deployment of the page will be slower than the wiki due to the build requirement. (JTD says potentially up to 10 mins)

Required tasks
- [x] Migrate old documentation to the new format
- [x] Enable GitHub pages and set it to deploy from /docs of the main branch.
- [x] Remove the wiki submodule from the wiki
- [x] Add page link to the read me and the repo about section
[...]

PR #641, opened 2024-10-14
[608] Migrate Documentation to GitHub Pages

hk21702:
Partially implements #608. Migrates documentation to use GitHub pages using the "Just the Docs" Jekyll theme. Also, includes a better formatted README file.

All information from the old wiki has been migrated. Assets have not been updated. Some content has been improved and or added, but most is the same from the old wiki. 

Some pages were broken up into sub-pages to take better advantage of the flexibility that GitHub pages gives us.

image
image


Patryk-Malinowski (commented):
What do you think about removing the table of contents for User and Dev Guide? And I only mean the table in the in the index.md files. I don't know how full these will get but I still imagine it will mainly be linking to other pages and would look cleaner and less cluttered without the TOC?

image

image




I feel like this gets worse the more links are added and can imagine a new user opening the page and being confused/overwhelmed? This can also maybe give the impression of the TOC links being different from the above links due to some differences in naming? Like ```Using the Rule Editor``` and ```Rule Editor``` could make a user think they are different links at first?

image

#########################################

Would it be a good idea to move the ```External Metadata``` away from basic usage?
In the image below, maybe at the end of that part say something like 'to learn more about Community rules...' etc. with a link so the users can read up about the steam workshop and community rules databases? 
image


However at the same time I guess we want to highlight why they should use them... so maybe a way of saying 'we strongly recommend you look at this link and download/enable this'.

Or maybe keep a short and really simple paragraph on each in Basic Usage explaining what they are used for and that we strongly recommend users set it up with a link leading to the steps on how to do this?

And we can keep more detailed information away from Basic Usage if they want to give that a read?

Patryk-Malinowski:
Oh and otherwise I think it looks great. Looking forward to when this gets merged :)

hk21702:
> What do you think about removing the table of contents for User and Dev Guide? And I only mean the table in the in the index.md files. I don't know how full these will get but I still imagine it will mainly be linking to other pages and would look cleaner and less cluttered without the TOC?

I initially wanted to do this, but as far as I can tell, you have to have a parent page when you categorize stuff like this. I could be wrong though and I'll double check.

hk21702:
> Would it be a good idea to move the External Metadata away from basic usage?
In the image below, maybe at the end of that part say something like 'to learn more about Community rules...' etc. with a link so the users can read up about the steam workshop and community rules databases?

We can. I'll look at changing it in this pr, but the goal was to just migrate existing stuff from the old wiki.

Patryk-Malinowski:
> I initially wanted to do this, but as far as I can tell, you have to have a parent page when you categorize stuff like this. I could be wrong though and I'll double check.


Just to make sure, I meant keep the parent/index.md files but remove TOC. I think I just used has_toc: false at the top of index.md to remove the TOC.

hk21702:
> Just to make sure, I meant keep the parent/index.md files but remove TOC. I think I just used has_toc: false at the top of index.md to remove the TOC.

Oh sorry, yeah can do that. Weird... for some reason that wasn't rendering at all for me on my side.

hk21702:
> > Would it be a good idea to move the External Metadata away from basic usage?
> > In the image below, maybe at the end of that part say something like 'to learn more about Community rules...' etc. with a link so the users can read up about the steam workshop and community rules databases?
> 
> We can. I'll look at changing it in this pr, but the goal was to just migrate existing stuff from the old wiki.

Looking at this again, I will just simplify it a little. Moving it out entirely is a little overkill as we already have a databases page. But, with the new format, it can be organized a lot better to make it easier on the eyes.

hk21702:
> > Just to make sure, I meant keep the parent/index.md files but remove TOC. I think I just used has_toc: false at the top of index.md to remove the TOC.
> 
> Oh sorry, yeah can do that. Weird... for some reason that wasn't rendering at all for me on my side.

Honestly, the automatic rendering you see here should be preferred, as it is automatic. So, instead, I'm going to axe the manual stuff I made back when I couldn't get it to show up on its own.

hk21702:
I updated a bit more content, but for this PR I'm going to pause on any content updates unless it is critical.

Patryk-Malinowski (approved):
Removing the manual links and just leaving the generated TOC was definitely the right call. Will save us the hassle of remembering and adding new links etc. as more subpages are added.

The external metadata part looks great too, I definitely think it is easier to digest now.

Patryk-Malinowski:
> We can. I'll look at changing it in this pr, but the goal was to just migrate existing stuff from the old wiki.

Yeah sorry that might have been going past the 'structure review'. Just wanted to mention it. Obviously once we are using this we can start updating the content and adding more etc.

PR #644, opened 2024-10-14
Minor fixes

LionelColaso:
- [x] Refactor Argument name and Variable in function to lowercase
Argument name and Variable in function should be lowercase

- [x]  Fix variable referenced before assignment
Fixes Local variable 'package_id' might be referenced before assignment

- [x] Boolean checks should not be inverted
It is needlessly complex to invert the result of a boolean comparison.

### DIFF
--- a/app/models/animations.py
+++ b/app/models/animations.py
@@ -97,9 +97,9 @@ class LoadingAnimation(QWidget):
 
         logger.debug("Finished LoadingAnimation initialization")
 
-    def check_animation_stop(self, frameNumber: int) -> None:
+    def check_animation_stop(self, frame_number: int) -> None:
         """Checks if the animation should stop."""
-        if self.animation_finished and frameNumber == self.movie.frameCount() - 1:
+        if self.animation_finished and frame_number == self.movie.frameCount() - 1:
             logger.debug("Animation is finished")
             self.stop_animation()
 
--- a/app/utils/metadata.py
+++ b/app/utils/metadata.py
@@ -2141,7 +2141,7 @@ class SteamDatabaseBuilder(QThread):
                 dynamic_query.pfids_by_appid()
                 # Make sure we have PublishedFileIds to work with...
                 if (
-                    not len(dynamic_query.publishedfileids) > 0
+                    len(dynamic_query.publishedfileids) == 0
                 ):  # If we didn't get any pfids
                     self.db_builder_message_output_signal.emit(
                         "Did not receive any PublishedFileIds from IPublishedFileService/QueryFiles! Cannot continue!"
--- a/app/controllers/mods_panel_controller.py
+++ b/app/controllers/mods_panel_controller.py
@@ -30,8 +30,8 @@ class ModsPanelController(QObject):
                 # Widget should always be of type ModListWidget
                 if isinstance(widget, ModListWidget):
                     package_id = widget.metadata_manager.internal_local_metadata[mod_data["uuid"]]["packageid"]
-                self._remove_from_all_ignore_lists(package_id)
-                logger.debug(f"Reset warning toggle for: {package_id}")
+                    self._remove_from_all_ignore_lists(package_id)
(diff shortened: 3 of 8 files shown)


hk21702:
Can you add a description of what this is fixing?

Also why replace json.dump with dumps and then putting that into write? I don't think there is a significant performance difference at our scale.

LionelColaso:
> why replace json.dump with dumps and then putting that into write? I don't think there is a significant performance difference at our scale.

The issue "Expected type 'SupportsWrite[str]', got 'TextIO' instead" is due to a type mismatch in the json.dump() function.
In Python 3.10 and later, the json.dump() function expects the second argument to be a file like object that supports writing strings (SupportsWrite[str]).
However, the output variable is a TextIO object, which is a file-like object that supports writing bytes.

twstagg:
> > why replace json.dump with dumps and then putting that into write? I don't think there is a significant performance difference at our scale.
> 
> The issue "Expected type 'SupportsWrite[str]', got 'TextIO' instead" is due to a type mismatch in the json.dump() function. In Python 3.10 and later, the json.dump() function expects the second argument to be a file like object that supports writing strings (SupportsWrite[str]). However, the output variable is a TextIO object, which is a file-like object that supports writing bytes.

Can you provide steps to reproduce this issue? I am on Windows 10, and simply deleting my empty userRules.json file and watching it get recreated without your patch.

app/utils/metadata.py
```diff
@@ -2141,7 +2141,7 @@ def run(self) -> None:
                 dynamic_query.pfids_by_appid()
                 # Make sure we have PublishedFileIds to work with...
                 if (
-                    not len(dynamic_query.publishedfileids) > 0
+                    len(dynamic_query.publishedfileids) <= 0
```
twstagg, line 2144: to simplify, it should probably just be `if len(dynamic_query.publishedfileids == 0` - i do get that this can be simplified though. It will never be less than 0.

app/models/animations.py
```diff
@@ -97,9 +97,9 @@ def __init__(self, gif_path: str, target: Callable[..., Any]):
 
         logger.debug("Finished LoadingAnimation initialization")
 
-    def check_animation_stop(self, frameNumber: int) -> None:
+    def check_animation_stop(self, framenumber: int) -> None:
```
twstagg, line 100: If you are not going to camel case, please use `frame_number` with an underscore.

twstagg (changes requested):

app/utils/metadata.py
LionelColaso: yeah right it will never be negitive so  i used your suggestion Thanks
if len(dynamic_query.publishedfileids == 0

app/models/animations.py
LionelColaso: yeah done thanks

LionelColaso:
> > > why replace json.dump with dumps and then putting that into write? I don't think there is a significant performance difference at our scale.
> > 
> > 
> > The issue "Expected type 'SupportsWrite[str]', got 'TextIO' instead" is due to a type mismatch in the json.dump() function. In Python 3.10 and later, the json.dump() function expects the second argument to be a file like object that supports writing strings (SupportsWrite[str]). However, the output variable is a TextIO object, which is a file-like object that supports writing bytes.
> 
> Can you provide steps to reproduce this issue? I am on Windows 10, and simply deleting my empty userRules.json file and watching it get recreated without your patch.

The file  creation works without the patch, this was to suppress pycharm from connstantly showing it as an issue and make it more complaint If its not needed i can simply remove the commit from the pull request

app/utils/metadata.py
hk21702: For safety its better to just check > since we don't have an unsigned type.

Though tbh I don't really see this as that big of a simplification from what it was before compared to what it is in this pr either way. Feels like semantics

LionelColaso: yeah sure

twstagg (approved):

PR #645, opened 2024-10-14
json_dump (just incase if required)

LionelColaso:
Expected type 'SupportsWrite[str]', got 'TextIO' instead

The issue "Expected type 'SupportsWrite[str]', got 'TextIO' instead" is due to a type mismatch in the json.dump() function. In Python 3.10 and later, the json.dump() function expects the second argument to be a file like object that supports writing strings (SupportsWrite[str]). However, the output variable is a TextIO object, which is a file-like object that supports writing bytes.

### DIFF
--- a/app/controllers/app_controller.py
+++ b/app/controllers/app_controller.py
@@ -36,7 +36,7 @@ class AppController(QObject):
         if not user_rules_path.exists():
             initial_rules_db = DEFAULT_USER_RULES
             with open(user_rules_path, "w", encoding="utf-8") as output:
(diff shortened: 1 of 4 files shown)


twstagg:
I don't think this would be needed.` json.dump` already writes the json data directly to the file, so wrapping it with `output.write` doesn't provide any additional functionality. In fact, it could potentially make the code slightly less efficient, as it involves an extra function call. If there is a real reason to wrap it then i think that we could discuss in the future for something relevant.

PR #648, opened 2024-10-15
[513] Additional Shortcuts and Improved Organization

LionelColaso:
- [x]  Added additional shortcuts under File > Open... > RimWorld
     Root Directory
    Config Directory
    Local Mods Directory
    Steam Mods Directory

- [x] Organized shortcuts such that related shortcuts are grouped under an additional folder
   RimSort
   RimWorld
   

- [x] Added Notification to set paths if they not set

image
image

### DIFF
--- a/app/controllers/menu_bar_controller.py
+++ b/app/controllers/menu_bar_controller.py
@@ -12,7 +12,12 @@ from app.views.menu_bar import MenuBar
 
 
 class MenuBarController(QObject):
-    def __init__(self, view: MenuBar, settings_controller: SettingsController, mods_panel_controller: ModsPanelController) -> None:
+    def __init__(
+        self,
+        view: MenuBar,
+        settings_controller: SettingsController,
+        mods_panel_controller: ModsPanelController,
+    ) -> None:
         super().__init__()
 
         self.menu_bar = view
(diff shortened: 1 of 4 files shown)


hk21702 (dismissed):
I'd remove the separator between RimSort and RimWorld as its just two options. The separation isn't adding any clarity but is taking up space.

Additionally, the labels are getting kind of long. Consider an idiomatic approach where you rename "Shortcuts" to "Open...", leave RimSort and RimWorld alone, but for the final buttons, don't have "open" or "RimWorld/RimSort". The categorization itself should be clear enough to denote what you're accessing. 

e.g., instead of "Shortcuts > RimSort > Open RimSort Directory", do "Open... > RimSort > Root Directory".

LionelColaso:
> I'd remove the separator between RimSort and RimWorld as its just two options. The separation isn't adding any clarity but is taking up space.
> 
> Additionally, the labels are getting kind of long. Consider an idiomatic approach where you rename "Shortcuts" to "Open...", leave RimSort and RimWorld alone, but for the final buttons, don't have "open" or "RimWorld/RimSort". The categorization itself should be clear enough to denote what you're accessing.
> 
> e.g., instead of "Shortcuts > RimSort > Open RimSort Directory", do "Open... > RimSort > Root Directory".

yeah Done

Patryk-Malinowski (commented):
lgtm

Wondering if we should notify user if for example they attempt to open steam mods folder, but its blank/incorrect do we display error window and say check path etc.? It could be confusing when they press ```open something``` and just nothing happens.

I can see none of those shortcut options did that originally, just logged an error. So if we want to do this you would want to add it to all of them.

LionelColaso:
> lgtm
> 
> Wondering if we should notify user if for example they attempt to open steam mods folder, but its blank/incorrect do we display error window and say check path etc.? It could be confusing when they press `open something` and just nothing happens.
> 
> I can see none of those shortcut options did that originally, just logged an error. So if we want to do this you would want to add it to all of them.

yeah sure will add notifications if the paths are not set

LionelColaso:
> lgtm
> 
> Wondering if we should notify user if for example they attempt to open steam mods folder, but its blank/incorrect do we display error window and say check path etc.? It could be confusing when they press `open something` and just nothing happens.
> 
> I can see none of those shortcut options did that originally, just logged an error. So if we want to do this you would want to add it to all of them.

Added notification to ask user if they want to set path if they don't exist

Patryk-Malinowski (approved):
lgtm

PR #660, opened 2024-10-21
Add completion notification  to runner panel

LionelColaso:
Added Operation Completed Notification to runner_pannel to let user know its safe to close the window

Screenshot 2024-10-21 150928

image

### DIFF
--- a/app/windows/runner_panel.py
+++ b/app/windows/runner_panel.py
@@ -20,7 +20,11 @@ from app.utils.app_info import AppInfo
 from app.utils.steam.webapi.wrapper import (
     ISteamRemoteStorage_GetPublishedFileDetails,
 )
-from app.views.dialogue import show_dialogue_conditional, show_dialogue_file
+from app.views.dialogue import (
+    BinaryChoiceDialog,
+    show_dialogue_conditional,
+    show_dialogue_file,
+)
 
 
 class RunnerPanel(QWidget):
@@ -404,3 +408,11 @@ class RunnerPanel(QWidget):
 
         # Cleanup process
         self.process.terminate()
+        diag = BinaryChoiceDialog(
+            title="Process Complete",
+            text="Process complete, you can close the window.",
+            positive_text="Close Window",
+            negative_text="Ok",
+        )
+        if diag.exec_is_positive():
+            self.close()


LionelColaso:
> Typo: "Operation" should not be capitalised.

its already capital ?

app/windows/runner_panel.py
```diff
@@ -402,5 +406,9 @@ def finished(self) -> None:
             if "todds" in self.process.program():
                 self.change_progress_bar_color("success")
 
-        # Cleanup process
+        # Cleanup process and show notification
         self.process.terminate()
+        show_information(
+            title="Operation completed",
+            text="Operation completed, you can close this window.",
```
hk21702, line 413:
```suggestion
            text="Operation completed, you may close this window.",
```

hk21702 (changes requested):

app/windows/runner_panel.py
LionelColaso: i just used what i saw in other apps if u still prefer i will cghange "can" to "may"
image

app/windows/runner_panel.py
```diff
@@ -402,5 +406,9 @@ def finished(self) -> None:
             if "todds" in self.process.program():
                 self.change_progress_bar_color("success")
 
-        # Cleanup process
+        # Cleanup process and show notification
         self.process.terminate()
+        show_information(
```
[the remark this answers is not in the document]
LionelColaso: yeas sure any specifics you want to name the two buttons ?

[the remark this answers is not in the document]
hk21702: See previous comment for example.

twstagg (approved):
Looks good to me. I approve pending @hk21702 approval.

It should be noted that this implementation only pops this message up when the runner panel has a `QProcess` associated, i.e. `steamcmd` or `todds`.

This does not address the original issue that was reported with DB Builder as it does not use the `QProcess`. We can address that later though. When DB Builder is finished, we will need to implement this sort of thing separately. I do need to finish making the process pool able to send messages back to the UI, so i can implement something accordingly when i complete that work.

app/windows/runner_panel.py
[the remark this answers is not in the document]
LionelColaso: done

LionelColaso:
> Looks good to me. I approve pending @hk21702 approval.
> 
> It should be noted that this implementation only pops this message up when the runner panel has a `QProcess` associated, i.e. `steamcmd` or `todds`.
> 
> This does not address the original issue that was reported with DB Builder as it does not use the `QProcess`. We can address that later though. When DB Builder is finished, we will need to implement this sort of thing separately. I do need to finish making the process pool able to send messages back to the UI, so i can implement something accordingly when i complete that work.

ohh i  thought  def finished(self)  was common for all

app/windows/runner_panel.py
```diff
@@ -404,3 +404,10 @@ def finished(self) -> None:
 
         # Cleanup process
         self.process.terminate()
+        answer = show_dialogue_conditional(
+            title="Operation completed",
+            text="Operation completed, you can close the window.",
+            information="Do you want to close this window.",
```
hk21702, line 410: Punctuation

app/windows/runner_panel.py
```diff
@@ -404,3 +404,10 @@ def finished(self) -> None:
 
         # Cleanup process
         self.process.terminate()
+        answer = show_dialogue_conditional(
```
hk21702, line 407: Just use a BinaryChoiceDialog or show_dialogue_conditional, omit the information part, and label the two buttons in a clear manner like the examples I gave before.

hk21702 (changes requested):

app/windows/runner_panel.py
LionelColaso: done

app/windows/runner_panel.py
LionelColaso: done

app/windows/runner_panel.py
```diff
@@ -404,3 +404,9 @@ def finished(self) -> None:
...
+            text="Operation complete, you can close the window.",
+            button_text_override=["Close Window"],
+        )
+        if answer == "Close Window":
+            self.close()
```
hk21702, line 412: I think show_dialogue_conditional doesn't allow you to override the cancel button text. Try the following with BinaryChoiceDialog. Also, add a title and rename operation to process for consistency with the phrasing used in the runner panel logs.

```suggestion
        diag = BinaryChoiceDialog(
            title="Process Complete",
            text="Process complete, you can close the window.",
            positive_text="Close Window",
            negative_text="Ok",
        )
        if diag.exec_is_positive():
            self.close()
```

hk21702 (changes requested):

app/windows/runner_panel.py
LionelColaso: ok i just used your suggestion and done with it

also seems show_dialogue_input is depriciated ?
image

hk21702: Yes. That is intended. It's just a wrapper around QInputDialog right now and doesn't add anything new.

hk21702 (approved):

PR #664, opened 2024-10-24
Report RimSort version at init of application in logs

hk21702:
Report RimSort version at init of application in logs at info level. Replaces the existing debug level RimSort application init log.

This is to make it a little easier to analyze user reports when logs are provided.

Example when the version is unknown:
image

### DIFF
--- a/app/__main__.py
+++ b/app/__main__.py
@@ -208,5 +208,5 @@ if __name__ == "__main__":
 
         logger.debug("Running using Nuitka bundle")
 
-    logger.debug("Initializing RimSort application")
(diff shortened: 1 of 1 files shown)


LionelColaso (approved):
is there a way we could make it so that if the version number is unknown it uses the commit is displayed ?

hk21702:
> is there a way we could make it so that if the version number is unknown it uses the commit is displayed ?

Only if locally it is an actual git repository. And then I guess you could try and look at it using pygit2/gitpython.

Issue #57, opened 2023-03-31
Implement setting to change font scale


MitztheKat:
The GUI is good - other than the fact it's very tiny compared to rimpy. A setting to change font scale that's graphical would be great!

PR #667, opened 2024-10-28
Ask to enable steam integration at setup

hk21702:
At the initial setup of an instance, display a dialog asking if the user would like to enable steam integration. The only checks for displaying this or not are if this is initial setup of this instance, and if steam integration is already on for whatever reason (for grandfathering in existing setups). 

image

Internally, this introduces a new attribute at the instance level to track whenever or not this was the initial setup. At init of the object it is True, but after all setup steps in main_window, it is forced to False. The current implementation violates MVC a bit and will need to be potentially overhauled more deeply in a future refactor.

Also, improves the internals of instance's `as_dict` so it does not need to be manually expanded whenever a new attribute is added to `instance`.

### DIFF
--- a/app/models/instance.py
+++ b/app/models/instance.py
@@ -39,6 +39,9 @@ class Instance(msgspec.Struct):
 
     :param steam_client_integration: Whether to integrate the Steam client with the instance
     :type steam_client_integration: bool
+
+    :param initial_setup: Whether the instance is in the initial setup state
+    :type initial_setup: bool
     """
 
     name: str = "Default"
@@ -54,6 +57,8 @@ class Instance(msgspec.Struct):
     steamcmd_ignore: bool = False
     steam_client_integration: bool = False
 
+    initial_setup: bool = True
+
     def __setattr__(self, name: str, value: Any) -> None:
         # If the value is the same as the current value, do nothing
         if getattr(self, name) == value:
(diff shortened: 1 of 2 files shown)


hk21702:
As a side note, I'd like feedback on whenever or not all dialog str at the text level (first main line) should be big and bold like in the screenshot for better differentiation. In the image, it is h2.

Limfjorden:
I think the text should provide examples of features that require steam. I would change the text to something like

`This will, among other things, allow you to use mods from the workshop by subscribing on steam as you would normally.`

hk21702:
> > As a side note, I'd like feedback on whenever or not all dialog str at the text level (first main line) should be big and bold like in the screenshot for better differentiation. In the image, it is h2.
> 
> personally not a fan of h2 but its better if its done in #57

I can make it h3 instead. #57 is talking about something different and is a global scale of font size. This would only affect dialog text.

hk21702:
> I think the text should provide examples of features that require steam. I would change the text to something like
> 
> `This will, among other things, allow you to use mods from the workshop by subscribing on steam as you would normally.`

image

LionelColaso:
> I can make it h3 instead. #57 is talking about something different and is a global scale of font size. This would only affect dialog text.

Use size u think looks, fits and allings better in the text properly  in the diag

LionelColaso (approved):

Issue #663, opened 2024-10-24
Latest Edge CTD


GitWolf22k:
### Release Type

Compiled (Windows x86-64)

### Version

v1.0.11-edge7+

### Operating System

Win 10

### Relevant logs

https://0x0.st/XUP9.log

### Describe the bug

CTD when sort button used after modlist reduction by 50%. App will not open unless settings.json is deleted.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

hk21702:
I believe this is an issue caused when upgrading from a prior edge release. The steamcmd_auto_clear_depot_cache was introduced earlier, but then moved to be an instance setting. Thus, if you had a settings file that had steamcmd_auto_clear_depot_cache set, but not under the edge7 schema of being an instance setting, then you get a crash.

hk21702:
Weirdly, I can't replicate this issue. Based on the logs, there is a crash when loading the settings file if it contains an unrecognized entry. However, the existing code should already have a mitigation for this. Additionally, I can't replicate this crash if I try to mess with my own settings file, the mitigation seems to work. This is both with the compiled edge release, and running from interpreter at head of main.

@GitWolf22k do you happen to have your old settings file still? If so, do you mind uploading it? You are welcome to edit out any identifying info in it.

GitWolf22k:
Sure. Here tis.
[settings.json](https://github.com/user-attachments/files/17502918/settings.json)

hk21702:
Unfortunately, I still can't replicate the crash. Will leave the issue open for now in case something new happens

PR #669, opened 2024-10-31
Fix crash when __setattr__ for instances, where the key does not exist.

hk21702:
Fixes crash when using __setattr__ for instances, where the key does not exist.

This should fix crashes where there is an unexpected entry in the settings file for an instance. This copies the behavior for settings that are global scoped. The unexpected entry is simply ignored.

### DIFF
--- a/app/models/instance.py
+++ b/app/models/instance.py
@@ -59,11 +59,11 @@ class Instance(msgspec.Struct):
 
     initial_setup: bool = True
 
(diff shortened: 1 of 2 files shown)


hk21702:
I suspect this fixes and closes #663

Patryk-Malinowski (approved):
Not able to replicate #663 but it works in regards to unexpected fields in settings.json under any of the instances.

PR #674, opened 2024-11-06
Fix save button blinking

### LINKED ISSUE #618: [Bug]: Blinking save btn when active list is only modified visually
### Release Type

Compiled Windows x86 64-bit

### Version

v1.0.10 edge 51

### Operating System

Win 11

### Relevant logs

_No response_

### Describe the bug

Currently, the save btn flashes when filters are changed for the active list (searching mods, changing shown mod types) even when there is no real change to the actual modlist. Additionally, the flashing does not go away until a save action is done, even if the filter actions are undone.

Note: This bug doesn't seem to be 100% consistent. Sometimes it occurs, othertimes the expected behavior occurs.

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

Patryk-Malinowski:
Closes #618

When saving the modlist, it didnt save the mod uuids to a list that it used for comparison later hence triggering the animation. The save list seemed to be updated on refresh only (or if opening up rimsort, because it calls the ```_do_refresh``` method)

This means if you open up rimsort and start using the filters, it works as expected. But once you make changes and 'save' is when the issues can be reproduced.

Video replicating the issue.

https://github.com/user-attachments/assets/7cad045f-073a-4b50-8659-615f905e3578

No harm checking if the save button animation can be triggered differently from what Is shown in the video to make sure this fix still addresses it...

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -1761,7 +1761,7 @@ class MainContent(QObject):
                         active_mods.append(package_id + "_steam")
                         continue  # Append `_steam` suffix if Steam mod, continue to next mod
                 active_mods.append(package_id)
-        active_mods_uuids = metadata.get_mods_from_list(mod_list=[mod for mod in active_mods])[0]
+        active_mods_uuids, inactive_mods_uuids, _, _ = metadata.get_mods_from_list(mod_list=[mod for mod in active_mods])
         self.active_mods_uuids_last_save = active_mods_uuids
         logger.info(f"Collected {len(active_mods)} active mods for saving")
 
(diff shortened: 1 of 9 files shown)


Patryk-Malinowski:
Additionally save the modlists to their restore states/variables as previously it was only done on startup.

LionelColaso (approved):
Works nice catch

app/views/main_content_panel.py
```diff
@@ -1761,6 +1761,8 @@ def _do_save(self) -> None:
                         active_mods.append(package_id + "_steam")
                         continue  # Append `_steam` suffix if Steam mod, continue to next mod
                 active_mods.append(package_id)
+        active_mods_uuids, inactive_mods_uuids, _, _ = metadata.get_mods_from_list(mod_list=[mod for mod in active_mods])
```
hk21702, line 1764: Have not been able to test on my own, but is it necessary to make a copy of the active_mods list?

Patryk-Malinowski: No it is not. Great catch.

PR #681, opened 2024-11-09
Add Python shebang to main.py

yodatak:
Hello i use is to try to make a flatpak with 
      - install -D app/__main__.py /app/bin/RimSort
and i think its usesfull for all

Thanks !


hk21702:
Considering that this is for flatpack (linux) purposes, it might be appropriate to use `python3` instead of just `python`.

LionelColaso (approved):
Looks Good Now

PR #682, opened 2024-11-09
Fix searching in active and inactive modlists

Patryk-Malinowski:
When searching in the active modlist, any mods with errors would show up regardless of your search criteria (including filters).

image
image

It prevents invalid items from showing up when using any filters too.
image
image


LionelColaso:
already fixed in #653

Patryk-Malinowski:
> Should be already fixed in #653 can you check ? If not then Good Catch

On my end it still seems to pick it up:
image

image

LionelColaso:
yeah then go for it :-)

hk21702:
I haven't tested or looked at this more deeply, but just as a note to anyone reviewing this, check if this change breaks other stuff related to invalid mods.

Patryk-Malinowski:
For clarity see below. 

This is separate from the ```invalid``` property used in ```internal_local_metadata```. The one in ```internal_local_metadata``` would refer to an invalid mod if it is missing an about.xml for example and would be displayed like this:
image


But the ```invalid``` property used here is from a QListWidgetItem and it is used to indicate whether a mod has errors or not.
image


I do think the QListWidgetItem one could benefit with being renamed, to prevent confusion with the ```invalid``` property used in ```internal_local_metadata```

[Picture](https://github.com/user-attachments/assets/3593e294-24bc-4b74-9612-03bdd4a204b5) attempting to showcase the difference between the data with a mod that has errors but is valid. 

Also a [another picture](https://github.com/user-attachments/assets/22680195-196d-4f62-abc9-3c24199083ea) showcasing an invalid item (missing about.xml).

LionelColaso (approved):
Works and did not find any issues for now

PR #685, opened 2024-11-13
[678] Implement warning when Rentry response code is not 200. Fix Rentry import action duplication.

### LINKED ISSUE #678: Inform of Error when Reentry Import Fails with Status other than 200
### Release Type

Compiled (Windows x86-64)

### Version

v1.0.10

### Operating System

Windows 11

### Relevant logs

_No response_

### Describe the bug

**Issue:**

Reentry Import Fails silently if request to reentry returns status code != 200. 

As long as an exception is not thrown during the request, the code will return without logging any error after [here](https://github.com/RimSort/RimSort/blob/1a380b5ad026d8cf7c74eebd82727403901ff267/app/utils/rentry/wrapper.py#L164-L176). 

As such `self.package_ids` is still empty when `import_rentry_link()` returns
and `_do_import_list_rentry()` returns early [here](https://github.com/RimSort/RimSort/blob/1a380b5ad026d8cf7c74eebd82727403901ff267/app/views/main_content_panel.py#L1329C8-L1329C42).

**Expected Behavior:**

If reentry does not return 200 then either log an error or open a popup informing the user what the error response was. 

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

hk21702:
Closes #678 

Implements a warning dialog and log input whenever the response code from Rentry when importing from Rentry is not 200.

Also fixes a bug where the Rentry import process was duplicated and occurred twice per link.


Patryk-Malinowski:
Did not review yet but found this #687 

In main that window pops up twice, in this PR it pops up once. Maybe you already fixed it but missed something? Like I said didn't have a chance to review yet but just mentioning this.

LionelColaso:
Whole Rentry Import is Broken so you will have to fix that has well
attached video below
https://drive.google.com/file/d/1tD91qE-PICjiMRy9xdy93pMpiOYpGXze/view?usp=sharing

hk21702:
> Whole Rentry Import is Broken so you will have to fix that has well attached video below https://drive.google.com/file/d/1K8AiSw_YyqSjcigR_KrjhAJhDAA7q2Cq/view?usp=sharing

That is just Rentry returning a 403 response code. As far as I can tell, there isn't much we can do about it. Will probably need to do something with the request headers to get it to give a valid response. Either way, is out of scope for this PR.

LionelColaso (approved):
Atleast now user will  know the reason for failure

PR #690, opened 2024-11-14
[608] Update Wiki Links

hk21702:
Updates wiki links

 - Read Me
 - Issues Templates
 - Issues Page
 - App Wiki Link
 
 Also cleans out the wiki entry in .gitmodule


LionelColaso (approved):
Works

PR #691, opened 2024-11-15
rentry: avoid crashes

LionelColaso:
- [x]  Added Doc strings

- [x]  Avoids crash when there is no response or network issue

- [x]  Reopen Input dialog if invalid rentry link while rentry import

- [x]  Added TODO so to remind why rentry import is broken

### DIFF
--- a/app/utils/rentry/wrapper.py
+++ b/app/utils/rentry/wrapper.py
@@ -14,16 +14,24 @@ from app.views.dialogue import (
     show_warning,
 )
 
-# Constants
+# Constants for Rentry API endpoints
 BASE_URL = "https://rentry.co"
-BASE_URL_RAW = f"{BASE_URL}/raw"
 API_NEW_ENDPOINT = f"{BASE_URL}/api/new"
 _HEADERS = {"Referer": BASE_URL}
+# TODO: rentry-auth code
+"""
+Contact (support@rentry.co) for auth code to use here.
+This header then gives access to all posts at /raw)
+Need this auth code for rentry import to work
+Info about the issue https://github.com/radude/rentry/issues/34
+"""
 
 
 class HttpClient:
+    """A simple HTTP client for making requests to the Rentry API."""
+
     def __init__(self) -> None:
-        # Initialize a session for making HTTP requests
+        """Initialize a session for making HTTP requests."""
         self.session = requests.Session()
 
     def make_request(
@@ -33,14 +41,26 @@ class HttpClient:
         data: dict[str, Any] | None = None,
         headers: dict[str, str] | None = None,
     ) -> requests.Response:
-        # Perform a HTTP request and return the response
+        """
+        Perform an HTTP request and return the response.
+
+        Args:
+            method (str): The HTTP method (GET, POST, etc.).
+            url (str): The URL for the request.
+            data (dict | None): Optional data to send with the request.
+            headers (dict | None): Optional headers to include in the request.
+
+        Returns:
+            requests.Response: The response object from the request.
+        """
         headers = headers or {}
         request_method = getattr(self.session, method.lower())
         response = request_method(url, data=data, headers=headers)
-        response.data = response.text
+        response.data = response.text  # Store the response text for convenience
         return response
 
     def get(self, url: str, headers: dict[str, str] | None = None) -> requests.Response:
+        """Perform a GET request."""
         return self.make_request("GET", url, headers=headers)
 
     def post(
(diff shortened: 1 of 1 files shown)


app/utils/rentry/wrapper.py
```diff
@@ -64,53 +73,95 @@ def __init__(self, text: str):
 
         try:
             response = self.new(text)
+            self.response = response
```
Patryk-Malinowski, line 76: I dont think we need this. The finally block should be able to use the local ```response``` variable directly.

Patryk-Malinowski (changes requested):
Currently we get warnings about network connection twice when exporting. One from the ```new()``` method 
and one after it checks the response from that method.
image

I think we should probably have the network warning pop up only once.


We get a third warning about failure to upload but that should be ok to leave as it could cover other failures.

Patryk-Malinowski:
Oh also at first I though this issue was in main, but if you cancel/exit the window you are stuck in a loop of this:
image

Patryk-Malinowski:
> Oh also at first I though this issue was in main, but if you cancel/exit the window you are stuck in a loop of this: image
> 
> I know main has this issue but without looping. I might look into fixing that thing.

Basically have this done. Will make a PR today/tomorrow for it in case you want to wait for that to get merged before making changes to your PR.

app/utils/rentry/wrapper.py
```diff
@@ -34,6 +34,15 @@ def make_request(
         headers: dict[str, str] | None = None,
     ) -> requests.Response:
         # Perform a HTTP request and return the response
```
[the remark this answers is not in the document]
LionelColaso: OppS Thanks

app/utils/rentry/wrapper.py
LionelColaso: Ok removed

app/utils/rentry/wrapper.py
```diff
@@ -64,53 +73,95 @@ def __init__(self, text: str):
 
         try:
             response = self.new(text)
+            self.response = response
+
+            # show warning and exit if response is None
```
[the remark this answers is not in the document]
LionelColaso: Done

LionelColaso:
> > Oh also at first I though this issue was in main, but if you cancel/exit the window you are stuck in a loop of this: image
> > I know main has this issue but without looping. I might look into fixing that thing.
> 
> Basically have this done. Will make a PR today/tomorrow for it in case you want to wait for that to get merged before making changes to your PR.
> 
> PR made. #714

I removed the loop and also your PR #714 looks good 
Hopefully i did not miss anything now

Patryk-Malinowski:
I merged #714 

Check if you need to make any changes because of it and re-request review when ready :)

LionelColaso:
> I merged #714
> 
> Check if you need to make any changes because of it and re-request review when ready :)

already done

app/utils/rentry/wrapper.py
```diff
@@ -108,100 +140,150 @@ def new(self, text: str) -> Any:
...
+        Args:
+            e (Exception): The exception that occurred during the network operation.
+        """
+        logger.error(f"A network error occurred while processing Rentry: {str(e)}")
+        show_fatal_error(
+            title="Network Error",
+            text="A network error occurred while processing Rentry, Please check if your internet is working",
+            details=f"{str(e)}",
+        )
+        return None  # Return None to indicate failure
```
Patryk-Malinowski, line 286: Change this to handle exceptions in a generic way. It might be due to network error, or something completely else. Just use generic language saying error/exception occurred, upload your logs to discord/github  etc.

app/utils/rentry/wrapper.py
```diff
@@ -108,100 +140,150 @@ def new(self, text: str) -> Any:
...
 
         except Exception as e:
-            logger.error(
-                f"An error occurred while fetching rentry.co content: {str(e)}"
-            )
-            show_fatal_error(
-                title="Error",
-                text=f"An error occurred: {str(e)}",
-            )
+            RentryError().show_error_warning(e)  # Handle exceptions during the process
```
Patryk-Malinowski, line 221: This should handle normal exceptions. Add except for network exceptions like ```RequestException``` and then create a method to handle the network exceptions (show warning etc. whatever it needs to do)

app/utils/rentry/wrapper.py
Patryk-Malinowski: Maybe rename the method to something simpler too. Since ```show_error_warning``` is unclear whether its an error or a warning, especially with the comment saying its a warning. Maybe a generic ```show_error``` or ```show_generic_error```?

app/utils/rentry/wrapper.py
```diff
@@ -49,57 +69,69 @@ def post(
...
-                self.handle_upload_failure(response)
-            else:
-                self.upload_success = True
-                url = response.get("url")
-                self.url = url if isinstance(url, str) else None
-        finally:
+            response = self.new(text)  # Attempt to upload the text
+            self.handle_response(response)  # Handle the response from the upload
+        except Exception as e:
+            RentryError().show_error_warning(e)  # Handle upload failure
```
Patryk-Malinowski, line 103: This should handle normal exceptions. Add except for network exceptions like RequestException and then create a method to handle the network exceptions (show warning etc. whatever it needs to do)

Patryk-Malinowski (changes requested):
Look at comments. 

I guess network error would imply it should be displayed as an error, however the show_fatal_error dialogue also include a message telling the user to report this to the devs, but in the case of a network error its probably just on their end... 

In short I think a warning dialogue would be better for network connection errors?

app/utils/rentry/wrapper.py
LionelColaso: changes done

app/utils/rentry/wrapper.py
LionelColaso: done

app/utils/rentry/wrapper.py
LionelColaso: done

LionelColaso:
> Look at comments.
> 
> I guess network error would imply it should be displayed as an error, however the show_fatal_error dialogue also include a message telling the user to report this to the devs, but in the case of a network error its probably just on their end...
> 
> In short I think a warning dialogue would be better for network connection errors?

You are right user does not need to report network errors related to connection my changes should be good now

app/utils/rentry/wrapper.py
```diff
@@ -108,100 +141,158 @@ def new(self, text: str) -> Any:
...
+        """
+        Show a warning for network errors.
+
+        Args:
+            e (Exception): The exception that occurred during the network operation.
+        """
+        logger.error(f"A network error occurred while processing Rentry: {str(e)}")
+        show_warning(
+            title="Network Error",
+            text="Network error occurred while processing Rentry, Please check your internet connection",
```
Patryk-Malinowski, line 292: Just add a full stop :)

Patryk-Malinowski (commented):
Looks good, personally I would like to additionally see a basic except for general exceptions wherever it has one for handling RequestExceptions:
image

So it would be something like this:
image

I guess its just an additional layer to prevent a crash?

LionelColaso:
> Looks good, personally I would like to additionally see a basic except for general exceptions wherever it has one for handling RequestExceptions: image
> 
> So it would be something like this: image
> 
> I guess its just an additional layer to prevent a crash?

ok done

Patryk-Malinowski:
> ok done

Can you just throw in a fatal error window after the exception just so the user isnt left in the dark about an exception occuring, because they wont see the logger stuff. Again just something generic like ```An exception occurred when attempting rentry export/import etc.``` and they can get in touch with devs then...

LionelColaso:
> > ok done
> 
> Can you just throw in a fatal error window after the exception just so the user isnt left in the dark about an exception occuring, because they wont see the logger stuff. Again just something generic like `An exception occurred when attempting rentry export/import etc.` and they can get in touch with devs then...

Yes Sure Done

Patryk-Malinowski (approved):
LGTM

PR #706, opened 2024-12-09
Critical: Fix workflows

LionelColaso:
- [x] Fixes Issue that causes "Mac" build fails due to recent changes in Nuitka 
fixes error
FATAL: Cannot use both '--mode=' and deprecated options that specify mode.

- [x] For Pytest replace `libgl1-mesa-glx` with` libgl1 libglx-mesa0`
fixes error 
E: Package 'libgl1-mesa-glx' has no installation candidate

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,8 +150,7 @@ jobs:
         with:
           nuitka-version: main
           script-name: app/__main__.py
-          onefile: false
-          mode: standalone
+          mode: ${{ runner.os == 'macos' && 'app' || 'standalone' }}
           static-libpython: auto
           product-version: >-
             ${{ steps.sem_version.outputs.major }}.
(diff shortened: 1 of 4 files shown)


Patryk-Malinowski (approved):
LGTM. I have no idea if there are benefits to having it ```standalone``` vs ```app``` on Mac. Ideally we would stick with what we had before, but I assume once we figure out how to make it work standalone on Mac again we can always update it. 

Unless someone else could chime in regarding the impact of this change from ```standalone -> app```, I am happy enough with this as a fix, at least for now.

If I have some time I might look more into standalone vs app out of interest...

LionelColaso:
> LGTM. I have no idea if there are benefits to having it `standalone` vs `app` on Mac. Ideally we would stick with what we had before, but I assume once we figure out how to make it work standalone on Mac again we can always update it.
> 
> Unless someone else could chime in regarding the impact of this change from `standalone -> app`, I am happy enough with this as a fix, at least for now.
> 
> If I have some time I might look more into standalone vs app out of interest...

You can refer to the issue [here](https://github.com/Nuitka/Nuitka/issues/3237)

PR #728, opened 2025-01-01
Add setting to hide invalid mods when filtering

Patryk-Malinowski:
Enabled by default
Located in Advanced tab
image

If enabled it will hide invalid items when any filters are active:
image
image
image

Otherwise it shows them:
image
image
image
image

Behavior should apply to active and inactive modlists.

### DIFF
--- a/app/controllers/settings_controller.py
+++ b/app/controllers/settings_controller.py
@@ -512,6 +512,9 @@ class SettingsController(QObject):
         self.settings_dialog.mod_type_filter_checkbox.setChecked(
             self.settings.mod_type_filter_toggle
         )
+        self.settings_dialog.hide_invalid_mods_when_filtering_checkbox.setChecked(
+            self.settings.hide_invalid_mods_when_filtering_toggle
+        )
         self.settings_dialog.show_duplicate_mods_warning_checkbox.setChecked(
             self.settings.duplicate_mods_warning
         )
@@ -655,6 +658,9 @@ class SettingsController(QObject):
         )
         self.settings.mod_type_filter_toggle = (
             self.settings_dialog.mod_type_filter_checkbox.isChecked()
+        )
+        self.settings.hide_invalid_mods_when_filtering_toggle = (
+            self.settings_dialog.hide_invalid_mods_when_filtering_checkbox.isChecked()
         )
         self.settings.duplicate_mods_warning = (
             self.settings_dialog.show_duplicate_mods_warning_checkbox.isChecked()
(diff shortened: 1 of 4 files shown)


Patryk-Malinowski:
I think it should be on by default, maybe even not an option because why would you want invalid items showing up when searching for something? But no harm having the setting available. Maybe a future feature to show all invalid items and their paths etc...? Lmk what ye think

LionelColaso:
> I think it should be on by default, maybe even not an option because why would you want invalid items showing up when searching for something? But no harm having the setting available. Maybe a future feature to show all invalid items and their paths etc...? Lmk what ye think

Showing invalid mods is essential I think otherwise users won't know if any mod is corrupt or invalid
it should be disabled by default unless someone really want not to know about them
you could add a notification if there are invalid mods ?

Patryk-Malinowski:
> Showing invalid mods is essential I think otherwise users won't know if any mod is corrupt or invalid it should be disabled by default unless someone really want not to know about them you could add a notification if there are invalid mods ?

Just to make sure, with this enabled on my default you will not see invalid items ONLY when applying filters (Searching, or mod type filter etc..)

Thinking about it, maybe it should be off by default in case you actually want to search for invalid items or when sorting by ```Showing git mods``` for example you see all the invalid mods are all .git and you can start digging deeper then etc.

LionelColaso:
> > Showing invalid mods is essential I think otherwise users won't know if any mod is corrupt or invalid it should be disabled by default unless someone really want not to know about them you could add a notification if there are invalid mods ?
> 
> Just to make sure, with this enabled on my default you will not see invalid items ONLY when applying filters (Searching, or mod type filter etc..)
> 
> Thinking about it, maybe it should be off by default in case you actually want to search for invalid items or when sorting by `Showing git mods` for example you see all the invalid mods are all .git and you can start digging deeper then etc.
> 
> Also most users probably would never realize this setting option would be there, so they could miss it and never realize how many invalid items they have...
> 
> Ill change it to off by default, and if someone ever wants this functionality the setting is there.

I like the idea of this being off by default the pull is good for merge but
what do you think about a setting to notify user if invalid mods are detected ?
if yes do you want to add it in the same pull or make another?

Patryk-Malinowski:
> what do you think about a setting to notify user if invalid mods are detected ?
> 
> if yes do you want to add it in the same pull or make another?


I would do it in a different pull.

When would it check for invalid mods? Only at startup or at startup + on refresh etc.

LionelColaso:
> > what do you think about a setting to notify user if invalid mods are detected ?
> > if yes do you want to add it in the same pull or make another?
> 
> I would do it in a different pull.
> 
> When would it check for invalid mods? Only at startup or at startup + on refresh etc.

It can be on startup, refresh or sorting where ever you think best (it should be enabled by default)

LionelColaso (approved):
Good to go

PR #731, opened 2025-01-02
Minor fixes for mod_type filter

LionelColaso:
1) Toggle now works for mod_type filter
2) Add Label indicating some settings require Restart of App

with "Enable mod type filter" Toggle Enabled
image

with "Enable mod type filter" Toggle Disabled
image

New Label for Advanced Settings
image


Patryk-Malinowski (approved):
lgtm

PR #732, opened 2025-01-02
Show error mods

### LINKED ISSUE #297: Implement Warning/Error filters using the total warnings/errors calculated in the active mod list
There doesn't appear to be a way to only show mods within a list that have warnings or errors, or sort them to the top, for easier viewing. This makes a number of update and pack maintenance tasks much more annoying to perform and has been a feature in Rimpy for longer than I've used it. This is especially necessary as we're waiting for many mods to update to 1.5 and will be for a while.

The best way to implement this without altering the design language would be making the warning number notification on the bottom bar a button that only show mods that have warnings in the active modlist. This same logic should also apply to the error notification. Visual feedback when the cursor hovers over them would also be a nice touch.

**Describe alternatives you've considered**
Idk really, this is a fairly ubiquitous feature and I was surprised it isn't already implimented.

**Additional context**
(If these are already there it makes sense to have them act as filters.)
Screenshot_456

Patryk-Malinowski:
Closes #297 

WIP

### DIFF
--- a/app/controllers/mods_panel_controller.py
+++ b/app/controllers/mods_panel_controller.py
@@ -12,11 +12,12 @@ class ModsPanelController(QObject):
 
         self.mods_panel = view
 
-        self.show_warnings = False
-        self.show_errors = False
+        # Only one label can be active at a time
+        self.warnings_label_active = False
+        self.errors_label_active = False
+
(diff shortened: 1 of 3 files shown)


LionelColaso (approved):
Good for merge

PR #733, opened 2025-01-02
Avoid some redundant error/warning recalculations

Patryk-Malinowski:
Avoids some redundant error/warning recalculations. As far as I can tell there was no need for all these recalculations. I removed some explicit triggers, and added a parameter to avoid recalculations when clearing search. All these changes seemed to have no unexpected affects on RimSorts behavior.

This just utilizes the existing logic of recalculating errors/warnings whenever a mod is inserted/removed from a list. It could be optimized further, might look into that...

Originally when clicking refresh it recalculated each list three times, now it does it once.

Originally when moving a mod from Inactive->Active through double click it recalculated each list twice, now it does it once.

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -437,8 +437,7 @@ class MainContent(QObject):
                 for item in items_to_move:
                     iml.insertItem(count, item)
                     count += 1
-            self.mods_panel.active_mods_list.recalculate_warnings_signal.emit()
-            self.mods_panel.inactive_mods_list.recalculate_warnings_signal.emit()
+        # List error/warnings are automatically recalculated when a mod is inserted/removed from a list
 
     def __handle_inactive_mod_key_press(self, key: str) -> None:
         """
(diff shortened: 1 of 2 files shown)


app/views/main_content_panel.py
```diff
@@ -1007,22 +1005,19 @@ def _do_refresh(self, is_initial: bool = False) -> None:
         # If we are refreshing cache from user action
         if not is_initial:
             # Reset the data source filters to default and clear searches
+            # Avoid recalculating warnings/errors when clearing search
+            # Recalculation for each list will be triggered by mods being reinserted into inactive and active lists automatically
             self.mods_panel.active_mods_filter_data_source_index = len(
                 self.mods_panel.data_source_filter_icons
             )
-            self.mods_panel.signal_clear_search(list_type="Active")
+            self.mods_panel.signal_clear_search(list_type="Active", avoid_recalculate=True)
```
LionelColaso, line 1013: You could use "recalculate_list_errors_warnings=False" instead of "avoid_recalculation=True" 
since you are using "self.recalculate_list_errors_warnings(list_type=list_type)" to calculate errors
this way it would be easier in future i guess

Patryk-Malinowski: Good point considering potential confusion of what would actually be getting recalculated.

LionelColaso (approved):
Good for merge

PR #735, opened 2025-01-03
Move qlistwidgetitem data from dict to a class

### LINKED ISSUE #617: [Suggestion]: Change QListWidgetItem data from dict to a class/dataclass
### Origin of idea

Because of increasing complexity of the data being stored, we could benefit from converting QListWidgetItem data to be a class rather than a dict.

Convert QListWidgetItem data to be a class rather than a dict.

### Alternatives considered

_No response_

### Additional context

This may require you to also create a wrapper for QListWidgetItem and replace all instances of QListWidgetItem in the code to use this wrapper.

The reason for this (for me anyway) was that when `.setData()` was called with a dict it would trigger a signal. (If i remember correctly it would be QListWidget's `itemChanged` signal)

But when trying to call `setData()` with a class it would not trigger this signal. You might want to handle this in your wrapper, or come up with an alternative solution.

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

Patryk-Malinowski:
Closes #617

### DIFF
--- a/app/utils/custom_list_widget_item.py
+++ b/app/utils/custom_list_widget_item.py
@@ -1,6 +1,8 @@
+from PySide6.QtCore import QObject, Signal
 from PySide6.QtWidgets import QListWidgetItem
-from PySide6.QtCore import Qt, Signal, QObject
+
 from app.utils.metadata import MetadataManager
+
 
 class CustomListWidgetItem(QListWidgetItem, QObject):
     """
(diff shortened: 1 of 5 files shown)


LionelColaso:
Have you had a chance to see #403

LionelColaso (approved):
Good to go

PR #738, opened 2025-01-04
Initial fix for Rentry

Patryk-Malinowski:
WIP

### DIFF
--- a/app/utils/rentry/wrapper.py
+++ b/app/utils/rentry/wrapper.py
@@ -1,3 +1,4 @@
+import os
 import re
 import sys
 from json import loads as json_loads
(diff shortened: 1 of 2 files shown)


app/utils/rentry/wrapper.py
```diff
@@ -231,6 +227,7 @@ def import_rentry_link(self) -> None:
                     if match[0] or match[1]
                 ]
                 logger.info("Parsed package_ids successfully.")
+                logger.info(f"Number of package_ids found: {str(len(self.package_ids))}")
```
LionelColaso, line 230: Use logger.debug unless you really want to spam normal log

Patryk-Malinowski: Should it not just print once whenever someone imports?

LionelColaso: Yes i suppose but it's no benefit to normal user except increase in log size

Patryk-Malinowski: Oh yeah valid. Will change it. Thanks!

LionelColaso (approved):
Works

PR #741, opened 2025-01-05
Make sure 'hide invalid mods when filtering' checkbox is checked/unchecked

Patryk-Malinowski:
The hide invalid mods when filtering checkbox would display itself as unchecked after a RimSort restart, although it was still on.

Great catch @LionelColaso


LionelColaso (approved):
👍

PR #754, opened 2025-01-25
Delete mods when unsubscribing on steam

Patryk-Malinowski:
Actually deletes mods when choosing to unsubscribe on steam.

I am unsure whether the name of the action should be changed?
image

I did mention you also delete the mod in the confirmation window.
image


Patryk-Malinowski:
I though this would not delete the existing .vscode folder but just stop tracking it on my end so we could ignore any local changes to it...

app/views/mods_panel.py
```diff
@@ -7,7 +7,7 @@
 from pathlib import Path
 from shutil import copy2, copytree, rmtree
 from traceback import format_exc
-from typing import cast
+from typing import List, Set, cast
```
LionelColaso, line 10: You could avoid importing typings if you replace 
"List" with "list"
"Set" with "set"

LionelColaso:
Seems like You have not configured ruff to run automatically, you can set it with the configs in this repo, if possible enable it to run on file save if you run it manually

Patryk-Malinowski:
> Seems like You have not configured ruff to run automatically, you can set it with the configs in this repo, if possible enable it to run on file save if you run it manually

I have ruff running on every commit I make. But it did not do anything about the Set or List.

app/views/mods_panel.py
Patryk-Malinowski: For some reason I though mypy would give errors if I didnt use these, but it seems happy without them :)

Patryk-Malinowski:
> > Seems like You have not configured ruff to run automatically, you can set it with the configs in this repo, if possible enable it to run on file save if you run it manually
> 
> 
> 
> I have ruff running on every commit I make. But it did not do anything about the Set or List.

Ill change it to run on save

LionelColaso:
> > > Seems like You have not configured ruff to run automatically, you can set it with the configs in this repo, if possible enable it to run on file save if you run it manually
> > 
> > 
> > I have ruff running on every commit I make. But it did not do anything about the Set or List.
> 
> Ill change it to run on save

You can manually run ruff as well using
`ruff format app/`

Settings for reference

image

LionelColaso:
> Move some of the logic into their own methods.
> 
> Actually deletes mods when choosing to unsubscribe on steam.
> 
> I did mention you also delete the mod in the confirmation window. image

Why not ask users or provide 2 conditions 
1) Unsubscribes and deletes mods
2) Unsubscribe and keep mods (this just unsubscribes but keeps the mods)

Patryk-Malinowski:
> > Move some of the logic into their own methods.
> > Actually deletes mods when choosing to unsubscribe on steam.
> > I did mention you also delete the mod in the confirmation window. image
> 
> Why not ask users or provide 2 conditions
> 
>     1. Unsubscribes and deletes mods
> 
>     2. Unsubscribe and keep mods (this just unsubscribes but keeps the mods)

Can do. But I kept it as unsubscribe and delete as thats the default behavior if you unsubscribe on steam.

If converting a mod to local, does it unsubscribe from steam?

LionelColaso:
> If converting a mod to local, does it unsubscribe from steam?

Yes it does not Unsubscribe, this is intentional so that user can have a snapshot copy of the mod and not update the mod even if it updates on steam (user will have 2 copy of the mods, but local mods will be preferred over steam)

Patryk-Malinowski:
> > If converting a mod to local, does it unsubscribe from steam?
> 
> Yes it does not Unsubscribe, this is intentional so that user can have a snapshot copy of the mod and not update the mod even if it updates on steam (user will have 2 copy of the mods, but local mods will be preferred over steam)

I will add an option to unsubscribe and keep mod. But will steam not redownload it... It could cause problems or confusion if it starts rdownloading them while subscribed, whether straight away or at some random point...

LionelColaso:
> > > If converting a mod to local, does it unsubscribe from steam?
> > 
> > 
> > Yes it does not Unsubscribe, this is intentional so that user can have a snapshot copy of the mod and not update the mod even if it updates on steam (user will have 2 copy of the mods, but local mods will be preferred over steam)
> 
> I will add an option to unsubscribe and keep mod. But will steam not redownload it... It could cause problems or confusion if it starts rdownloading them while subscribed, whether straight away or at some random point...

right, i guess its better if you do this instead ?
Unsubscribe and keep mods  (this just unsubscribes but keeps the mods by converting them to local mods)

Patryk-Malinowski:
> > > > If converting a mod to local, does it unsubscribe from steam?
> > > 
> > > 
> > > Yes it does not Unsubscribe, this is intentional so that user can have a snapshot copy of the mod and not update the mod even if it updates on steam (user will have 2 copy of the mods, but local mods will be preferred over steam)
> > 
> > 
> > I will add an option to unsubscribe and keep mod. But will steam not redownload it... It could cause problems or confusion if it starts rdownloading them while subscribed, whether straight away or at some random point...
> 
> right, i guess its better if you do this instead ? Unsubscribe and keep mods (this just unsubscribes but keeps the mods by converting them to local mods)
> 
> but again this is upto you what u think since i dont use steam this was just a thought incase someone wants this option

Yeah makes sense. Even though this is achievable using separate operations, having it as one could save someone the extra clicks so I would be happy to include that option.

Patryk-Malinowski:
Resubscribe is set up currently to delete everything but .dds textures. Do we want to keep this in? Might be a bit confusing for the user when their mod disappears after resubscribing...

We could either force steam validation afterwards, notify the user to run steam validation or not delete anything ourselves.

nyabinary:
So whats the update on this?

Patryk-Malinowski:
> So whats the update on this?

I will be finishing up my existing PRs over the next couple of weeks/months as I find time to do it.

This and improved windows path autodetect will probably be first. Followed by mod coloring and user mod notes.

God-damnit-all:
Consider integrating #1014, which would involve checking for `About/PublishedFileId.txt`, or, as a fallback, if the directory's name is 7 to 11 digits and nothing else, use that.

Not every Workshop mod has an `PublishedFileId.txt` file; I think it's something publishing from within the game automatically adds, so people who publish via alternative means might not have it in their mod. (Perhaps downloading/updating through RimSort should automatically add it?)

PR #768, opened 2025-02-08
Fix indication of mod installation status in Workshop pages

LionelColaso:
#721 was broken this fixes it

image

image


Chunnyluny (approved):
It looks good to me. Go Ahead

PR #777, opened 2025-02-17
Add user mod notes

Patryk-Malinowski:
Allows users to write personal notes.

I will make a separate PR(s) upgrading the notes box to be collapsible (Also able to disable it in settings all together) and possibly support markdown.

image
image


LionelColaso (approved):
Awesome Works as expected

Issue #64, opened 2023-04-30
Implement dynamic "autodetect paths" functionality for all supported platforms


arcayr:
- [x] Implement dynamic path detection for Darwin
- [ ] Implement dynamic path detection for Linux, with complete factoring of all "supported" distribution methods for Steam
    - Ubuntu/Debian
    - Arch
    - Nuances for flatpak distributed Steam
    - Nuances for snap distributed Steam
    - Can we make this support people who choose to run RimWorld from within a wine/proton runner?
    - Any misc. scenarios I may have not mentioned here
- [x] #538
- [x] Ensure consistency in detection method(s) for all platforms
    - Detect Steam install (can be ascertained from from registry on Windows platform)
    - From Steam install, find libraryfolders.vdf as well as `workshop/content`
    - Parse this file to get the Steam library folder(s)
    - Parse Steam library folder(s) for RimWorld game
    - From there, we have paths to RimWorld workshop folder, as well as the RimWorld game itself

Linux path autodetection is Debian-specific

[The Linux entries in the game configuration panel](https://github.com/oceancabbage/RimSort/blob/5dde94befea61af1890e2e256ea4e67b2667f7ef/view/game_configuration_panel.py#LL793C1-L797C10) only check the locations on Debian machines:

        linux_paths = [
            f"{expanduser('~')}/.steam/debian-installation/steamapps/common/RimWorld",
            f"{expanduser('~')}/.config/unity3d/Ludeon Studios/RimWorld by Ludeon Studios/Config",
            f"{expanduser('~')}/.steam/debian-installation/steamapps/workshop/content/294100",
        ]

Most other distributions use the standard `$HOME/.steam/steam/` instead of `$HOME/.steam/debian-installation`.

In my local copy I simply changed the values to those used on my machine. I imagine [this if block](https://github.com/oceancabbage/RimSort/issues/new?permalink=https%3A%2F%2Fgithub.com%2Foceancabbage%2FRimSort%2Fblob%2F5dde94befea61af1890e2e256ea4e67b2667f7ef%2Fview%2Fgame_configuration_panel.py%23LL832C15-L834C19) could be patched to check for both in this case:

        elif system_name == "Linux":
            # Debian keeps some data in 'debian-installer' directory instead of the default.
            if os.path.exists(f"{expanduser('~')}/.steam/debian-installation"):
                os_paths = linux_deb_paths
            else:
                os_paths = linux_paths

[...]

PR #786, opened 2025-02-25
Make windows path autodetect more robust

Patryk-Malinowski:
Makes the windows autodetect for paths more robust. Should be more likely to find the game mods folder now when its on a different drive compared to steam installation etc. Otherwise should use the same fallback paths as previously.

`find_steam_rimworld` should be cross-platform compatible, will do some checking on Linux and might update the other autodetect logic to try using it.

Related #64


app/utils/generic.py
```diff
@@ -5,12 +5,14 @@
...
 import webbrowser
 from errno import EACCES
+from io import TextIOWrapper
 from pathlib import Path
 from re import search, sub
 from stat import S_IRWXG, S_IRWXO, S_IRWXU
 from typing import Any, Callable, Generator
 
 import requests
+import vdf  # type: ignore
```
LionelColaso, line 15: dont you need to add vdf to requirements.txt ?

Patryk-Malinowski: Yes! Thanks for that

app/utils/generic.py
```diff
@@ -384,3 +386,72 @@ def check_valid_http_git_url(url: str) -> bool:
...
+    :return: Rimworld Path if found, blank str otherwise
+    """
+    def __load_data(f: TextIOWrapper) -> str:
+        """
+        Helper function that returns RimWorld path from libraryfolders.vdf
+        if found inside, empty string otherwise.
+        """
+        rimworld_path = ""
+        data = vdf.load(f)
+        library_folders = data["libraryfolders"]
```
hk21702, line 430: While probably not an issue in real operations, I'd consider using .get() every time a dictionary created from a file is involved.

hk21702 (commented):
Generally seems okay and also confirmed that it works. However, I'd like to see more safety when dealing with dictionaries created from reading files.

nyabinary:
Would this look through Heroic Launcher for Linux users?

Patryk-Malinowski:
> Would this look through Heroic Launcher for Linux users?

This PRs scope will be Windows only

LionelColaso:
> > Would this look through Heroic Launcher for Linux users?
> 
> This PRs scope will be Windows only

Is this finished or pending ?

Patryk-Malinowski:
> Is this finished or pending ?

Pending. But should be quick to finish when I get around to it.

LionelColaso (approved):
Merge when you think it's ready

PR #787, opened 2025-02-25
game_folder:  validate game folder using "Version.txt"

LionelColaso:
Simple validation only checks for Version.txt in the game_folder if not found does not let user set the game_folder


app/controllers/settings_controller.py
```diff
@@ -921,15 +923,35 @@ def _on_global_ok_button_clicked(self) -> None:
     @Slot()
     def _on_game_location_text_changed(self) -> None:
         game_folder = self.settings_dialog.game_location.text()
-        self.settings_dialog.game_location_open_button.setEnabled(game_folder != "")
-        # Automatically set local folder from game folder
-        if game_folder:
+        version_file = str(game_folder / Path("Version.txt"))
+        if game_folder != "" and os.path.exists(version_file):
```
twstagg, line 927: Replacing the condition here is not good i don't think

`if game_folder` also accounts for None, etc. If you are now explicitly checking the string, it might not be as reliable as before is my thought process there?

app/controllers/settings_controller.py
```diff
@@ -921,15 +923,35 @@ def _on_global_ok_button_clicked(self) -> None:
...
+            self.settings_dialog.game_location_open_button.setEnabled(True)
```
twstagg, line 928: Well now I see why you did the first thing. Perhaps we just combine both conditions into the single statement instead?

twstagg (commented):
All in all, I like what you are trying to do - making this reliable.

Might I add a suggestion to this:

- Before checking the file, check for the existence of the platform-specific executable files needed to actually execute the game.
- Use the fact from this as the first piece of the validation, in conjunction with you checking the Version.txt file

I'm also not certain we should be using Version.txt as this validation. The game can run & function without it, as far as I know.

app/controllers/settings_controller.py
LionelColaso: Should be good now

PR #790, opened 2025-02-26
Accessibility settings - Global font setting

### LINKED ISSUE #57: Implement setting to change font scale

Patryk-Malinowski:
Closes #57

### DIFF
--- a/app/controllers/app_controller.py
+++ b/app/controllers/app_controller.py
@@ -2,8 +2,10 @@ import json
 import sys
 
 from PySide6.QtCore import QObject
+from PySide6.QtGui import QFont
 from PySide6.QtWidgets import QApplication
 
+import app.utils.rimsort_boot_config as rimsort_boot_config
 from app.controllers.main_window_controller import MainWindowController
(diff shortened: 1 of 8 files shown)


LionelColaso:
Some Suggestions 
1) instead of creating a new tab, reuse the theme tab and rename it if you want
2) check theme_controller for how you can apply the font without restarting the app
3) set limit to the minimum and maximum font size a user can set (setting unreasonable size will lead to unusable UI)

Patryk-Malinowski:
>     1. instead of creating a new tab, reuse the theme tab and rename it if you want
That would make the most sense. Will do this.
>     2. check theme_controller for how you can apply the font without restarting the app
Will look into it, thanks!
>     3. set limit to the minimum and maximum font size a user can set (setting unreasonable size will lead to unusable UI)
This is actually already in the code :) Although I might change the actual limits.

LionelColaso:
Implemented in https://github.com/RimSort/RimSort/pull/859

PR #799, opened 2025-03-11
Add files needed for flatpak of the app

yodatak:
Its a follow up from here 
There is some issue on the license of the app because of sdk presence but i dont know how to fix see https://github.com/flathub/flathub/pull/6197#pullrequestreview-2637565316

https://github.com/flathub/flathub/pull/6197

some more task to follow
https://github.com/flathub/flathub/pull/6197#issuecomment-2692322847

If somebody want to help your welcome


twstagg (approved):
Thanks. This metadata looks fine to me at face value.

PR #801, opened 2025-03-14
Fix re_steamcmd_action

LionelColaso:
- [x] Added Debugging for re_steamcmd_action
- [x] Added Missing Signal for re_steamcmd_action


twstagg (approved):
Good looks!

PR #880, opened 2025-05-23
Fix: Revert unintended text changes

Ocean456:
fix: update tooltip text for mod loading errors to English


LionelColaso (approved):
good its just a Tooltip

PR #885, opened 2025-05-28
Fix window geometry inconsistency by saving calculated defaults

Ocean456:
The changes ensure that calculated window geometry is saved persistently.


LionelColaso (approved):
Good for Merge

Issue #435, opened 2024-06-21
FIXME: "Run" button bypasses Steam


Limfjorden:
**Describe the bug:**
When I click the run button, the game launches without steam. This means workshop mods don't get loaded.

**To Reproduce:**

1. Launch RimSort (make sure Steam is also launched and steam client integration is set to on in RimSort)
2. Press "Run" in RimSort
3. See error; game has been launched without Steam

**Expected behavior:**
The run button should run the game with Steam

**Please complete the following information:**

- OS: macOS Ventura 13.6.6
- RimSort version: v1.0.9-edge14+f58043f
- Please provide a [RimSort.log](https://github.com/user-attachments/files/15929296/rimsortlog.txt)
- I am running from a compiled RimSort release

**Did you first check the wiki for a solution? (macOS Gatekeeper workaround is there)**
Yes.

hk21702:
Think I've seen this reported another time on Mac. Haven't seen it reported for other platforms. Did you make sure that your steam was running and logged in? Also, does the game function normally if launched from steam? (with the correct mod list)

Also, are you on arm or X86? Did you use the correct corresponding build?

Limfjorden:
@hk21702 I use intel and I did use the right build. Might be an idea to add a "processor" category to the bug template here on github.

hk21702:
Based on other reports, it appears that this is specific to mac, but processor agnostic. It's difficult for me to look into it though because I don't have a mac.

adampizzo:
I have an M2 mac and see the same behavior. I also confirmed that once the mod list is saved, it launches fine from Steam Directly, but not from clicking run in RimSort.

Honestly, just installed it a few minutes ago. I had to run the xattr command for RimSort.app. I couldn't find the libstream_api.dylib, so I was wondering if that had anything to do with it. Once I found out where all of my proper folder paths are (it was not able to find everything by default), all the mods pull up working and fine from the look of things.

Happy to help test anything (Maybe you can point me to where that libstream_api.dylib is 😅)

Limfjorden:
@hk21702 Hi, what can I do to assist you with this issue? Do you need me to test different solutions (I have built it myself before). I would really like to have this issue fixed, as without it, instancing is broken on mac.

Limfjorden:
> (...) I had to run the xattr command for RimSort.app. I couldn't find the libstream_api.dylib, so I was wondering if that had anything to do with it. (...) (Maybe you can point me to where that libstream_api.dylib is 😅)

[...]

PR #930, opened 2025-06-18
Fixed steam_appid.txt being in the wrong location on macOS

larsy1995:
Fixed steam_appid.txt being in the wrong location on macOS by making ternary statement that moves the steam_appid.txt outside the game directory or app bundle on macOS. 
With the file adjacent to the app bundle the steam integration also works with the new Apple Games application.

Also changed more of the code to use pathlib. 

This fixes issues like: 
https://github.com/RimSort/RimSort/issues/435

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -3229,7 +3229,7 @@ class MainContent(QObject):
             ) as f:
                 f.write("294100")
         elif not steam_client_integration and steam_appid_path.exists():
-            os.remove(steam_appid_path)
+            steam_appid_path.unlink()
 
         # Launch independent game process without Steamworks API
         launch_game_process(game_install_path=game_install_path, args=run_args)


LionelColaso:
i had this solution in mind which was way already being tested
image

larsy1995:
> i had this solution in mind which was way already being tested image

I mean, I don't think the solution is to remove the .app from bundle file path, but I have not tested that myself. That would still place the steam_appid.txt inside the games root folder though. The older version still created and removed that file just fine, the integration just didn't work, which it does with my solution. But again, haven't tested yours.

LionelColaso:
> I mean, I don't think the solution is to remove the .app from bundle file path, but I have not tested that myself. That would still place the steam_appid.txt inside the games root folder though. The older version still created and removed that file just fine, the integration just didn't work, which it does with my solution. But again, haven't tested yours.

Since i dont have access to Mac, My solution was based on what i could gather from screenshot below
image
image
image

larsy1995:
> This would be better solution ? #933

I tried that one, but it didn't do anything for the steam integration part.  
I don't really see how removing the .app part of the bundle would do anything, especially since steam opens the bundle from the RimWorld folder in which the steam_appid.txt needs to reside. This PR ensures the latter, that the txt is created in the correct location and not inside the "root" of the game folders.

LionelColaso:
> > This would be better solution ? #933
> 
> I tried that one, but it didn't do anything for the steam integration part. I don't really see how removing the .app part of the bundle would do anything, especially since steam opens the bundle from the RimWorld folder in which the steam_appid.txt needs to reside. This PR ensures the latter, that the txt is created in the correct location and not inside the "root" of the game folders.

ok i am approving the request go ahed and merge

larsy1995:
> > > This would be better solution ? #933
> > 
> > 
> > I tried that one, but it didn't do anything for the steam integration part. I don't really see how removing the .app part of the bundle would do anything, especially since steam opens the bundle from the RimWorld folder in which the steam_appid.txt needs to reside. This PR ensures the latter, that the txt is created in the correct location and not inside the "root" of the game folders.
> 
> ok i am approving the request go ahed and merge

I don't seem to have the permission to merge?

PR #943, opened 2025-06-20
Add option to update databases on startup

LionelColaso:
Introduce a setting that allows users to automatically update databases from GitHub when the application starts. This feature includes a checkbox in the settings dialog and triggers updates for various configured databases if enabled.


Ocean456 (commented):
looks good

PR #948, opened 2025-06-21
Missing dds support mod warning dialog

### LINKED ISSUE #796: Add an button to install the mod for loading optimized images
### Origin of idea

On rimpy it had a text informing that there was needed to use a mod to load those optimized images.

Create a button or warn of the use of the mod to load the optimized images, or else it would not make effect.(it doesn't make effect without the mod right?)

### Alternatives considered

_No response_

### Additional context

Not sure if the new optimizer tool make changes without needing a mod to load, I assume the mod still is required.

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

xDragoxboi:
Adds a dialog to warn user if active mod list does not contain Graphic settings+ or Performance Fish.

well_shit

Closes #796

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -2047,8 +2047,42 @@ class MainContent(QObject):
                 "Cached mod lists for restore function not set as client started improperly. Passing on restore"
             )
 
+    
+    def _check_and_warn_if_dds_mods_missing(self) -> bool:
+        """
+        Checks if crucial DDS support mods are active and warns if not.
+        Returns True if a warning was shown.
+        """
+        REQUIRED_PACKAGE_IDS = {"telefonmast.graphicssettings", "bs.performance"} # I could only find these two so far
+
+        try:
+            for uuid in self.mods_panel.active_mods_list.uuids:
+                package_id = self.metadata_manager.internal_local_metadata[uuid]["packageid"]
+                if package_id.lower() in REQUIRED_PACKAGE_IDS:
+                    return False
+                
+        except (AttributeError, KeyError) as e:
+            logger.error(f"Error accessing metadata during DDS mod check (short version): {e}. Warning user anyway.")
+            
+        # If loop completes without returning False.
+        dialogue.show_warning(
+            title=self.tr("DDS Support Mod Missing"),
+            text=self.tr(
+                "To load .dds textures, the game requires Graphic Settings+ or Performance fish."
+            ),
+            information=self.tr(
+                "The 'Optimize Textures' feature will convert images to .dds files. However, without one of the aforementioned mods active, the game will not be able to load these optimized .dds textures, and you may not see any performance improvement or visual change. Please ensure at least one of these mods is active."
+            ),
+        )
+        return True 
+    
+
     # TODDS ACTIONS
     def _do_optimize_textures(self, todds_txt_path: str) -> None:
+
+        # Check for dds support mods and warn if missing.
+        self._check_and_warn_if_dds_mods_missing()
+
         # Setup environment
         todds_interface = ToddsInterface(
             preset=self.settings_controller.settings.todds_preset,


xDragoxboi:
Tested

With Graphic settings+ Active
With Performance Fish Active
With Both Active
With Both Inactive

LionelColaso:
Use [constants](https://github.com/RimSort/RimSort/blob/main/app/utils/constants.py) to store `REQUIRED_PACKAGE_IDS` and rename it to `KNOWN_MODS_DDS_SUPPORT ` or whatever 
add a conditional dialogue to download the missing mod install the mod as stated in #796

app/views/main_content_panel.py
```diff
@@ -2047,8 +2047,42 @@ def _do_restore(self) -> None:
                 "Cached mod lists for restore function not set as client started improperly. Passing on restore"
             )
 
+    
+    def _check_and_warn_if_dds_mods_missing(self) -> bool:
+        """
+        Checks if crucial DDS support mods are active and warns if not.
+        Returns True if a warning was shown.
+        """
+        REQUIRED_PACKAGE_IDS = {"telefonmast.graphicssettings", "bs.performance"} # I could only find these two so far
```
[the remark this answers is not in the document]
xDragoxboi: > You will have to import `KNOWN_DDS_SUPPORT_MODS` `REQUIRED_PACKAGE_IDS = KNOWN_DDS_SUPPORT_MODS`
> 
> here is an example https://github.com/RimSort/RimSort/blob/main/app/views/mods_panel.py#L41 https://github.com/RimSort/RimSort/blob/main/app/views/mods_panel.py#L1994


Yep! 👍, 
I commit the main content panel updates last night and went to bed but vscode decided it doesn't wanna do that so, it was a successful commit on the editor but didn't actually push commit to GitHub for some reason.


I'll push the changes from GitHub manually soon as I can.

app/views/main_content_panel.py
```diff
@@ -2047,8 +2047,46 @@ def _do_restore(self) -> None:
...
+            )
+        
+        if choice == "Download":
+            steamcmd_wrapper = (
+                    self.steamcmd_wrapper                    
+                    )
+            if not steamcmd_wrapper.setup:
+                self._do_setup_steamcmd()
+            if steamcmd_wrapper.setup:
+                        self._do_download_mods_with_steamcmd(publishedfileids=["1678847247"]) # Downloading Graphic settings+
```
LionelColaso, line 2082: Some people don't Use Graphic settings+ including me, so it's better you give user the choice which mod they want to download

app/views/main_content_panel.py
```diff
@@ -2047,8 +2047,46 @@ def _do_restore(self) -> None:
...
+        # Loop finished meaning no dds mods found
+        # show option to download
+        choice = dialogue.show_dialogue_conditional(
+            title="Missing Graphic settings+",
+            text="You are missing the Graphic settings+ mod. Do you want to download it?",
+            information="The 'Optimize Textures' feature will convert images to .dds files. However, without one of the aforementioned mods active, the game will not be able to load these optimized .dds textures, and you may not see any performance improvement or visual change. Please ensure at least one of these mods is active.",
+            button_text_override=["Download"]
+            )
+        
+        if choice == "Download":
```
LionelColaso, line 2075: Why Only steamcmd option ? you have to have both steam and steamcmd and let user decide 
or use the base_mods_panel or missing_mods_panel to trigger the download that way user can decide which mod they want to download

app/views/main_content_panel.py
xDragoxboi: Gotcha, was already thinking of adding separate buttons for each.

LionelColaso: > Gotcha, was already thinking of adding separate buttons for each.

An example in case https://github.com/RimSort/RimSort/blob/main/app/views/main_content_panel.py#L1490

xDragoxboi:
oopsies

xDragoxboi:
s2
suc
s1

LionelColaso:
> s2 suc s1

Problem
Your Current Implementation only checks if the mods are in "Active list" and gives the download prompt even if the mods are already downloaded and in "Inactive list"

Solution
when mod is already downloaded, you should offer user to add it to active list instead of showing the download prompt

Suggestion
Convert the prompt into a menu bar button and add it in the menu if you just want users to download the mod which best suits how your current implementation is

LionelColaso:
You can check https://github.com/RimSort/RimSort/blob/main/app/controllers/main_window_controller.py#L67 for how missing dependencies are added and do something similar or reuse the existing logic

Continous:
It may be better to rephrase the warning to say "No DDS-supporting mod was detected, below are some known mods which enable support." Just in case some ephemeral future mod comes along with support. Additionally, an option to never receive the warning dialogue again.

katanacrimson:
For awareness, DDS texture files are now natively supported by recent builds of Rimworld 1.6; mods are no longer necessary to use these files for textures.

PR #951, opened 2025-06-23
Refactor Git config initialization and add internet connection checks

LionelColaso:
Improve code maintainability by centralizing GitOperationConfig initialization. Introduce internet connection checks before network operations, enhancing user feedback and preventing errors during offline scenarios. A dialog now alerts users when there is no internet connection.

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -1557,6 +1557,10 @@ class MainContent(QObject):
             self.__missing_mods_prompt()
 
     def _do_import_list_workshop_collection(self) -> None:
+        # Check internet connection before attempting task
+        if not check_internet_connection():
+            dialogue.show_internet_connection_error()
+            return
         # Create an instance of collection_import
         # This also triggers the import dialogue and gets result
         collection_import = CollectionImport(metadata_manager=self.metadata_manager)


app/utils/git_utils.py
```diff
@@ -96,6 +100,72 @@ def create_with_handler(
...
 
+    @classmethod
+    def create_with_timeout(
+        cls, fetch_timeout: int = 30, connection_timeout: int = 10
+    ) -> "GitOperationConfig":
+        """Create a config with custom timeout values."""
+        return cls(fetch_timeout=fetch_timeout, connection_timeout=connection_timeout)
+
+
+def _check_network_connectivity(
```
LionelColaso, line 111: Please reuse the existing `check_internet_connection` in app/utils/generic.py

app/utils/git_utils.py
```diff
@@ -96,6 +100,72 @@ def create_with_handler(
         """Create a config with a specific notification handler."""
         return cls(notify_errors=True, notification_handler=handler)
 
+    @classmethod
+    def create_with_timeout(
```
LionelColaso, line 104: it would be better if `create_with_timeout `was in a way that it could be used by other operations as well and not just git_utils.py
for example how  `check_internet_connection `method is implemented so that it can be used by any method if required

PR #967, opened 2025-06-30
Refactor default settings for database and mod warnings

LionelColaso:
Adjust default settings to disable database expiry, clarify notification messages, and enable warnings for duplicate mods and missing downloads.


Dragonissa:
Wouldn't it make sense to change the default settings to download the git database repositories? I fail to see in the user guide that you actually have to enable this and that is for many others the case.

PR #995, opened 2025-07-10
Add environment variable RIMSORT_DISABLE_UPDATER to disable update checking

weirdrock:
For the in progress Nix package, RimSort is run from "source", not nuitka. This adds an easy way to hide the "skipping update check" dialog that runs every time the app is refreshed.

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -746,6 +746,10 @@ class MainContent(QObject):
         4. Downloads and extracts the update if user confirms
         5. Launches the appropriate update script for the platform
         """
+        if os.getenv("RIMSORT_DISABLE_UPDATER"):
+            logger.debug("RIMSORT_DISABLE_UPDATER is set, skipping update check silently.")
+            return
+        
         logger.debug("Checking for RimSort update...")
 
         # NOT NUITKA


app/views/main_content_panel.py
```diff
@@ -746,6 +746,10 @@ def _do_check_for_update(self) -> None:
         4. Downloads and extracts the update if user confirms
         5. Launches the appropriate update script for the platform
         """
+        if os.getenv("RIMSORT_DISABLE_UPDATER"):
+            logger.debug("RIMSORT_DISABLE_UPDATER is set, skipping update check silently.")
+            return
+        
         logger.debug("Checking for RimSort update...")
 
         # NOT NUITKA
```
LionelColaso, line 755: Does This not skip the update since its not built with Nuitka ?

weirdrock: It does skip the update already, but it produces a dialog the user has to dismiss.

I assumed the dialog exists to tell the user that the application is running in a "abnormal" state where it cannot be updated. 

In the Nix package, however, using Nix to update the app is the intended way to update it, so the update checker should not be running, and there should be no warning.

If this could be handled through a config option instead, that would work too. I think it is a bad experience for Nix users to have to constantly close the "Skipping update check" dialog, with no way to disable it.

LionelColaso (approved):
Good For Merge

PR #1003, opened 2025-07-12
Quote current directory during update on Windows

jgoodson:
Sometimes, when you install Windows 11 and use a Microsoft account it decides that your username and therefore your home directory in C:\Users\ should be your full name with a space in it. The update script quotes the source directory path but not the destination, so the robocopy command fails if you have a space anywhere in the latter path.

Tested with an update from 1.0.23 to 1.0.24 on Windows 11.


LionelColaso (approved):
Looks Good

Issue #70, opened 2023-05-08
Loading placeholder keep flashing and webbrowser keep moving because of the loading bar


cylian914:
The place holder for loading the website keep flashing, can it be just remove or only load when the webbrowser start ?
https://user-images.githubusercontent.com/73445993/236800763-77540eb2-a20d-4bff-9296-f5c972d1cf40.mp4

Also the loading bar keep disappearing, the bar should always  be there even if it's invisible

twstagg:
Per discussion, i'm open to replacing this image. We'll keep it as-is for now. I personally think it looks a bit better than the white background the webview has during that stage of loading the page.

hk21702:
It might be better to just do a solid shade that is similar to that of the background of steam? Then you get less sudden bright lights.

EttyKitty:
> Also the loading bar keep disappearing, the bar should always be there even if it's invisible

I can offer reduced, fixed 8px height; the bar is always displayed; the % is only displayed when loading; the progress color is changed to fit steam colors; the font size is decreased.
The text doesn't look very good, but how much do you care. Can remove it completely.

<details>
<summary>Screenshots</summary>

<img width="1239" height="85" alt="Image" src="https://github.com/user-attachments/assets/02a8c7cc-0612-413d-b843-6540200264c0" />

<img width="1707" height="65" alt="Image" src="https://github.com/user-attachments/assets/eaabd2fa-1383-47c0-b848-6db93ab7a7d5" />

</details>

> The place holder for loading the website keep flashing, can it be just remove or only load when the webbrowser start ?

Here, just a solid color looks okay imo.

<details>
<summary>Screenshot</summary>

<img width="1709" height="780" alt="Image" src="https://github.com/user-attachments/assets/cd042460-d273-4232-a91b-9c7ff4473bbe" />

</details>

You can simply remove `setPixmap` and replace `ImageLabel` with `QLabel`, and it'll just use the current theme color. I think there are no caveats?

EttyKitty:
Can this issue count as solved now that #1020 was merged?

https://github.com/user-attachments/assets/c64b46e6-fdf4-4812-8ca6-203e11535f72

LionelColaso:
> Can this issue count as solved now that [#1020](https://github.com/RimSort/RimSort/pull/1020) was merged?
> 2025-07-15.08-49-33.mp4

Yes Thankyou I thought the issue was tagged and closed along with merge
Closing Now

PR #1007, opened 2025-07-13
Refactor TroubleshootingDialog UI for Enhanced Design and Usability

Ocean456:
This pull request refactors the `TroubleshootingDialog` to modernize the UI, improve user experience, and enhance maintainability.

image

PR #1020, opened 2025-07-15
feat(browser): Stop placeholder flashing; Edit the progress bar

EttyKitty:
Hopefully, addresses things from #70.

### What's done
- Make the progress bar persistent, so that the page stops changing size when it disappears.
- Remove the progress bar border, so that it doesn't look ugly always sitting on the screen.
- Reduce the progress bar height to 8 pixels, so that it takes less space.
- The % text only visible when loading.
- Smaller % font.
- Progress bar color now properly changes with the theme.
- The placeholder only showed once, when the browser is opened for the first time.

### How tested
- Spent some time in the browser.
- Formatted/fixed with ruff.

### DIFF
--- a/app/utils/steam/browser.py
+++ b/app/utils/steam/browser.py
@@ -158,17 +158,11 @@ class SteamBrowser(QWidget):
         )
         # self.nav_bar.addSeparator()
         self.progress_bar = QProgressBar()
-        self.progress_bar.setObjectName("default")
+        self.progress_bar.setObjectName("browser")
         self.progress_bar.setRange(0, 100)
         self.progress_bar.setFixedHeight(8)
         self.progress_bar.setTextVisible(False)
         self.progress_bar.setVisible(True)
-        self.progress_bar.setStyleSheet("""
-            QProgressBar {
-                font-size: 10px;
-                border: 0px;
-            }
-        """)
 
         # Build the downloader layout
         self.downloader_layout.addWidget(self.downloader_label)
--- a/app/utils/steam/browser.py
+++ b/app/utils/steam/browser.py
@@ -6,7 +6,7 @@ from typing import Any
 
 from loguru import logger
 from PySide6.QtCore import QPoint, Qt, QUrl, Signal
-from PySide6.QtGui import QAction, QPixmap
+from PySide6.QtGui import QAction
 from PySide6.QtWebEngineCore import QWebEnginePage
 from PySide6.QtWebEngineWidgets import QWebEngineView
 from PySide6.QtWidgets import (
@@ -24,8 +24,6 @@ from PySide6.QtWidgets import (
     QWidget,
 )
 
-from app.models.image_label import ImageLabel
-from app.utils.app_info import AppInfo
 from app.utils.generic import extract_page_title_steam_browser
 from app.utils.metadata import MetadataManager
 from app.utils.steam.webapi.wrapper import (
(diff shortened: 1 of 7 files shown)


app/utils/steam/browser.py
```diff
@@ -114,16 +112,11 @@ def __init__(self, startpage: str, metadata_manager: MetadataManager):
 
         # BROWSER WIDGETS
         # "Loading..." placeholder
-        self.web_view_loading_placeholder = ImageLabel()
+        self.web_view_loading_placeholder = QLabel()
         self.web_view_loading_placeholder.setAlignment(Qt.AlignmentFlag.AlignCenter)
         self.web_view_loading_placeholder.setSizePolicy(
             QSizePolicy.Policy.Expanding, QSizePolicy.Policy.Expanding
         )
-        self.web_view_loading_placeholder.setPixmap(
```
LionelColaso, line 122: Could you keep he image ?

app/utils/steam/browser.py
```diff
@@ -157,8 +150,19 @@ def __init__(self, startpage: str, metadata_manager: MetadataManager):
         )
         # self.nav_bar.addSeparator()
         self.progress_bar = QProgressBar()
-        self.progress_bar.setMinimum(0)
-        self.progress_bar.setMaximum(100)
+        self.progress_bar.setRange(0, 100)
+        self.progress_bar.setFixedHeight(8)
+        self.progress_bar.setTextVisible(False)
+        self.progress_bar.setVisible(True)
+        self.progress_bar.setStyleSheet("""
```
LionelColaso, line 157: Stylesheets are in themes folder, it would be better I think if it was there, unless you want to use the same colors for all themes

app/utils/steam/browser.py
```diff
@@ -157,8 +150,19 @@ def __init__(self, startpage: str, metadata_manager: MetadataManager):
         )
         # self.nav_bar.addSeparator()
         self.progress_bar = QProgressBar()
-        self.progress_bar.setMinimum(0)
-        self.progress_bar.setMaximum(100)
+        self.progress_bar.setRange(0, 100)
+        self.progress_bar.setFixedHeight(8)
```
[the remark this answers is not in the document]
LionelColaso: I am ok with removing the progress bar, but I would like to keep the background image
I, too, think the progress bar is not needed since the reload /refresh button already shows the page is fully loaded or not
But then there are users who have slow internet or issues such as in screenshot since progress bar at least shows that the internet is too slow or page is loading too slow 
the page took more than 4–5 mins to load
image

app/utils/steam/browser.py
EttyKitty: I guess I misunderstood. I thought the problem was the image flashing, as the user used a dark theme and that was the only "flashing" thing.
Is the point of the issue is to somehow bypass the need for the placeholder? Like modern browsers do, with keeping the old page up until the new one is semi-loaded? I'm not even sure how to do it and why it's not happening. Hm…

EttyKitty: > I'm not even sure how to do it and why it's not happening. Hm…

Wait, I think know why.

LionelColaso:
Just a quick question is better to load the browser current is `self.showMaximised()` on line 197 
`self.showMaximised() `or `self.showNormal()`

EttyKitty:
> self.showMaximised() or self.showNormal()

Not sure. I personally like my stuff maximized, usually. If it was up to me, and if it's a binary choice - I would like all possible RimSort windows maximized at the start.
But ideally, you probably want them to retain their last state (even after relaunching the program). Or at least have a global toggleable user setting "maximize all windows", maybe?

LionelColaso:
> > self.showMaximised() or self.showNormal()
> 
> Not sure. I personally like my stuff maximized, usually. If it was up to me, and if it's a binary choice - I would like all possible RimSort windows maximized at the start. But ideally, you probably want them to retain their last state (even after relaunching the program). Or at least have a global toggleable user setting "maximize all windows", maybe?

Yes, I will just add a Toggle, so all Individual windows can be configured in settings
Thanks

LionelColaso:
Also if you can shed some ideas or light here https://github.com/RimSort/RimSort/pull/1007

EttyKitty:
p.s. It's probably a good idea to squash this PR.

LionelColaso (approved):
Good For Merge @EttyKitty Thanks Again

PR #1022, opened 2025-07-15
fix(browser): "Add to list" button getting stuck

EttyKitty:
### The issue
"Add to list" button is not deleted after the user opens a page of any mod and then goes back to the browse mods screen.
Caused by me forgetting that I've changed the `if` block above `else` that contains button deletion, causing it to not get executed if page = browse mods.
### The fix
Now the button is deleted on every `_web_view_load_started` call. I think there is no reason not to.
### The tests
Tested that the original problem is solved, and that apparently no problems with individual mod or collection "Add to list" buttons popped out - they work fine.


LionelColaso (approved):
Good for merge

PR #1031, opened 2025-07-17
feat(browser): Quick add/remove buttons on tiles

EttyKitty:
### Description
- Add badges/buttons to quickly add/remove mods from the list, to mod tiles in the workshop browser.
  - Supports mod tiles on collection pages as well.
- Refactor some bits of the code related to this.
### More Detailed Description
- Badges:
  - Hidden by default for not-installed and not-added mods, unless you hover the mod, to not clutter the UI.
  - Have a minor effect when you click on them, to add responsiveness.
  - Have a hover-over effect.
- JS bridge stuff:  
  - Use `QWebChannel` to have a bridge between JS and python.
  - No idea if this is the best way, I just found it to work, and it was the first thing I found on the internet.
  - Had to manually inject `qwebchannel.js` with `QWebEngineScript`, as I have no idea how to get it to work otherwise.
- JS script stuff:
  - Move the main badge/button JS script to a separate JS file, to have proper intellisense and less bloat in the `SteamBrowser` class.
  - Use `Template()`>`substitute()`>`json.dumps()` to replace template vars in JS with vars from python, to avoid the need to inject said variables directly into the page.
### TODO
- Nail the file structure, as I've just placed everything in the same folder, which may be not ideal.
- Test this on other systems perhaps, as I'm not sure if there could be problems with the bridge and stuff.
- Generally do more edge-case testing as well.
- ~Add a hover-over effect to badges, because all cool kids should have that.~
- In some other PR, split the huge `_web_view_load_finished()` method into smaller parts.
- In some other PRx2, maybe move some other big JS function to separate files.
### Preview
<details>
<summary>Spoiler</summary>

https://github.com/user-attachments/assets/3d0d5539-22de-4f45-bc6f-026705aabf13

</details>

### DIFF
--- a/app/utils/steam/browser.py
+++ b/app/utils/steam/browser.py
@@ -1,13 +1,18 @@
+import json
 import os
 import platform
 import re
+from enum import Enum
 from functools import partial
+from pathlib import Path
+from string import Template
 from typing import Any
 
 from loguru import logger
 from PySide6.QtCore import QPoint, Qt, QUrl, Signal
 from PySide6.QtGui import QAction, QPixmap
-from PySide6.QtWebEngineCore import QWebEnginePage
+from PySide6.QtWebChannel import QWebChannel
+from PySide6.QtWebEngineCore import QWebEnginePage, QWebEngineScript
 from PySide6.QtWebEngineWidgets import QWebEngineView
 from PySide6.QtWidgets import (
     QHBoxLayout,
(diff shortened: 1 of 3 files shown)


EttyKitty:
Hover-over effect was added. Now I just need to test this more and perhaps get some tips re the file/folder structure, if any.
Then it should be ready for review.

LionelColaso (approved):
Awesome

PR #1050, opened 2025-07-20
Centralize pygit2 imports and SSL handling

Ocean456:
Centralized pygit2 import logic with SSL handling in `git_utils.py`.
Removed duplicate try-catch blocks.


LionelColaso (approved):
Nice

PR #1052, opened 2025-07-20
Add auto-translation and one-click workflow support

Ocean456:
- Introduces auto-translate command that fills unfinished strings via Google, DeepL or OpenAI.
- Adds a new process command: update .ts → auto-translate → compile .qm in one step

PR #1060, opened 2025-07-21
Add Danish translation

Limfjorden:
This pull request will add a Danish translation for all strings in RimSort.


axoroll7:
Why the file `greenlet.h` ?

Limfjorden:
I don't know. It was automatically created at some point. If it's not needed I can remove it.

LionelColaso:
> I don't know. It was automatically created at some point. If it's not needed I can remove it.

https://rimsort.github.io/RimSort/development-guide/translation-guidelines

also worth checking 
https://github.com/RimSort/RimSort/pull/1052

oceanfrost:
Some tips:
`greenlet.h` is not needed.
The current build workflow doesn't include compiling the language files, so you need to upload the .qm files as well.
Add the corresponding mapping in the language controller.

Limfjorden:
> Why the file `greenlet.h` ?

Greenlet.h is gone now.

> The current build workflow doesn't include compiling the language files, so you need to upload the .qm files as well.

Yeah, I'll do that when I've translated all the strings.

> Add the corresponding mapping in the language controller.

Good call, I'll fix it.

Issue #1066, opened 2025-07-22
add to winget


ericogamo:
### Suggestion/proposed solution?

Hi, it would be really nice if you could add Rimsort to Winget.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

jonesmz:
Winget has RimSort as an installable package now.

However the offered package is version 1.0.30.

And the built-in auto-updater doesn't appear to work with a winget installed RimSort.

And there's also no start menu short cut installed by the winget installer.

jonesmz:
Seems like the pull request that added the winget package used the zip file, instead of the msi.

It might be worth providing an updated winget package description that favors the MSI instead.

https://github.com/microsoft/winget-pkgs/pull/276902

LionelColaso:
@jonesmz 
Thanks for letting me know
I have not created that, but I will see what I can do

PR #1079, opened 2025-07-28
Add MSI package creation and upload steps for Windows builds

oceanfrost:
This pull request adds support for generating MSI installer packages for Windows builds using the WiX Toolset.
This change ensures RimSort Windows users can install the app via a standard MSI package, enhancing support for user-friendly installation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -179,11 +179,148 @@ jobs:
           echo "Executable found at $executable"
         shell: bash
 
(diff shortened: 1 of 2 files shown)


LionelColaso:
If we are going to do .msi builds then you should also consider https://github.com/RimSort/RimSort/issues/1066

LionelColaso (approved):
Looks Good

PR #1099, opened 2025-07-30
Detect and set the application language

LionelColaso:
Detects and sets then application language If OS language is detected and language exists in the app, Sets the app language to the detected OS language


oceanfrost:
would using Qt locale be better

LionelColaso:
> would using Qt locale be better
Maybe not sure but current implementation is less intrusive and gives opportunity to reload the app if we want to do it

oceanfrost:
`app/utils/steam/steambrowser/browser.py`
QWebEngineProfile.defaultProfile().setHttpAcceptLanguage()
can you consider this? modify the language of the built-in browser as well

LionelColaso:
> `app/utils/steam/steambrowser/browser.py` QWebEngineProfile.defaultProfile().setHttpAcceptLanguage() can you consider this? modify the language of the built-in browser as well

yes sure will in next pr

PR #1103, opened 2025-07-30
Feature: Unsaved Changes Prompt on Game Launch

### LINKED ISSUE #86: QOL Feature Suggestion
Notify user if user tries to close RimSort if save button is blinking when closing RimSort or user tries to close RimSort without saving

Artemonim:
This PR improves the user experience by preventing accidental loss of mod list changes.

close #86 

---

### Description

Previously, if a user modified their active mods and clicked "Run Game" without saving, the game would launch with the last-saved configuration, and the recent changes would be ignored.

Now, before launching the game, the application checks if the active mod list has been modified. If it has, a dialog appears, prompting the user with the following options:
- **Save and Run**: Saves the current mod list and then launches the game.
- **Run Anyway**: Launches the game with the previously saved mod list, ignoring the current changes.
- **Cancel**: Aborts the launch and returns the user to the application.

### Changes

- Added the core logic to `main_content_panel.py` to check for unsaved changes and display a dialog.
- Added new unit tests in `test_main_content_run.py` to cover all scenarios (Save, Run Anyway, Cancel, and no changes).
- Added internationalization (i18n) strings for the new dialog text and provided translations for German, Spanish, French, Japanese, Russian, and Chinese.


app/views/main_content_panel.py
```diff
@@ -3375,6 +3375,19 @@ def _on_do_build_steam_workshop_database(self) -> None:
 
     @Slot()
     def _do_run_game(self) -> None:
+        if self.mods_panel.active_mods_list.uuids != self.active_mods_uuids_last_save:
+            answer = dialogue.show_dialogue_conditional(
+                title=self.tr("Unsaved Changes"),
+                text=self.tr("You have unsaved changes. What would you like to do?"),
+                button_text_override=[self.tr("Save and Run"), self.tr("Run Anyway")],
+            )
+            if answer == "Save and Run":
```
Ocean456, line 3384: it doesn't work in other languages — you should use `self.tr`.

Ocean456: yeah, you added internationalization support for the button, but when selecting it, the return value answer is Сохранить и запустить in RU.
so it doesn't enter the condition branch.
you should also apply translation to the condition, i.e., self.tr("Save and Run").

Artemonim: I don't understand how I missed it 🤯

PR #1115, opened 2025-08-01
Format Files with Ruff

LionelColaso:


nyabinary:
Why not run a formatter workflow/check to check that every PR passes a formatter check?

LionelColaso:
> Why not run a formatter workflow/check to check that every PR passes a formatter check?

Not sure why super-linter did not catch it and passed one or more of the pulls

Feel free to create a pull if you know why

PR #1124, opened 2025-08-02
Refactor dependency graph generation with conflict resolution and remove cycle breaker

Ocean456:
- Refactored `gen_tier_two_deps_graph()`:
  - Explicit rules (`loadTheseBefore`) are collected and prioritized.
  - Inferred rules (About.xml `dependencies`) are collected only if enabled.
  - If a conflict is detected (e.g., `A -> B` from inferred but `B -> A` from explicit), the inferred rule is discarded.
  - Logs how many inferred rules were ignored due to conflicts.


app/sort/dependencies.py
```diff
@@ -220,79 +221,108 @@ def gen_tier_two_deps_graph(
...
+                for dep in loadTheseBefore:
+                    if isinstance(dep, tuple):
+                        if (
+                            dep[0] not in tier_one_mods
+                            and dep[0] not in tier_three_mods
+                            and dep[0] in active_mod_ids
+                        ):
+                            explicit_dependencies.add(dep[0])
+                    else:
+                        logger.error(f"loadTheseBefore entry is not a tuple: [{dep}]")
```
LionelColaso, line 255: I am wondering if showing a notification would be a good idea if there is an error instead of just logging it ?

LionelColaso: I leave it up to you to decide, you can keep as it is, and we could do the notification if future if required

app/sort/dependencies.py
```diff
@@ -220,79 +221,108 @@ def gen_tier_two_deps_graph(
...
+        for inferred_dep in inferred_rules.get(package_id, set()):
+            # Check for conflict: does any explicit rule say inferred_dep should load before package_id?
+            has_conflict = False
+
+            # Check if inferred_dep has an explicit rule that conflicts
+            if inferred_dep in explicit_rules:
+                if package_id in explicit_rules[inferred_dep]:
+                    # Conflict: explicit rule says inferred_dep -> package_id,
+                    # but we're trying to add package_id -> inferred_dep
+                    logger.debug(
```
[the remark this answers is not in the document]
Ocean456: > use logger.warning so it can be logged even if debugging is disabled

will do

PR #1125, opened 2025-08-02
Switch to uv for package management

hk21702:
This PR switches the package manager from plain `pip` to `uv`.

This enables faster and cleaner management of Python dependencies.

Additionally, this PR removes basic Python dependency management from the `distribute.py` script as people should really be interacting with `uv` directly. Consequently, the `--skip-pip` tag has been removed as that is just default behavior now. This may be a breaking change for certain workflows.

GitHub workflows have been updated to utilize `uv` and the old `requirements.txt` files have been removed.

For usage, run `uv sync` (By default, the dev dependency group is included). To also install build dependencies, run `uv sync --group build).

Note that building SteamworksPy using `distribute.py` is a bit wonky, though from what I can tell it was wonky before these changes already, so it isn't a regression.


LionelColaso (approved):
Yeah Works without issues

PR #1135, opened 2025-08-03
fix: Remove networkx-stubs dependency

hk21702:
Attempt number two to fix flaky linter complaints.


LionelColaso (approved):
Hope this fixes it for ever

Issue #1149, opened 2025-08-06
Order Inactive Mods


Manuel-S:
### Suggestion/proposed solution?

I'm missing a way to order the inactive mods, for example by last file change date, or by author.

For example it would make it much easier to find newly subscribed or recently updated mods.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

LionelColaso:
Can You explain in detail what exactly you want implemented ?

oceanfrost:
Closed in #1156

Issue #1086, opened 2025-07-29
Sort by folders' date modified/created


JulyMorning:
### Origin of idea

common sense

### Suggestion/proposed solution?

Sort by folders' date modified/created. It would be a lot easier to find recently dowloaded or updated mods that you want to add to your list.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

LionelColaso:
> ### Origin of idea
> 
> common sense
> ### Suggestion/proposed solution?
> 
> Sort by folders' date modified/created. It would be a lot easier to find recently dowloaded or updated mods that you want to add to your list.

read and continue the discussion here https://github.com/RimSort/RimSort/issues/1095#issuecomment-3146888251

LionelColaso:
Added in #1086

PR #1156, opened 2025-08-07
Add inactive mods sorting by modified time

averrin:


LionelColaso:
Also consider https://github.com/RimSort/RimSort/issues/1149 and https://github.com/RimSort/RimSort/issues/1086 seems to be in scope of your PR

LionelColaso:
To fix lint issues related to Ruff, you need to install ruff using `uv sync --group build`  and either run it manually or on file save
for more https://docs.astral.sh/ruff/formatter/

LionelColaso:
@averrin if you could consider saving the info/data generated such as size etc. take a look at https://github.com/RimSort/RimSort/pull/740 how it's saving the data/info

averrin:
> @averrin if you could consider saving the info/data generated such as size etc. take a look at #740 how it's saving the data/info

@LionelColaso 
Do you mean persist it between sessions? I dont think that it is necessary, you want to sort by this fields only if something changed, so it is better to update it every time (while it can be changed externally).

LionelColaso:
> > @averrin if you could consider saving the info/data generated such as size etc. take a look at #740 how it's saving the data/info
> 
> @LionelColaso Do you mean persist it between sessions? I dont think that it is necessary, you want to sort by this fields only if something changed, so it is better to update it every time (while it can be changed externally).

its ok i can be done later if its required you should go ahead and merge

PR #1168, opened 2025-08-09
[789] forcibly add steamworks.dylib into mac build

oceancabbage:
For some unknown reason, using Nuitka prefixes and postfixes, you can't directly specify `SteamworksPy.dylib`. If you do prefix `SteamworksPy` it won't work because it'll include the wrong processor when building for the other processor. I don't feel like changing [SteamworksPy itself](https://github.com/philippj/SteamworksPy/blob/master/steamworks/__init__.py#L78), so here's a brute force way of resolving the issue-- after building, copy the processor-postfixed `SteamworksPy*.dylib` to `SteamworksPy.dylib`.

This seems to work, putting `SteamworksPy.dylib` in the app (along with the processor-postfixed one), and SteamCMD actions work now + no more "missing SteamworksPy.dylib" exceptions.

Open to other suggestions.

### DIFF
--- a/distribute.py
+++ b/distribute.py
@@ -326,6 +326,8 @@ def copy_swp_libs() -> None:
         shutil.copyfile(STEAMWORKSPY_BUILT_LIB, STEAMWORKSPY_LIB_FIN)
 
 
+
+
 def get_latest_todds_release() -> None:
     # Parse latest release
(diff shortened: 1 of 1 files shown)


LionelColaso (approved):
Awesome

Issue #1167, opened 2025-08-09
Feature request: import mod list from save


bonesbro:
### Origin of idea

It would be helpful if you could import a mod list from a save.  Two use cases:
 - Upgrading to a new RW version, like 1.5 to 1.6
 - Swapping between saves with different mod lists.  RimWorld itself can do this, but you have to waste a few minutes loading the game first.

### Suggestion/proposed solution?

The very start of a .rws file has the ordered mod list with almost the same XML schema as a modlist file, so it's probably pretty straightforward...

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

oceancabbage:
You can actually already do this via `File -> Open Mod List...`. However, that one opens a RimSort directory by default. I've submitted [a PR](https://github.com/RimSort/RimSort/pull/1169) which adds a `File -> Import -> From Save file...` which opens up the RimWorld save folder by default. The functionality is the same though.

bonesbro:
You're awesome, thanks.

PR #1169, opened 2025-08-09
[1167] Import mod list from save

oceancabbage:
Introduces `File -> Import -> From Save File...`. This will open up your saves folder and you can select a `.rws` to import. If you are subscribed to all the mods it'll just load it in the active mods list. If you're missing some mods, the missing mods modal will show up with those mods, and if you have SteamDB configured (and the mods exist in the DB), you will get the option of downloading those mods.

The functionality here is basically the same as (and ripped of from) `_do_import_list_file_xml`, which is used in the `File -> Open Mod List` feature, except this one opens the save directory. We can extract out the common code into a helper function if folks feel strongly.

Related issue: https://github.com/RimSort/RimSort/issues/1167


LionelColaso (approved):
That extra effort, looks good though

Issue #1170, opened 2025-08-09
Highlight new mods


averrin:
### Origin of idea

I'd like to knew what mods i've just added. And why my save lags af.

### Suggestion/proposed solution?

The only reliable and not too complicated way i imagined is comparing the mod list with the recent save.
As always, i made a small PoC

<img width="521" height="764" alt="Image" src="https://github.com/user-attachments/assets/2a6b985c-3ee1-4c49-a76b-5e1de4d7d0e6" />

### Alternatives considered

Messing with saved mod lists or something like that, but it requires autosave of lists, persistence through sessions and so on

### Additional context

The "recent save" conception isnt very solid though.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

oceancabbage:
This is a sweet idea. Definitely worth a toggle for enabling this feature. We can search the saves folder and pick the latest `.rws` file, and compare the loaded mod list with the active mod list. Any differences can be marked with `new` and counted in the little new counter thing on the bottom right. Definitely consider making a pull request with your POC, or let me know if you'd like me to give it a shot.

PR #1174, opened 2025-08-09
Compare modlists with the recent save

averrin:
A PoC for #1170


app/controllers/settings_controller.py
```diff
@@ -108,6 +116,14 @@ def __init__(self, model: Settings, view: SettingsDialog) -> None:
...
         self.settings_dialog.settings_custom_height_spinbox.setEnabled(True)
 
+        # Advanced: wiring for save-comparison indicator toggle
+        try:
+            self.settings_dialog.show_save_comparison_indicators_checkbox.toggled.connect(
+                self._on_toggle_show_save_comparison_indicators
+            )
+        except Exception:
+            pass
+
```
oceancabbage, line 126: This seems to be duplicate code from 85-92 in the same file?

LionelColaso:
Just a suggestion If you want a clean merge, you can rebase your branch with main

averrin:
Honestly, it don't like rebasing. But if it is required,  I use it.

LionelColaso:
> Honestly, it don't like rebasing. But if it is required, I use it.

It's not a requirement just do whatever you prefer

LionelColaso:
@averrin it does not work when I add new mod, only detects the new mod when I click sort
is it intentional ?
Between even if you don't update the branch the auto merge will do it automatically unless there is a conflict

averrin:
@LionelColaso no, it worked a couple of hours ago, but not now. I will try to figure out what happened

averrin:
@LionelColaso looks like i didnt test it well after lint fixes (but i was pretty sure that i did...)
i didnt figureout how to run mypy like in the pipeline (it throws a huge amount errors about everything). 
I hate coding on windows =/

LionelColaso:
> @LionelColaso looks like i didnt test it well after lint fixes (but i was pretty sure that i did...) i didnt figureout how to run mypy like in the pipeline (it throws a huge amount errors about everything). I hate coding on windows =/

If you are using vs code you can use the settings.json file which is already configured 
you can also test with mypy locally as well using 
`mypy --config-file pyproject.toml --install-types --non-interactive app/`
considering you if you only modifying in app folder
you can also run ruff in similar way or configure it to run on file save
`ruff format app/`

LionelColaso:
Works now
but does not show Count unless there is error or warning since you hooked it into error and warning 
overall its cool
image
image

averrin:
Heh, i always have warnings and didnt even know that this panel can hide...

LionelColaso:
> Heh, i always have warnings and didnt even know that this panel can hide...

Lol that's Hilarious

LionelColaso (changes requested):
@averrin you will have to resolve conflicts and let me know when it's ready

averrin:
@LionelColaso I found some detail. Initially i made settings flag for showing "new" icons in the list. But it doesnt affect the filter that works with independently of this flag. 
Is it ok? Im not sure is the setting necessary at all, or should it disable the feature completely.

LionelColaso:
> @LionelColaso I found some detail. Initially i made settings flag for showing "new" icons in the list. But it doesnt affect the filter that works with independently of this flag. Is it ok? Im not sure is the setting necessary at all, or should it disable the feature completely.

Why would you need settings to enable this unless there is a huge performance loss

LionelColaso:
so basically you do not need to have a toggle for user just save it by default 
its just a new mod icon like error and warning it should be used by default not a user choice 
image

averrin:
> > @LionelColaso I found some detail. Initially i made settings flag for showing "new" icons in the list. But it doesnt affect the filter that works with independently of this flag. Is it ok? Im not sure is the setting necessary at all, or should it disable the feature completely.
> 
> Why would you need settings to enable this unless there is a huge performance loss

1. I don't think that there is a huge impact, but save files and modlists can be big. On the other hand, I didn't experienced any slowness 
2. There is jcpd-something linter that errors on similar lines. How should I satisfy it? I do not like the idea to use the only function there, but on the other hand, if I understand it correctly, inactive list handler does nothing right now

LionelColaso:
Between it works but if you make it work on its own by creating similar function as error and warning and without reusing it would be great
image

LionelColaso:
>     2. There is jcpd-something linter that errors on similar lines. How should I satisfy it? I do not like the idea to use the only function there, but on the other hand, if I understand it correctly, inactive list handler does nothing right now

so basically its complaining of similar code so you can just refactor them and create a new function and call that function 
image

averrin:
I feel that it is nice to have a setting to disable potentially heavy and optional operation. So i make it stronger.

LionelColaso:
> I feel that it is nice to have a setting to disable potentially heavy and optional operation. So i make it stronger.

Then keep it if it affects performance you can move it to performance tab 
image

LionelColaso:
> I feel that it is nice to have a setting to disable potentially heavy and optional operation. So i make it stronger.
also, this can be one function for both mod list since it's basically doing the same thing
476339447-eb277362-8fae-4625-9420-60ca227f363a

averrin:
I've done both things
image

LionelColaso (approved):
Awesome

LionelColaso:
@averrin can you do something about this has a performance impact

image

averrin:
> @averrin can you do something about this has a performance impact
> 
> image

that is from the previous MR, but yeah, my bad. I hope it is better now.

LionelColaso:
> > @averrin can you do something about this has a performance impact
> > image
> 
> that is from the previous MR, but yeah, my bad. I hope it is better now.

i am sorry the performance tab just got renamed https://github.com/RimSort/RimSort/pull/1184
You u may have conflicts

averrin:
> > > @averrin can you do something about this has a performance impact
> > 
> > 
> > that is from the previous MR, but yeah, my bad. I hope it is better now.
> 
> i am sorry the performance tab just got renamed #1184 You u may have conflicts

It is kinda surprising. And what is the plan now? I do not think, that my setting fits the new tab name. I think the previous one was better...

PR #1176, opened 2025-08-09
Speed up filtering by ~92% on large modlists

### LINKED ISSUE #1063: Improve mod filter performance

On the main screen, there are a number of filter options...

Image

As you can see, I download a lot of mods, and at this point, using any filter causes the process to soft lock for about 3 seconds before it finally loads.

Possibly related is adding a mod to your active mods or removing it from the active mods, which also leads to a long delay that soft locks the process.

Some performance improvements are needed.

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

Patryk-Malinowski:
Closes #1063 

On an inactive modlist of 7256 mods.

Before it took ~0.99 seconds after each letter is typed in search filter. (Cached)

After it takes ~0.08 seconds after each letter is typed in search filter. (Cached)

If I did the math right its around 90%+ increase in speed.

Similar speed increase applies to using the filters (Steam mod, git mod, C#/XML etc.) The C#/XML mod filter is a bit slower, but I think that might be somewhere down the chain where GUI gets updated?

(For both before and after, the first letter/filter application is slower until QT caches stuff. Refreshing modlist deletes the cached stuff. Either way its much quicker now)

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -1216,13 +1216,13 @@ class MainContent(QObject):
                 self.mods_panel.data_source_filter_icons
             )
             self.mods_panel.signal_clear_search(
-                list_type="Active", recalculate_list_errors_warnings=False
+                list_type="Active",
             )
             self.mods_panel.inactive_mods_filter_data_source_index = len(
                 self.mods_panel.data_source_filter_icons
             )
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Yeah, Does have a noticeable improvement then before

PR #1182, opened 2025-08-10
Fix SQLite failing to open Aux database when it doesn't exist

oceancabbage:
Was running into infinite crashes on startup. Turns out you can get in a state where SQLite fails to open the aux metadata database because the instance directory for the file (e.g., …/instances/Default/) doesn't exist yet. I updated `MetadataDbController.__init__` to create the database file’s parent directory (`mkdir(parents=True, exist_ok=True)`) before initializing the SQLAlchemy engine. This ensures the path exists and prevents the “unable to open database file” error.

### DIFF
--- a/app/controllers/metadata_db_controller.py
+++ b/app/controllers/metadata_db_controller.py
@@ -12,7 +12,17 @@ from app.utils.steam.steamfiles.wrapper import acf_to_dict
 
 class MetadataDbController:
     def __init__(self, db: Path | str) -> None:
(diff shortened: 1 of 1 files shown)


LionelColaso (changes requested):
waiting response

oceancabbage:
> mhh not sure it should be here or in app_info.py or aux_db_utilspy

I don't have strong feelings. I thought I'd just keep it in the same file since it's a small fix.

LionelColaso (approved):
Looks good

PR #1183, opened 2025-08-10
Revise translation

Ocean456:
Enhances translation support by updating Chinese translations, adding Turkish translations, improving UI internationalization, and refining text handling.


app/views/dialogue.py
```diff
@@ -173,10 +170,6 @@ def show_information(
     info_message_box.setTextFormat(Qt.TextFormat.RichText)
     info_message_box.setIcon(QMessageBox.Icon.Information)
     info_message_box.setObjectName("dialogue")
-    info_message_box.setStandardButtons(QMessageBox.StandardButton.Ok)
-    info_message_box.button(QMessageBox.StandardButton.Ok).setText(
-        QCoreApplication.translate("show_dialogue_information", "OK")
-    )
```
oceancabbage, line 179: Are these no longer needed?

Ocean456: > Are these no longer needed?

should be, as the qbase translation file has already been introduced. these basic elements will have translations in there

oceancabbage (approved):

PR #1188, opened 2025-08-10
Add option for clear to move DLC as well

oceancabbage:
Someone on the RimWorld Discord mentioned they'd like the ability for Clear to also move RimWorld DLC. Adding a toggle in Settings -> Advanced that allows you to do this.


LionelColaso (approved):
Works

Issue #1187, opened 2025-08-10
Support for alternativePackageIds


oelsart:
### Origin of idea

Currently, RimSort does not seem to support the new 1.6 tag `alternativePackageIds`.

### Suggestion/proposed solution?

Support `alternativePackageIds` and disable Missing Dependencies errors when the mod with the package id specified in the tag is active.

### Additional context

Some mods (Including mine) have another package id as a dev version. Like [Humanoid Alien Races](https://steamcommunity.com/sharedfiles/filedetails/?id=2640710953), [MultiFloors](https://steamcommunity.com/sharedfiles/filedetails/?id=3537626590) and [Vehicle Map Framework](https://steamcommunity.com/sharedfiles/filedetails/?id=3528683767)

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

LionelColaso:
I guess you are basically asking support for this right https://rimworldwiki.com/wiki/Modding_Tutorials/About.xml

<img width="1103" height="326" alt="Image" src="https://github.com/user-attachments/assets/a6c4278b-d7fe-47c3-9c10-264f8e3ec796" />

oelsart:
Yes, It's in `modDependencies`. I forgot to mention.

oceancabbage:
@oelsart can you provide some examples of mods that use `alternativePackageIds`? I've a draft PR that I need to do some testing on with such mods.

oelsart:
@oceancabbage 
Sorry, I could only find one so far.
mod: [Map Vehicles](https://steamcommunity.com/sharedfiles/filedetails/?id=3426502635)
dependencies : [Vehicle Map Framework](https://steamcommunity.com/workshop/filedetails/?id=3426502333), [[Dev] Vehicle Map Framework](https://steamcommunity.com/sharedfiles/filedetails/?id=3528683767)

I'll keep looking a little more.

oceancabbage:
No worries, this will do for testing, tysm

oelsart:
I searched for a list of mods over 2500, but it seems that it is only used here lol. A few people on the RimWorld Discord server are hoping for support, so if RimSort supports it, more people will use it. Looking forward to the update!

oceancabbage:
[...]

PR #1191, opened 2025-08-11
Support for alternativePackageIds

oceancabbage:
This PR attempts to add support for the new 1.6 `alternativePackageIds`, which is a new field in `modDependencies`, e.g.

```
<modDependencies>
  <li>
    <packageId>erdelf.HumanoidAlienRaces</packageId>
    <displayName>Humanoid Alien Races</displayName>
    <steamWorkshopUrl>steam://url/CommunityFilePage/839005762</steamWorkshopUrl>
    <alternativePackageIds IgnoreIfNoMatchingField="True">
      <li>erdelf.HumanoidAlienRaces.dev</li>
    </alternativePackageIds>
  </li>
</modDependencies>
```

Internally, dependencies are now stored as a list of either a plain packageId or a tuple of (packageId, {"alternatives": set[...]}) so we can represent a primary dependency alongside its declared alternatives.

Alternative handling is strictly opt-in via a new Advanced setting: “Consider alternativePackageIds as satisfying dependencies.” When enabled, missing-dependency checks accept either the primary dependency or any declared alternative as satisfying the requirement. The “Check Dependencies” button, the optional pre‑sort dependency check, and the per‑mod error tooltips all respect this toggle. Tooltips also gain an “Alternative Dependencies” section to help users discover valid substitutes when the primary is missing (hidden when the toggle is off).

Sorting also respects alternatives when two settings are enabled: using About.xml dependencies for sorting, and the Advanced alternative toggle. In that case, tier‑two sorting will prefer the primary dependency if it’s active; otherwise, itwill use the first active alternative to build an inferred load-before edge. Explicit load rules remain authoritative and take precedence over inferred edges.

Additional stability fixes were made around JSON logging of mod info by teaching the flattener to convert tuples (and nested sets) before serialization. Default behavior remains unchanged unless the Advanced alternative setting is enabled.

TL;DR:
Net effect summary
- Default behavior is unchanged unless the Advanced alternative setting is enabled.
- When enabled, alternatives can satisfy dependencies in Check Dependencies, pre‑sort checks, and tier‑two inferred sorting; tooltips also surface alternative options for unmet deps.

Relevant issue: https://github.com/RimSort/RimSort/issues/1187.

### DIFF
--- a/app/controllers/main_window_controller.py
+++ b/app/controllers/main_window_controller.py
@@ -73,6 +73,12 @@ class MainWindowController(QObject):
         # Create a dictionary to store missing dependencies
         missing_deps: dict[str, set[str]] = {}
 
+        # Precompute active package IDs for quick lookup
+        active_ids = {
+            self.metadata_manager.internal_local_metadata[u].get("packageid")
+            for u in active_mods
+        }
+
         # Check each active mod's dependencies
         for uuid in active_mods:
             mod_data = self.metadata_manager.internal_local_metadata[uuid]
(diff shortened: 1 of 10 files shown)


oceancabbage:
I tested this out with https://steamcommunity.com/sharedfiles/filedetails/?id=3426502635, as suggested on the linked issue. Seems to work as expected. Peep screenshots:

Screenshot 2025-08-10 at 10 16 41 PM

Screenshot 2025-08-10 at 10 17 05 PM

Screenshot 2025-08-10 at 10 17 22 PM

LionelColaso (approved):
cool

Issue #1150, opened 2025-08-06
Mod Dependency Error but the mod doesn't require dependency for specific versions


RayVee51:
### Release Type

Compiled (Windows x86-64)

### Version

1.0.35

### Operating System

Windows 10

### Relevant logs

Not needed

### Describe the bug

Example: So there is this mod called Long distance https://steamcommunity.com/sharedfiles/filedetails/?id=2028511550 that requires Hugslib for older versions but not for 1.4, 1.5 and 1.6. It still displays an error in Rimsort.

<img width="490" height="51" alt="Image" src="https://github.com/user-attachments/assets/eab25e88-7524-4544-ac4f-ab0f64ff41e2" />

<img width="1920" height="911" alt="Image" src="https://github.com/user-attachments/assets/8f6a0b20-3e4a-4d69-a742-87e64556254a" />

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

LionelColaso:
@RayVee51 
It's not a bug on RimSort side the mod author has a defined a `modDependencies `rule which indicates that it needs `hugslib ` no matter the game version
you can delete it and it will stop complaining

<modDependencies>
        <li>
            <packageId>UnlimitedHugs.HugsLib</packageId>
            <displayName>HugsLib</displayName>
            <steamWorkshopUrl>steam://steamcommunity.com/workshop/filedetails/?id=818773962</steamWorkshopUrl>
        </li>
    </modDependencies>

<img width="1262" height="934" alt="Image" src="https://github.com/user-attachments/assets/534bae03-733b-4d0b-b281-fbb30007266e" />

LionelColaso:
For more context, https://rimworldwiki.com/wiki/Modding_Tutorials/About.xml

<img width="1099" height="827" alt="Image" src="https://github.com/user-attachments/assets/df2ba64d-07f8-4e9d-90d4-5a5027cd134c" />

LionelColaso:
This should not be an issue anymore, please confirm if v1.0.36 fixes the issue for you

LionelColaso:
This is due to `modDependencies `tag in the About file as well as blank `modDependenciesByVersion `tag for v1.6 which at least i have not seen before in any other mod

```
<modDependencies>
      <li>
        <packageId>UnlimitedHugs.HugsLib</packageId>
[...]

Issue #1074, opened 2025-07-26
Use Dependency rules for sorting can cause circular dependency


AJarOfDirt0:
### Release Type

Compiled (Windows x86-64)

### Version

1.0.31

### Operating System

Windows 10

### Describe the bug

Found specifically with Reel's Facial Animation Textures, they have a dependency that must be loaded after. The sorting considers both the dependency and the load after and lists it as a circular dependency. Sorting needs to have a condition to ignore the dependency sorting if it conflicts with loadafter setting.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

Issue #988, opened 2025-07-08
Circular Dependency check pings on mods being in both modDependency and loadBefore of About.xml


Wyrd-Oh:
### Release Type

Compiled (Windows x86-64)

### Version

v1.0.22

### Operating System

Windows 10

### Describe the bug

The circular dependency check treats something listed in "modDependencies" but also listed in the "LoadBefore" section as an error. This may be an edge case because of topological sorting but deliberate on the mod authors part.

Pictured first is the About for reel.facialanims and second is vanillaexpanded.vtexe.facialanims - The first is the triggering one and the second is provided as reference as it does not include the first as a dependency or in the loadBefore/loadAfter

<img width="239" height="296" alt="Image" src="https://github.com/user-attachments/assets/05971a44-5630-4dfd-a7d9-7de3dc529521" /> 

<img width="856" height="693" alt="Image" src="https://github.com/user-attachments/assets/28ec25c5-18ed-4870-8c0f-1b8f59941acf" />

<img width="854" height="793" alt="Image" src="https://github.com/user-attachments/assets/9a90b6da-b701-4868-9b9b-aea8811046ca" />

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

LionelColaso:
`Problem`
As you noted and i checked as well
[Reel's Facial Animation Textures](https://steamcommunity.com/sharedfiles/filedetails/?id=3118315633) has [Vanilla Textures Expanded - [NL] Facial Animation](https://steamcommunity.com/sharedfiles/filedetails/?id=2816938779) both as its `dependency` and `load before` rule for `Vanilla Textures Expanded - [NL] Facial Animation` in its `About.xml` file which is causing the loop

`Solution`
1) Disable  `Use dependency rules for sorting` and Restart RimSort it will fix the issue
2) you can also ask the mod author to check the wrong load rule 

<img width="1102" height="704" alt="Image" src="https://github.com/user-attachments/assets/9c99156f-46c0-4b95-9b15-c818f0605ddd" />

Issue #961, opened 2025-06-29
Community Rules sorting problems


Lorsieab2:
### Release Type

Compiled (Windows x86-64)

### Version

1.0.21

### Operating System

Windows 11

### Describe the bug

I am using a huge modlist +1500 mods. I keep getting sorting errors with the following groups of mods:
Alpha Animals, Combat Extended and Pawnmorpher 
Reel's Facial Animation Textures and Vanilla Textures Expanded - [NL] Facial Animations
Both of these groups give a "circular dependencies" error. I've been able to fix the Alpha Animals-Combat Extended-Pawnmorpher error by deleting the "loadAfter" entry for Pawnmorpher in Alpha Animals' community rules, but I can't fix the Reel's and Vanilla Textures Expanded error.

![Image](https://github.com/user-attachments/assets/3f867b45-30a5-4976-bec8-ba420dc32385)

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

Lorsieab2:
Forgot to say this is topological sorting.
There's a couple of mods that cause "Unable to Sort" errors, like Early Times Continued, Vanilla Armor Expanded, Gerrymon's Vanilla Upscaled Textures, etc.

LionelColaso:
@Lorsieab2 solution to your issue is mentioned here https://github.com/RimSort/RimSort/issues/988

PR #1210, opened 2025-08-14
Add toggle for versioned About.xml tags to take precedence

oceancabbage:
Addresses https://github.com/RimSort/RimSort/issues/1150

Changes:
- ByVersion precedence: Implemented toggleable support across dependencies, incompatibilities, loadAfter/loadBefore, and description.
 - ON: use matching vX.Y only; empty = no requirement; no match = fallback to base; non-additive.
 - OFF: ignore all byVersion tags; use base only (preserves prior behavior).
- Safety: Guarded game version parsing and byVersion shapes; use safe `.get('li')` with empty defaults; no crashes on missing/empty tags.
- Idempotency: Dependencies, incompatibilities, and load rules use sets; no duplicates when base/byVersion overlap.
- Warnings: Aligned with toggle and precedence; byVersion empties/missing produce debug logs only; base-tag warnings unchanged.

Tested by going into some mod's About.xml, adding a `loadBeforeByVersion` `v1.6` entry for some mod that was loaded before. When toggle is OFF, nothing is changed. When toggle is ON, an error shows and sorting puts the mod after, as expected.

This is kinda complicated so hopefully I don't break anything. I put all of this behind a toggle just in case.

### DIFF
--- a/app/controllers/main_window_controller.py
+++ b/app/controllers/main_window_controller.py
@@ -1,4 +1,5 @@
 import os
+from xml.etree.ElementTree import Element
 
 from loguru import logger
 from PySide6.QtCore import QObject, Slot
@@ -63,6 +64,39 @@ class MainWindowController(QObject):
         EventBus().reset_use_this_instead_cache.connect(
             self.on_reset_use_this_instead_cache
         )
+
+    def _parse_workshop_id(self, url: str) -> str | None:
+        """Extract a Steam Workshop ID from a dependency URL."""
+        workshop_id: str | None = None
+        if "?id=" in url:
+            workshop_id = url.split("?id=")[1]
+        elif "CommunityFilePage/" in url:
+            workshop_id = url.split("CommunityFilePage/")[1]
+        else:
+            return None
+        # Clean up trailing query or path
+        if "?" in workshop_id:
+            workshop_id = workshop_id.split("?")[0]
+        if "/" in workshop_id:
+            workshop_id = workshop_id.split("/")[0]
+        return workshop_id
+
+    def _find_workshop_id_in_deps(
+        self, deps_node: Element, target_pkg_id: str
+    ) -> str | None:
+        """Search a <modDependencies> or versioned child node for a matching packageId and return its Workshop ID."""
+        target = target_pkg_id.lower()
+        for dep in deps_node.findall("li"):
+            package_id = dep.find("packageId")
+            if (
+                package_id is not None
+                and package_id.text is not None
+                and package_id.text.lower() == target
+            ):
+                workshop_url = dep.find("steamWorkshopUrl")
+                if workshop_url is not None and workshop_url.text is not None:
+                    return self._parse_workshop_id(workshop_url.text)
+        return None
 
     def check_dependencies(self) -> None:
         # Get the active mods list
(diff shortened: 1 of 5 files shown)


divineDerivative:
Why does there need to be a toggle for this? Turning it off would result in incorrect behavior.

oceancabbage:
> Why does there need to be a toggle for this? Turning it off would result in incorrect behavior.

@divineDerivative because I have a feeling a lot of mods / mod authors aren't doing this "according to the docs". i.e. we're going to get a lot more issues on GitHub complaining about missing dependencies or what not if we make this the default option.

divineDerivative:
> @divineDerivative because I have a feeling a lot of mods / mod authors aren't doing this "according to the docs". i.e. we're going to get a lot more issues on GitHub complaining about missing dependencies or what not if we make this the default option.

So instead you want to be informing users of missing dependencies or incompatibles that aren't real, not reporting actual dependencies/incompatibles, or applying sorting rules that are not correct? When it comes to applying rules from About.xml, RimSort should be doing exactly what the game is doing. Anything else is misleading to users.

RimSort should not be trying to proactively 'fix' problems it thinks mod authors might make. That's what the community rules are for, when it is confirmed to be necessary for a specific mod. I'll point to the whole "treat dependencies as load before" thing, which I told ya'll was a terrible idea when it was first proposed.

oceancabbage:
> > @divineDerivative because I have a feeling a lot of mods / mod authors aren't doing this "according to the docs". i.e. we're going to get a lot more issues on GitHub complaining about missing dependencies or what not if we make this the default option.
> 
> So instead you want to be informing users of missing dependencies or incompatibles that aren't real, not reporting actual dependencies/incompatibles, or applying sorting rules that are not correct? When it comes to applying rules from About.xml, RimSort should be doing exactly what the game is doing. Anything else is misleading to users.
> 
> RimSort should not be trying to proactively 'fix' problems it thinks mod authors might make. That's what the community rules are for, when it is confirmed to be necessary for a specific mod. I'll point to the whole "treat dependencies as load before" thing, which I told ya'll was a terrible idea when it was first proposed.

Can you explain to me how this is proactively fixing problems we think mod authors might make? The initial/current implementation is just wrong, according to the docs. But that's what thousands of RimSort users have been using since I started this project. After being informed that we're missing the correct functionality, I've gone ahead and implemented it, as requested, but kept it behind a feature flag (for now), out of consideration that some users might be confused why, all of a sudden, they're getting warnings or errors not seen before, or why their sorting order is different. I see this as a cautious rolling-out of a fix for a bug that, for years, I didn't really feel any impact of from the community.

I don't recall what the issue with "treat dependencies as load before" was (or is). AFAIK that's behind a feature flag as well. That way everyone is happy. Isn't that the best we can do, as people who are doing this for free, for the benefit for others?

LionelColaso (approved):
Looks Good

divineDerivative:
> Can you explain to me how this is proactively fixing problems we think mod authors might make?

Because you literally said that's what you're doing.

> @divineDerivative because I have a feeling a lot of mods / mod authors aren't doing this "according to the docs". i.e. we're going to get a lot more issues on GitHub complaining about missing dependencies or what not if we make this the default option.

I really feel like the issues submitted for this option being off are going to far outweigh issues submitted for it being on. The issue that prompted this PR, [1150](https://github.com/RimSort/RimSort/issues/1150), was from a mod that was using the tags correctly but was getting sorted wrong because RimSort didn't consider those tags. You admit that this is a bug, but you want a toggle for whether it's fixed or not? The reason it wasn't an issue before now is because it's just not a common situation for a mod to have different requirements per version. But when they do, I think sorting it correctly in the first place is better than having users come here to ask why and telling them there's a toggle for it. I can't image a mod author using the by version tags on accident. Most don't even know they exist.

> that some users might be confused why, all of a sudden, they're getting warnings or errors not seen before, or why their sorting order is different

Just inform them that a change was made because RimSort wasn't recognizing rules it should have been.

> I don't recall what the issue with "treat dependencies as load before" was (or is). AFAIK that's behind a feature flag as well.

[1074](https://github.com/RimSort/RimSort/issues/1074) [988](https://github.com/RimSort/RimSort/issues/988) [961](https://github.com/RimSort/RimSort/issues/961) There were a lot of people coming to the Discord for this as well. There were three PRs to fix the issue that was caused by adding that "feature".

I think the correct response to users complaining that their sort order doesn't look right to them is to educate them on how sorting works. Not to add made up rules to make it look how they expect. As long as their game still runs correctly, it's fine. And if the new sorting does result in issues, they can report to the mod author or add a sorting rule.

I'm not trying to come off as harsh, but as a mod author, I expect the rules I put in my About.xml to be respected. RimSort should be using those rules the same way the game does. alternativePackageIds is another one that should not be a toggle, at least not for 1.6, since that is the new default behavior. If you're concerned about it causing unexpected issues, leave it as an Edge release for a week or so.

oceancabbage:
> Because you literally said that's what you're doing.

@divineDerivative No, I don't agree with you. When I make a significant change to something as **core** as error/warning detection and sorting for a mod manager, I'm going to do it in a way that's as least intrusive as possible, while basing future decisions such as make the default option ON or removing the option entirely on user feedback over multiple versions. If you interpreted what I said as "proactively fixing problems", it probably comes down to how you were trained or the kind of work you do professionally; I see it as avoiding issues from a sudden rollout of a potentially breaking change. 

As a mod author you probably know this, but there are people who don't pay much attention to anything, and are definitely NOT going to be testing an Edge release or reading any announcements from the RimSort team. So, when they come to me and say "hey, why's my mod list order changed all of a sudden, something broke and I liked how it was sorting previously because everything worked", I can just say "oh go and turn this toggle off", without having to tell them "well, sucks, nothing I can do now!" Or maybe we get an influx of issues like the one that spawned this PR, in which case it would be totally reasonable to make that option default ON, and if nobody complains for a while, remove that option entirely.

I understand where you're coming from-- the behavior was wrong to begin with. But even fixing behaviors that are wrong can be done with a gradual rollout strategy over what you're suggesting. Also, the "mod dependency as load order" thing isn't even a good analogy because that was an unjustified sorting decision we added on top of metadata compilation. Here, I'm literally fixing a bug and rolling it out slowly.

PR #1190, opened 2025-08-10
Optimize comparing last save

Patryk-Malinowski:
Optimize retrieving package ids from last save

PR #1219, opened 2025-08-17
Adds toggle for advanced filtering to improve mod panel performance

Ocean456:
Introduces an option to enable or disable advanced filtering features, allowing users to control whether additional mod data (e.g. folder size, author, modified date) is shown in the mods panel. Disabling advanced filtering hides related UI elements and skips heavy calculations, addressing performance concerns for large mod sets.

(I made this an option because I don’t like it scanning the whole disk and it makes the two mod panels different in size)

https://github.com/RimSort/RimSort/pull/1190#issuecomment-3194035286


LionelColaso (approved):
Looks Good

Issue #1083, opened 2025-07-28
Update fails when using custom temp folder for Pip


Gazzoo-byte:
### Release Type

Compiled (Windows x86-64)

### Version

1.0.29

### Operating System

Windows 11

### Describe the bug

The RimSort updater appears to be looking for the downloaded update in the wrong folder. Before the application closes it correctly lists the custom filepath I have set as a temp folder for Pip

<img width="309" height="164" alt="Image" src="https://github.com/user-attachments/assets/ab679080-f0e9-4261-85bb-21697c9c89b0" />

However the "Update source" does not match this, causing the update to fail

<img width="1118" height="628" alt="Image" src="https://github.com/user-attachments/assets/35425968-990f-4e91-8998-ebec3b180abf" />

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

griffinmathews-personal:
I'm having a similar issue on mac - except nothing happens after selecting to proceed with the update. Console logs show nothings happening and the updater doesn't start.

WiiTee:
Also happens on 1.0.32 --> 1.0.33

YAMLcase:
Getting it too on Windows 11.  I don't think it's a temp pip location issue, I'm using defaults.  1.0.32 -> 1.0.33

```
Successfully retrieved latest release. The update will be installed from: C:\Users\username\AppData\Local\Temp\RimSort
```

Freezly-max:
Same issue here on 1.0.32 -> 1.0.33, Win11 as well.

LionelColaso:
The update broke due to the new introduction of .msi builds for windows and RimSort downloading the .msi builds instead of .zip will be fixed soon

BansheeStudio:
ahhhhhh same here

zachinglis:
Same here. Creating the folder than gives me `ERROR: RimSort.exe not found in update source folder.`

Runzu87:
Same issue here with:

Windows 10

RimSort Version: 1.0.33 (and the previous version, I manually updated)

PR #1230, opened 2025-08-19
Handling for custom Python temp path

Gazzoo-byte:
On Windows, Python can have a custom temporary folder set by using the system Environment Variable %TMPDIR%, this change checks if %TMPDIR% exists and if so uses it as the update_source_folder, otherwise use the regular %TEMP% path.

Resolves the issue I was having with https://github.com/RimSort/RimSort/issues/1083


LionelColaso (approved):

LionelColaso:
Thanks good for merge

PR #1271, opened 2025-08-31
Revise translation

Ocean456:
enables full i18n support for Player Log UI and labels

completed the Chinese translations and added some Turkish translations from MHS.


LionelColaso (approved):
Looks Good

PR #1313, opened 2025-09-12
Handle Zstandard (zstd) save files

AmCh-Q:
Based off of PR https://github.com/RimSort/RimSort/pull/1190 for GZip, but adding Zstandard support to support save files produced by my mod [Save File Compression](https://github.com/AmCh-Q/RimWorldMod_SaveFileCompression).
Reference: https://github.com/RimSort/RimSort/pull/1190#issuecomment-3273658567

This PR requires [Zstandard bindings for Python](https://pypi.org/project/zstandard/) as a dependency -- Python 3.14 adds native ZStandard support, but I presume there won't be an upgrade to 3.14 soon.
I did not add the dependency myself because I don't know how.

Sample test save file compressed with Zstandard: [test1.6-zstd.zip](https://github.com/user-attachments/files/22306036/test1.6-zstd.zip) (the actual zstd-compressed save file is inside the zip package, unzip it first)

I am not very familiar with Python programming and the contribution rules here, so please forgive me if I am making any mistakes.

### DIFF
--- a/app/utils/xml.py
+++ b/app/utils/xml.py
@@ -1,9 +1,9 @@
 import gzip
-import zstandard as zstd
 import os
 from typing import Any
 
 import xmltodict
+import zstandard as zstd
 from bs4 import BeautifulSoup
 from loguru import logger
 from lxml import etree
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "toposort==1.10",
     "watchdog==6.0.0",
     "xmltodict==0.15.1",
+    "zstandard==0.23.0",
 ]
 
 [tool.uv.sources]
@@ -80,7 +81,7 @@ warn_unused_ignores = true
 exclude = ['venv', '.venv', 'submodules']
 
 [[tool.mypy.overrides]]
-module = ['steam.*', 'steamfiles.*', 'pytestqt.qtbot']
+module = ['steam.*', 'steamfiles.*', 'pytestqt.qtbot', 'zstandard']
 ignore_missing_imports = true
 
 [tool.ruff]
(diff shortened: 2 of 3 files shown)


LionelColaso:
> This PR requires [Zstandard bindings for Python](https://pypi.org/project/zstandard/) as a dependency -- Python 3.14 adds native ZStandard support, but I presume there won't be an upgrade to 3.14 soon. I did not add the dependency myself because I don't know how.

You don't need to add it since its already added https://github.com/RimSort/RimSort/blob/09e075120d99f0525b385e4bb5ddcf144c09d78f/uv.lock#L1169

the link error you are getting is related to ruff just run the below command to format
```
ruff check --fix app/
run format app/
```

AmCh-Q:
Thanks for letting me know of the dependency situation!
Unfortunately, I am completely new to Ruff and the concept of linting in general...
Let me know if there is something that really should be done on my end, otherwise I'd just trust that someone more skillful in that would fix/has already fixed that. =w=

LionelColaso:
> Thanks for letting me know of the dependency situation! Unfortunately, I am completely new to Ruff and the concept of linting in general... Let me know if there is something that really should be done on my end, otherwise I'd just trust that someone more skillful in that would fix/has already fixed that. =w=

I am trying to add the formatter to fix this automatically in the pull request, so hopefully it will get fixed automatically
If it ok with you i may do some changes incase if it required if there are any mypy or ruff errors later 
the changes on your part is good as it its 

incase you want to know about ruff https://docs.astral.sh/ruff/formatter/ and mypy https://github.com/python/mypy

AmCh-Q:
Thanks for the resources on ruff and mypy.
Go ahead and make any changes you need to.

PR #1320, opened 2025-09-14
Add language_data_folder property and set Qt env vars

LionelColaso:
Introduces a language_data_folder property in AppInfo for cleaner access to language data paths. Sets environment variables in main_thread to address Qt issues, including translation path, backend, and Chromium flags. Updates AppController to use the new property for translator initialization.


LionelColaso:
not required fixed in https://github.com/RimSort/RimSort/pull/1389

PR #1358, opened 2025-09-28
Add backup saves functionality

oceanfrost:
Introduce a feature that allows automatic backups of save files upon the first daily launch.


LionelColaso:
I like the idea but would suggest 
1) add a user defined limit to the number of backup saves to be stored just like screenshot you can check https://github.com/RimSort/RimSort/pull/1361 for how to incase you want to implement it 
image

2) the setting should be in Auxiliary DB tab since it already has saves related settings
image

oceanfrost:
> I like the idea but would suggest 
> 1) add a user defined limit to the number of backup saves to be stored just like screenshot you can check https://github.com/RimSort/RimSort/pull/1361 for how to incase you want to implement it 
> image
> 
> 2) the setting should be in Auxiliary DB tab since it already has saves related settings
> image
> 
> 
> 

I kind of rushed it back then. Yeah, having more of those configs would be better. I’ll add them in the next couple of days.

PR #1410, opened 2025-10-19
Fix crash when opening path that doesn't exist

Patryk-Malinowski:
image


LionelColaso:
@Patryk-Malinowski reverted this in https://github.com/RimSort/RimSort/pull/1427 since it causes breaks steam specific open see https://github.com/RimSort/RimSort/issues/1423

Patryk-Malinowski:
No worries, I'll find a better approach whenever I get the time.

PR #1583, opened 2025-12-10
game executable override

simatt99:
create a user option to override folder selection for the games executable. This helps if the user is running an unconventional setup such as a windows EXE on MacOS.


simatt99:
Further issues found, need to investigate before reopening

LionelColaso:
Just add a simple checkbox to avoid check

simatt99:
> Just add a simple checkbox to avoid check

Will add the checkbox, I also need to fix the OS detection. While the path gets overridden, when the game tries to launch, it still looks for a .app instead of looking for a windows EXE. I forgot about this part when making my fix

PR #1620, opened 2025-12-20
Check if Steam is running before launching with Steam integration

### LINKED ISSUE #1376: Check steam is running when steam integration is enabled
### Origin of idea

_No response_

If attempting to run RimWorld through RimSort, first check if steam is running if user has enabled steam integration.

### Alternatives considered

_No response_

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

cebarks:
Adds validation to verify Steam is running when Steam integration is enabled before launching RimWorld. Shows a warning dialog if Steam is not detected, allowing users to launch anyway or cancel.

- Add is_steam_running() utility function for cross-platform Steam process detection using psutil
- Check Steam status in _do_run_game() when steam_client_integration is enabled
- Show warning dialog with "Launch Anyway" and "Cancel" options
- Add comprehensive unit tests with 100% coverage
- Fail-open design prevents blocking users on detection errors

Closes #1376

Disclaimer: Claude Code was used to generate SOME of the code.  HOWEVER, all changes and implementations were eviewed and double checked by an experienced human for accuracy.

### DIFF
--- a/app/utils/generic.py
+++ b/app/utils/generic.py
@@ -462,6 +462,60 @@ def validate_game_executable(game_folder: str) -> bool:
     return False
 
 
+def is_steam_running() -> bool:
+    """
+    Check if the Steam client is currently running.
+
+    Uses psutil to enumerate processes and looks for platform-specific
+    Steam process names. This is used to validate that Steam is running
+    before launching RimWorld with Steam integration enabled.
+
+    Platform-specific process names:
+    - Windows: steam.exe
+    - Linux: steam or steamwebhelper
+    - macOS: Steam or steamwebhelper
+
+    :return: True if Steam is detected running, False otherwise
+    """
+    try:
+        import psutil
+
+        system_name = platform.system()
+
+        # Define platform-specific Steam process names
+        if system_name == "Windows":
+            steam_process_names = {"steam.exe"}
+        elif system_name == "Linux":
+            steam_process_names = {"steam", "steamwebhelper"}
+        elif system_name == "Darwin":  # macOS
+            steam_process_names = {"Steam", "steamwebhelper"}
+        else:
+            logger.warning(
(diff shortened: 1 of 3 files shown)


LionelColaso:
why not just add a notification here https://github.com/RimSort/RimSort/blob/main/app/utils/steam/steamworks/wrapper.py#L68

cebarks:
I was mainly trying to avoid adding alerts to a util file and wanted to keep it in the main view.  While diving deeper into this trying to use the SteamworksInterface instead of the process detection maybe we should create a global `SteamworksInterface` for the app instead of creating them adhoc and discarding them as needed like we currently do?

LionelColaso:
> I was mainly trying to avoid adding alerts to a util file and wanted to keep it in the main view. While diving deeper into this trying to use the SteamworksInterface instead of the process detection maybe we should create a global `SteamworksInterface` for the app instead of creating them adhoc and discarding them as needed like we currently do?

I think this should be closed and all changes should be in #1623 unless you want both separate

LionelColaso:
> While diving deeper into this trying to use the SteamworksInterface instead of the process detection maybe we should create a global `SteamworksInterface` for the app instead of creating them adhoc and discarding them as needed like we currently do?

yeah sure but then would it be a separate module or a utility ?

cebarks:
> > I was mainly trying to avoid adding alerts to a util file and wanted to keep it in the main view. While diving deeper into this trying to use the SteamworksInterface instead of the process detection maybe we should create a global `SteamworksInterface` for the app instead of creating them adhoc and discarding them as needed like we currently do?
> 
> I think this should be closed and all changes should be in #1623 unless you want both separate

Agreed, I'll move these changes into #1623

PR #1593, opened 2025-12-13
expose DB builder logic via `RimSort build-db` subcommand

cebarks:
This pull request adds a CLI mode, currently with the only sub-command being `db-builder`. It runs in headless mode without initializing the QT GUI stack. 

This should make it much easier to automate the updating of [Steam-Workshop-Database](https://github.com/RimSort/Steam-Workshop-Database).

#1592

```
Usage: python -m app build-db [OPTIONS]

  Build Steam Workshop metadata database.

  This command queries the Steam WebAPI to build a comprehensive database of
  mod metadata for the specified AppID. The database includes mod names, URLs,
  dependencies, and optional DLC requirements.

  The API key can be provided in three ways (in priority order):

  1. --api-key command line argument
  2. RIMSORT_STEAM_API_KEY environment variable
  3. Fallback to settings.json (if RimSort GUI is configured)

  Examples:

    # Using environment variable (recommended for security)
    export RIMSORT_STEAM_API_KEY=your_32_character_key_here
    rimsort build-db --output workshop.json

    # Using command line argument
    rimsort build-db --api-key ABC123... --output workshop.json

    # Update existing database with new data
    rimsort build-db --output workshop.json --update

    # Skip DLC data for faster builds
    rimsort build-db --output workshop.json --no-dlc-data --quiet

Options:
  --api-key TEXT              Steam WebAPI key (32 characters). Can also be
                              set via RIMSORT_STEAM_API_KEY environment
                              variable.
  --output PATH               Output JSON file path for the database.
                              [required]
  --appid INTEGER             Steam AppID (294100 is RimWorld).  [default:
                              294100]
  --expiry INTEGER            Database expiry time in seconds (0 = no expiry).
                              [default: 0]
  --dlc-data / --no-dlc-data  Include DLC dependency data (requires Steamworks
                              API).  [default: dlc-data]
  --update / --overwrite      Update existing database (merge) or overwrite
                              completely.  [default: overwrite]
  --quiet                     Suppress progress output (errors still shown).
  --help                      Show this message and exit.
```

Disclaimer: Claude Code was used to generate SOME of the code and documentation. HOWEVER, all changes and implementations were reviewed and double checked by an experienced human for accuracy.

PR #1622, opened 2025-12-20
add a Justfile for common commands

cebarks:


LionelColaso:
@cebarks between we forgot about https://github.com/RimSort/RimSort/pull/1593

Issue #1587, opened 2025-12-11
bug: Mod update via Steam Client Integration doesn't call SteamAPI


SypherRed:
### Release Type

Compiled (Windows x86-64)

### Version

v1.0.61

### Operating System

Windows 11

### Relevant logs

https://0x0.st/KCuh.log

### Describe the bug

If updates for mods have been found, updating with the steam client doesn't call steam to download the workshop content as it was before, even tho steam is active/open.

There is no message, that updating via API doesn't work.

<img width="902" height="632" alt="Image" src="https://github.com/user-attachments/assets/32befc51-7e3e-41d1-a821-6491d37073cb" />

### Additional context

If updating via the SteamCMD, the mods will be updated, but a duplication of mods will happen as mentioned in [#1460 ](https://github.com/RimSort/RimSort/issues/1460)

<img width="902" height="632" alt="Image" src="https://github.com/user-attachments/assets/27444344-3c80-4a1b-b401-5c348e538a0e" />

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

SypherRed:
Added a screenshot after mods were updated via steamCMD.

Edit: After refreshing, Rimsort still presents me the mods, which I just updated via steamCMD (and deleted the duplicates of it) for being in need to be updated.

LionelColaso:
Was steam running when doing the updates unless steam changed the commands or was not running the mods should have been updated  
i will do some testing 
do you by any change have this issue https://github.com/RimSort/RimSort/issues/1460

```
INFO][2025-12-11 23:25:12][6788][Dummy-39][main_content_panel][_do_steamworks_api_call][2158] : Received Steamworks API instruction: ['resubscribe', [2023507013, 3525414162]]
[INFO][2025-12-11 23:25:12][6788][Dummy-39][main_content_panel][_do_steamworks_api_call][2186] : Creating Steamworks API process with instruction ['resubscribe', [2023507013, 3525414162]]
[DEBUG][2025-12-11 23:25:17][6788][Dummy-39][animations][run][149] : WorkThread completed, returning to main thread
```

SypherRed:
As stated in the issue, steam was and is running completely before I open RimSort.
And yes, issue #1460 is mentioned above too with a screenshot.

[...]

Issue #1460, opened 2025-11-04
Few seemingly random mods dissappearing every time after updating mods through Rimsort


JinJanODWan:
### Release Type

Compiled (Windows x86-64)

### Version

1.0.50 and 1.0.51

### Operating System

Win11

### Describe the bug

<img width="1291" height="690" alt="Image" src="https://github.com/user-attachments/assets/7334a736-66fa-48a0-806d-e24bba1deb74" />

### Additional context

Title happens, without fail, every single time I use Rimsort's "update mods" feature. Around three to five seemingly unrelated mods disappear from list and then I'm met with incorrect "duplicate mod" message (same amount of invalid mods that were just removed) as seen in screenshot.

Only workaround I have found is finding the missing mods in Workshop , unsubscribing, and resubscribing.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

Skydea777:
Same Issue, started again after Updating RimSort to the Current 1.0.51 release, I thought the Problem was only on my end, but I have dozens of seemingly "missing" mods.
Also, after Updating it resets the change time to Linux Base Time January 1, 1970.

Example of one of the Mods I already Resubscribed multiple times:

<img width="950" height="268" alt="Image" src="https://github.com/user-attachments/assets/c55587a8-2a2d-4d81-8368-d1f0a32ed58b" />


And the Issue of Non Updating Mods / Mods not getting recognized it was Updated:

<img width="1030" height="1392" alt="Image" src="https://github.com/user-attachments/assets/d6803a51-9a1e-4806-b4fd-eee701005e16" />


It is more then a bit aggravating.

Have it since Release of the new Version, installed it 1 hour after release.
Reinstalled it.
Restarted the Computer
Checked Rimworld Consistency and so on ~
And so on and so on.

JinJanODWan:
To add to above: my "outdated" mods have also incorrectly started showing 1970 as download date same as above. Also, the list has just gotten longer and longer with each restart of Rimsort.

Skydea777:
Yeah... its getting longer. I rechecked the consistency again and.. yeah, the others are also getting longer.
"Shit's fucked."

<img width="949" height="664" alt="Image" src="https://github.com/user-attachments/assets/fb1cb0b6-0fec-4973-8bb4-b66142c62cc3" />

[...]

PR #1623, opened 2025-12-21
refactor steam integration

cebarks:
This refactors the steamworks integration heavily:
- replaces `multiprocessing` steamworks api calls with a QT thread implementation
  - allows for the reuse of the `SteamworksInterface` instance across all usecases
  - simplifies control flow and removes the need for chunking and IPC.
  - allows for async queuing of steamworks operations so the UI doesn't block  for it to return.
- query steamworks api for mod installation/update timestamp instead of parsing possibly stale ACF data
- refactor entire download workflow. There is now a single queue (viewed via the new downloads tab)
  - mod downloads/updates are now triggered using the steamworks `DownloadItem()` call instead of unsubscribing and resubscribing which is much more consistent. (also fixes 
  - mod download status are now tracked via callbacks and polling mod status instead of watching and parsing .acf files

This fixes:
- #1460 
- #1587 
- #1376
- #1630

Additional misc changes:
- upstreamed rimsort specific Steamworkspy changes:
  - philippj/SteamworksPy/pull/115
  - philippj/SteamworksPy/pull/114
  - philippj/SteamworksPy/pull/116
- adds a "Clean Orphaned Workshop Items" steam troubleshooting utility.
  - removes mods from the steam workshop metadata .acf file that no longer exist locally. This can happen if a user deletes the mod on disc.

Disclaimer: Claude Code was used to generate SOME of the code. HOWEVER, all changes and implementations were reviewed and double checked by an experienced human for accuracy.


LionelColaso:
#1460 and #1587 are the only 2 big bugs which are related and fixing them would really be of great help
Thanks

cebarks:
I haven't been able to reproduce #1460 locally. Maybe if @Skydea777 or @JinJanODWan can confirm that this PR fixes it? If not it might be best to address #1460 in a follow u PR.

cebarks:
I want to at least look into the following things before marking this ready to merge: 
- getting this PR (https://github.com/philippj/SteamworksPy/pull/91 originally by @twstagg) upstreamed to remove the submodule dependency of the fork. this simplifies packaging and this major steam integration rework seems like the right time to have it happen.
- research if there's a better situation for forcing a mod to update than ubsubscribing/resubscribing
- #1460
- #1376 (should be addressed by [e22b57f](https://github.com/RimSort/RimSort/pull/1623/commits/e22b57f3b07174e8d17801924702d0c79b082278))
- #1587 (ensure this PR actually addresses this, which currently I BELIEVE it does, but have not confirmed)

cebarks:
I've resubmitted philippj/SteamworksPy/pull/91 as philippj/SteamworksPy/pull/114

cebarks:
This currently uses my branch of steamworkspy until https://github.com/philippj/SteamworksPy/pull/116 is merged, at which point we can swap over.

.gitmodules
```diff
@@ -4,6 +4,6 @@
 	ignore = dirty
 [submodule "SteamworksPy"]
 	path = submodules/SteamworksPy
-	url = https://github.com/twstagg/SteamworksPy
+	url = https://github.com/cebarks/SteamworksPy
 	ignore = dirty
-	branch = master
+	branch = duplicate-fix
```
oceancabbage: Intended to be on this branch? (we won't forget about this in the long term?)

cebarks:
updated to pull in features from #1637 but based on this SteamworksInterface architecture.

Also fixes steam launching on linux and autostart setting

nyabinary:
What the status of this?
cc: @LionelColaso

cebarks:
> What the status of this?

This PR ended up getting way out of control so I've been planning to implement it from the ground up in a series of smaller PRs to make merging and review MUCH MUCH easier. 

I haven't had the time to work on it this year so far but I have some time off in the coming weeks and I'll start taking a look at it.

cebarks:
Closing in favor of the broken up PRs

PR #1633, opened 2025-12-26
Allow horizontal resizing mod lists

oceancabbage:
Attempt at making the mod lists horizontally resizable. Not perfect because I can't figure out how to make the dropdown buttons truncate... they seem to have a min width. But, it should be better than right now. Would appreciate testing on PC and Linux.

https://github.com/user-attachments/assets/7df2b997-e535-41de-9ab2-438fa1336161

### DIFF
--- a/app/views/mods_panel.py
+++ b/app/views/mods_panel.py
@@ -51,6 +51,7 @@ from PySide6.QtWidgets import (
     QMessageBox,
     QProgressDialog,
     QPushButton,
+    QSplitter,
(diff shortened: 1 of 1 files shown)


oceancabbage:
I didn't do much testing past just resizing the mod lists, so hopefully this doesn't break some workflow elsewhere.

cebarks:
I've also been wanting this forever

LionelColaso (approved):
Cool nice

PR #1649, opened 2025-12-29
Steam workshop database refactor: incremental steamDB build support

cebarks:
This depends on #1623 for it's better QThreadPool based steam webapi implementation and SteamworksInterface handling (via singleton model)

# Summary
Migrates the Steam Workshop database builder from synchronous requests + multiprocessing to a QThreadPool based implementation for better concurrency for CLI usage. This targets 6-20x performance improvements for bulk operations (steamdb build).

The database builder now supports incremental updates that dramatically reduce build times for routine updates by only querying mods that have changed since the last build. Right now that only works via the CLI command `build-db` (enabled by default)

Together, these changes drastically reduce the time for both a full rebuild (30+ minutes -> ~ <UNTESTED>) and a db `--update` (25+ minutes -> 

## Async changes
  - AsyncWebAPIClient - Context manager for aiohttp session lifecycle with connection pooling
  - _retry_with_backoff_async() - Async retry logic with exponential backoff (preserves existing behavior)
  - get_bulk_timestamps_async() - Concurrent timestamp queries (213-mod chunks)
  - IPublishedFileService_GetDetails_async() - Concurrent metadata queries
  - create_steam_db_async() - Async database builder orchestration

## steamDB builder refactor
  - save `created_at` and `updated_at` workshop item fields in database
  - query only time_updated and result fields (minimal payload) to quickly determine which mods need full metadata queries
  - "soft deletes" mods (adds `deleted` and `deleted_at` fields to the db) when they are removed from the workshop instead of removing from the database. 
  
Disclaimer: Claude Code was used to generate SOME of the code. HOWEVER, all changes and implementations were reviewed and double checked by an experienced human for accuracy.


cebarks:
## Proposed Enhancement: Two-Phase Dependency Resolution

While reviewing the async refactor, I noticed an opportunity to improve the dependency resolution workflow that would complement these changes nicely.

### Current Limitation

The existing `create_steam_db()` implementation mixes data fetching with dependency linking in a single pass, requiring two query rounds:
1. Query initial mods + try to link dependencies immediately
2. If a child dependency isn't cached yet, add to `missing_children` list
3. Query `missing_children` again

**Problem:** Deep dependency chains (A→B→C→D) are truncated at 2 levels. We only fetch children and grandchildren, stopping there.

### Proposed Two-Phase Approach

**Phase 1 (Fetch):** Iteratively discover and fetch all metadata with full API fields
- Query initial mods → collect child pfids → queue for fetching
- Query children → collect *their* children → queue for fetching  
- Repeat until no new children discovered (with max iteration safety limit)
- Store raw metadata only, no linking yet

**Phase 2 (Link):** Pure in-memory dependency linking
- All data already fetched, so linking always succeeds
- Clean separation of concerns

### Benefits

✅ **Handles arbitrary dependency depth** (not limited to 2 levels)  
✅ **Cleaner architecture** (separation of fetching vs. linking)  
✅ **Fixes 6 bugs:**
- List mutation during iteration (lines 617-623)
- No centralized API key redaction
- No iteration safety limit (infinite loop risk)
- Temporary `_children` data in final database
- Missing `import traceback`
- Inconsistent redaction patterns

✅ **Same API call count** for typical 2-level dependencies  
✅ **Zero breaking changes** - identical database output format  
✅ **Complements async refactor** - same query patterns, just better orchestration

The async improvements in this PR would make the iterative fetching even more performant since we could parallelize discovery rounds better with the QThreadPool approach.

PR #1694, opened 2026-01-10
Set toggle warning as checked when active. Change behavior of toggling multiple items.

### LINKED ISSUE #595: Suggestion: Change Toggle Warning to a checkable QAction
Currently the 'Toggle Warning' button is a default QAction.
image

It should be changed to a checkable QAction with changing icon.

Patryk-Malinowski:
image
image

Now if toggling warnings for multiple items it will enable it for all unless all of them are already enabled, then it will disable all. To me this seems more intuitive.

Closes #595


LionelColaso (approved):
Looks Good

PR #1756, opened 2026-02-07
Fix bulk mod coloring. Speed up by ~90%

### LINKED ISSUE #1748: Mod colors reset when changing modlist
### Release Type

Compiled (Windows x86-64)

### Version

1.0.69

### Operating System

Windows 11

### Relevant logs

_No response_

### Describe the bug

Sometimes when I change the colors of mods after moving them between lists, the colors will reset to whatever it was before I moved it. Example: The mods in my Active list were all different colors, then I changed them to be all green. I saved that as a modlist and put some new mods into the Active list. Instead of staying green, the colored mods reverted back to their previous colors before I changed them. Kind of annoying when troubleshooting mod crashes.

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

Patryk-Malinowski:
Previously if you did a select all and colored the mods it would only update the aux DB with mods that were already loaded... (ModListItemInner)
That is fixed now and also massively sped up bulk mod coloring/resetting.

Also added confirmation dialogs to reset all mod colors or toggled warnings from menu bar.

(Also going to fix issue with .msi build failing)

Closes #1748


Patryk-Malinowski:
I can remove wix changes if needed

PR #1792, opened 2026-02-21
Added Dividers feature to the mod list

Ketaros:
**Summary**
Adds the ability to insert named, collapsible dividers into the Active mod list, allowing users to visually organize their mods into sections.

**Features**
- Add dividers via right-click context menu ("Add divider here") on any item in the active mod list
- Rename dividers via right-click context menu on a divider
- Delete dividers via right-click context menu on a divider
- Collapse/expand sections by clicking the arrow on a divider or double-clicking it — hides/shows all mods between that divider and the next
- Drag and drop dividers to reposition them within the active mod list
- Mod count display on each divider showing how many mods are in its section
- Persistence — dividers (name, position, collapsed state) are saved in settings.json and restored on app restart
- Sort-safe — when Sort is clicked, all mods are sorted normally and dividers are moved to the bottom of the list (expanded) so the user can reposition them

**New Files**
- app/models/divider.py — DividerData model class with UUID prefix helpers
- app/views/divider_widget.py — DividerItemInner widget (arrow, name label, mod count, horizontal line)
- Modified Files
- app/models/settings.py — new active_mods_dividers field for persistence
- app/views/mods_panel.py — divider lifecycle (add, remove, rename, toggle, collapse, restore), context menus, guards to skip dividers in mod-specific logic (warnings, filtering, counting, search)
- app/views/main_content_panel.py — filter dividers from save/export/sort/clear operations, restore dividers after list rebuilds
- app/controllers/main_window_controller.py — exclude dividers from dependency checks and unsaved-changes detection
- app/controllers/file_search_controller.py — exclude dividers from file search scope

**Design Decisions**
Dividers use a __divider__ UUID prefix to distinguish them from real mods without changing existing data structures
Dividers are never written to ModsConfig.xml — they are UI-only and stored separately in settings.json
All existing mod iteration loops are guarded with is_divider_uuid() checks to prevent KeyError crashes when accessing internal_local_metadata

**Test Plan**
[ ] Add a divider via right-click, verify it appears with the entered name
[ ] Rename and delete dividers via right-click context menu
[ ] Collapse/expand a divider by clicking the arrow — mods below should hide/show
[ ] Drag a divider to a new position
[ ] Click Save, restart the app — dividers should be restored at their positions
[ ] Click Sort — mods should sort, dividers should move to the bottom
[ ] Click Clear — dividers should be removed
[ ] Export mod list — verify dividers are not included in ModsConfig.xml
[ ] Search/filter mods — dividers should not interfere with results


LionelColaso:
@Ketaros would you like to update and resolve conflicts

LionelColaso (approved):

PR #1806, opened 2026-03-01
Delete mods using del key

Patryk-Malinowski:
Fixed checking for .dds textures when enabling/disabling relevant action in deletion sub menu. (I think mods will always have `Textures` dir if adding/overwriting textures...)

Allows user to use del key to bring up the deletion menu.

This assumes the deletion menu correctly handles a list that contains a mix of mods that qualify for some options but not others etc.

### DIFF
--- a/app/views/deletion_menu.py
+++ b/app/views/deletion_menu.py
@@ -1,5 +1,4 @@
 import errno
-import os
 from pathlib import Path
 from shutil import rmtree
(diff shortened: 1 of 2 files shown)


Patryk-Malinowski:
Think theres an issue related to this but cannot find it

LionelColaso (approved):

PR #1837, opened 2026-03-17
Replace eval() with int() for publishedfileid conversion

cebarks:
## Summary

- Replace 6 `eval(str_pfid)` calls with `int(str_pfid)` across 4 files
- `eval()` on external data (Steam API responses, Rentry imports, JS bridge inputs) is an arbitrary code execution risk
- All call sites only need string-to-integer conversion — `int()` is the correct function

## Security Impact

**Severity: Critical** — The JS bridge path (`steambrowser/browser.py`) accepts strings from JavaScript in the embedded Steam browser. A malicious page could pass arbitrary Python code as a "publishedfileid."

## Test plan

- [x] Verify mod subscription/unsubscription still works via Steamworks API
- [x] Verify Rentry import still downloads mods correctly
- [x] Verify Steam browser download list still functions

### DIFF
--- a/app/utils/steam/steambrowser/browser.py
+++ b/app/utils/steam/steambrowser/browser.py
@@ -437,7 +437,7 @@ class SteamBrowser(QWidget):
         EventBus().do_steamworks_api_call.emit(
             [
                 "subscribe",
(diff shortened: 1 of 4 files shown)


LionelColaso (approved):
Looks Good Thanks

PR #1839, opened 2026-03-17
Harden subprocess execution

cebarks:
## Summary

- Remove unnecessary `shell=True` from Windows game launch — list-based arguments don't need shell interpretation
- Replace f-string + `shell=True` with list args for MSI installer launch — `msiexec` is an external binary, not a shell builtin
- Use `tempfile.mkstemp()` instead of predictable `/tmp/update.sh` path for elevated update scripts, tighten permissions to `0o700`

## Security Impact

**Severity: High** — `shell=True` with list args on Windows passes through `cmd.exe` which can interpret metacharacters. Predictable temp paths enable symlink/TOCTOU attacks on shared systems.

## Known limitation

`.bat`/`.cmd` wrapper commands on Windows will no longer work without `shell=True`. This is an unlikely edge case (wrappers are Linux-centric) noted for future enhancement with a `cmd /c` guard.

## Test plan

- [ ] Verify game launch works on Windows without `shell=True`
- [ ] Verify MSI update works on Windows with list args
- [ ] Verify elevated update script launch works on Linux with randomized temp path

### DIFF
--- a/app/utils/generic.py
+++ b/app/utils/generic.py
@@ -493,7 +493,6 @@ def launch_process(
             p = subprocess.Popen(
                 popen_args,
                 creationflags=subprocess.CREATE_NEW_PROCESS_GROUP,
-                shell=True,
                 cwd=cwd,
                 env=env,
             )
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Looks Good Thanks

PR #1918, opened 2026-04-23
Fix: Add hasattr check to prevent AttributeError when calling DownloadItem in Steamworks API

### LINKED ISSUE #1800: Update Mods with Steam not working
### Release Type

Compiled (Windows x86-64)

### Version

v1.0.72

### Operating System

Windows 11

### Relevant logs

https://0x0.st/PQvI.log

### Describe the bug

In "Updates found for Workshop mods" window pressing "Update Mods with Steam" doesn't work any longer. In Steam nothing is queued in Downloads, instead RimWorld game starts to think it is running, but RimWorld isn't actually running. When trying to stop it from Steam RimSort closes instead. 
RimSort log also gives [ERROR][2026-02-26 20:59:15][2720][Dummy-12][wrapper][_handle_resubscribe][517] : Failed to trigger download for 3112549163: 'SteamWorkshop' object has no attribute 'DownloadItem' for every mod.

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

SpeedoMan25:
Fixes #1800

### Description
This PR resolves two connected issues that were causing Steam Workshop mod updates to crash and get stuck in the download queue.

**1. Fixed the `AttributeError` crash during downloads**
The updater was trying to call a Steam command (`DownloadItem`) that is missing from the current version of the Steam library, which caused the thread to crash. I added a `hasattr()` check to verify the command exists before running it. If it's missing, it now safely skips the forced download, allowing the rest of the unsubscribe/resubscribe queue to finish without throwing Python errors.

**2. Fixed the "hanging" Steam downloads**
Previously, Steam wouldn't actually download the queued mods until RimSort was fully closed. This happened because the Steam update handler was executing inside the main application process (using `.run()`). Because the Steam API was bound to the main app, it tricked Steam into thinking RimWorld was actively running the entire time RimSort was open. Steam intentionally pauses Workshop downloads while a game is running. 

I updated the execution to use `.start()` and `.join()` so the Steam handler runs in a true, isolated background process. Now, once the mod queue is updated, the background process properly terminates, Steam realizes the "game" has closed, and the downloads start immediately!

### DIFF
--- a/app/utils/steam/steamworks/wrapper.py
+++ b/app/utils/steam/steamworks/wrapper.py
@@ -465,12 +465,15 @@ class SteamworksSubscriptionHandler(Process):
                 # Call DownloadItem to force Steam to queue the mod for download
                 # high_priority=True makes Steam skip other queued downloads
                 # Note: callback may not fire, but the download is still queued
-                steamworks_interface.steamworks.Workshop.DownloadItem(
-                    pfid,
-                    high_priority=True,
-                    callback=self._create_download_callback(steamworks_interface),
-                    override_callback=True,
-                )
+                if hasattr(steamworks_interface.steamworks.Workshop, "DownloadItem"):
+                    steamworks_interface.steamworks.Workshop.DownloadItem(
+                        pfid,
+                        high_priority=True,
+                        callback=self._create_download_callback(steamworks_interface),
+                        override_callback=True,
+                    )
+                else:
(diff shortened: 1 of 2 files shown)


LionelColaso:
@SpeedoMan25 Thanks a lot nice catch

LionelColaso (approved):

PR #1927, opened 2026-04-27
feat(browser): open mod URL on downloader list item double-click

exefer:
Double-clicking an item in the Steam Browser mod downloader list now opens its mod page, as shown in the item tooltip, improving navigation.


LionelColaso (approved):
Looks Good Thanks

PR #1928, opened 2026-04-27
chore(build): pin Nuitka + slim bundle (remove imageio, trim imports)

exefer:
- Pin Nuitka to 4.0.8 for stable builds
- Remove unused imageio
- Exclude extra SQLAlchemy dialects from bundle
- Enable no_asserts/no_docstrings and skip Qt translations

Reduces build size and improves consistency


LionelColaso (approved):
Looks Good Thanks

PR #1935, opened 2026-05-04
fix(git): prevent segfault crash on pygit2 fetch timeouts

### LINKED ISSUE #1932: Crashing after 1 minute of working. No fatal error in log
### Release Type

Compiled (Windows x86-64)

### Version

v1.0.75

### Operating System

Windows 10

### Relevant logs

[RimSort.log](https://github.com/user-attachments/files/27315497/RimSort.log)

### Describe the bug

Launcher is constantly crashing after 1 minute of working. Nothing that i can do seems to be particulary causing this problem bc even if do nothing its gonna close anyways after 1 minute or so. Cant update launcher thru itself :(
There are no big and scary fatal errors inside the log

### Additional context

Another log with the same problem 
[RimSort.old.log](https://github.com/user-attachments/files/27315678/RimSort.old.log)

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

SpeedoMan25:
Fixes #1932

### Description
This PR resolves a silent crash (segmentation fault) that caused the launcher to abruptly terminate without an error dialog or traceback exactly one minute after starting. This occurred specifically on slower network connections during background database updates.

**1. Prevented Segfaults on pygit2 Fetch Timeouts**
Previously, when a background git database fetch exceeded its time limit, `_fetch_with_timeout()` correctly threw a timeout warning, but the C-level `pygit2.Remote.fetch()` daemon thread was left running in the background. Shortly after, the application correctly attempted to clean up the git operation by calling `repo.free()`, which forcibly released the underlying C structs from memory. When the hanging daemon thread eventually tried to interact with that manually freed memory, it caused a fatal Segmentation Fault. To resolve this, a `_has_hanging_threads` flag is now set on the repository object whenever a timeout occurs. The `git_cleanup()` method checks for this flag and bypasses the explicit `repo.free()` call if a thread is still active, safely offloading memory cleanup to Python's garbage collector and keeping the structures intact until the daemon thread finishes.

### DIFF
--- a/app/utils/git_utils.py
+++ b/app/utils/git_utils.py
@@ -119,10 +119,11 @@ class GitOperationConfig:
         return cls(fetch_timeout=fetch_timeout, connection_timeout=connection_timeout)
 
 
-def _fetch_with_timeout(remote: pygit2.Remote, timeout: int) -> bool:
+def _fetch_with_timeout(repo: Repository, remote: pygit2.Remote, timeout: int) -> bool:
     """Fetch from remote with timeout handling.
 
     Args:
+        repo: The git repository object.
         remote: The pygit2 remote object.
         timeout: Timeout in seconds.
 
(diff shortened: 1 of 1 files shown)


LionelColaso (approved):
Looks Good Thanks

Issue #1630, opened 2025-12-24
Can't build Steam Database + RimSort Crashes


Lorsieab2:
### Release Type

Compiled (Windows x86-64)

### Version

1.0.63

### Operating System

Windows 11

### Relevant logs

https://0x0.st/PzIH.old.log

### Describe the bug

I'm trying to build the Steam Workshop database. When I get the PublishedFileIDs from the Workshop, RimSort crashes and there's a 503 Service Unavailable error in the log. https://0x0.st/PzIH.old.log

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

cebarks:
Is this still persisting? i just ran a full webapi db build and it worked fine.

Nonetheless, RimSort should be failing gracefully instead of crashing if it fails...

LionelColaso:
@Lorsieab2 as per the error

```
[INFO][2025-12-24 16:53:55][7660][Dummy-22][dialogue][show_fatal_error][269] : Showing fatal error box with input: [RimSort crashed], [The RimSort application crashed! Sorry for the inconvenience!], [Please contact us on the Discord/Github to report the issue.], [Traceback (most recent call last):
  File "C:\PROGRA~2\RimSort\app\utils\metadata.py", line 2737, in run
  File "C:\PROGRA~2\RimSort\app\utils\db_builder_core.py", line 86, in run
  File "C:\PROGRA~2\RimSort\app\utils\steam\webapi\wrapper.py", line 438, in pfids_by_appid
  File "C:\PROGRA~2\RimSort\app\utils\steam\webapi\wrapper.py", line 640, in IPublishedFileService_QueryFiles
  File "C:\PROGRA~2\RimSort\steam\webapi.py", line 167, in call
  File "C:\PROGRA~2\RimSort\steam\webapi.py", line 327, in __call__
  File "C:\PROGRA~2\RimSort\steam\webapi.py", line 431, in webapi_request
  File "C:\PROGRA~2\RimSort\requests\models.py", line 1026, in raise_for_status
requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url
]
```

it could be a temp issue, try again if it is not fixed see here to fix on your side https://www.geeksforgeeks.org/blogs/http-error-503-service-unavailable-fix/

cykonetic:
[...]

PR #1938, opened 2026-05-07
fix(db-builder): add new required Steam WebAPI QueryFiles parameters

cebarks:
We were missing semi-recently added query parameters on the webapi. This should fix the issue where the steam-db builder wasn't working properly (#1630)


LionelColaso (approved):
Looks Good Thanks

PR #1939, opened 2026-05-07
Attempt at producing an AppImage

### LINKED ISSUE #35: Suggestion: Add RimSort to package managers
Consider adding RimSort to package managers as to make it easier for any users to keep it up to date. While primarily relevant to Linux systems, I would personally also suggest [Scoop](https://scoop.sh/) and/or [Chocolatey](chocolatey.org/) for Windows users.

### LINKED ISSUE #1930: Distribute the Linux version as an AppImage instead of a hard compiled Ubuntu package
### Origin of idea

Hey,

So I am using Debian 13, and RimSort does not work on this distribution, at least not the Ubuntu 24.04 release.
That's likely to be a missing dependency issue. The logs are pointing out QT with a file that is unknown to `apt-file` and also a paths issue. This last one I checked on this Github and I found quite nebulous and obscure explanations which never really provided a solution.

This led me to believe that many of those problems could be solved by distributing this software as an AppImage instead of a raw Ubuntu compiled software.

AppImage are just like .deb and alike, but they are distribution agnostic, and will automatically track dependencies with the distribution package manager. This would massively improve your developer experience by just having one single Linux distribution to maintain.

### Alternatives considered

An alternative, which is what most people developing for a wide variety of OS do, would be to build not for Ubuntu, but for each base distribution. This means building for Debian, Fedora, RHEL, etc.
This is because everything that works on Debian (for example) will work on Ubuntu, and if it does not, then it's a bug. The opposite is not true, not everything that works on Ubuntu will work on Debian. Considering what RimSort does, you are very unlikely to need any Ubuntu specific features.

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

cebarks:
Create an AppImage distribution for linux that should be distribution agnostic.

I referenced this script by @usagirei (https://github.com/RimSort/RimSort/issues/35#issuecomment-3233693419) with some help from Claude to adapt it to the github CI.

I was able to successfully build it locally in a fedora 44 aarch64 so far.

Relates to #35

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -257,6 +257,10 @@ jobs:
         id: find_appimage
         run: |
           APPIMAGE=$(find build/ -name "RimSort-*.AppImage" | head -n 1)
+          if [[ -z "$APPIMAGE" ]]; then
+            echo "ERROR: No AppImage found in build/" >&2
+            exit 1
+          fi
           echo "APPIMAGE=$APPIMAGE" >> "$GITHUB_OUTPUT"
           echo "AppImage found at $APPIMAGE"
         shell: bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -245,7 +245,7 @@ jobs:
           subject-path: ./RimSort-${{ steps.sem_version.outputs.version }}-${{ matrix.platform }}-${{ matrix.arch }}.msi
 
       - name: Build AppImage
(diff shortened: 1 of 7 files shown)


LionelColaso:
@cebarks check https://github.com/RimSort/RimSort/issues/35 may help

LionelColaso (approved):
Looks Good and worked Thanks

PR #1944, opened 2026-05-10
Optimize artifact sizes and various small CI fixes

cebarks:
## Summary

- Reduce macOS release artifacts from ~700 MB to ~300 MB (~57% reduction)
- Reduce Linux release artifacts from ~258 MB to ~234 MB (~9% reduction)
- Reduce Windows release artifacts by ~9-12%
- Switch macOS/Linux release format from `.zip` to `.tar.gz` for better compression
- Fix release workflow to gzip existing build tars instead of extract→re-archive
- Disable Fedora/RPM builds (broken pre-existing — see below)

### Changes

**Nuitka exclusions** (`app/__main__.py`, `build.yml`):
- Exclude `numpy` (transitive optional dep of networkx, never imported at runtime, ~7 MB)
- Exclude `qtwebengine_devtools_resources.pak` (Chrome DevTools data, ~33 MB across 3 copies in macOS bundle)

**Post-build macOS optimization** (`packaging/optimize_macos_bundle.py`):
- Thins universal (x86_64 + arm64) fat Mach-O binaries in `.app` bundle's `Contents/Resources/` to the target architecture (~267 MB savings). PySide6 ships universal frameworks on PyPI, but Nuitka already extracts thin single-arch dylibs into `Contents/MacOS/` for runtime — the fat copies in Resources exist only for code signing structure
- Deduplicates data files in framework `Versions/A/Resources/` directories by replacing them with symlinks to the top-level `Resources/` copy (~70 MB savings from qtwebengine_locales, icudtl.dat, etc.)
- Removes `.ts` translation source files from locales/, keeping only compiled `.qm` files (~2 MB)
- Re-signs the bundle after all modifications

**Build/release pipeline improvements** (`build.yml`, `release.yml`):
- Build step now tars with the correct root directory name (`RimSort`/`RimSort.app`) instead of `output`
- Release workflow gzips the existing tar for macOS/Linux instead of extracting and re-archiving
- Commit match check extracts only `version.xml` via stdout instead of full archive extraction
- Switches macOS and Linux release archives from `.zip` to `.tar.gz` for ~36% better compression ratio
- Windows stays as `.zip`/`.msi` for native support
- Fixed `BUILD_RPM` boolean vs string comparison (YAML `true` != string `'true'`) that silently skipped all RPM steps
- Fixed release cleanup glob to preserve `.rpm` files

**Fedora/RPM builds disabled:**
- Discovered that Fedora matrix entries never actually ran inside their declared containers — `container:` was set as a matrix property but never applied at the job level, so builds always ran on Ubuntu
- RPM build steps (`dnf install`, `rpmbuild`) fail on Ubuntu since `dnf` is unavailable
- This was masked by the `BUILD_RPM` boolean bug causing all RPM steps to be skipped
- Disabled until properly addressed in a follow-up PR; AppImage support will cover Linux distribution

### CI-verified artifact sizes

| Platform | Before (v1.1.0) | After (this PR) | Savings |
|---|---|---|---|
| macOS ARM | 695 MB | **291 MB** | **58%** |
| macOS x86 | 718 MB | **323 MB** | **55%** |
| Linux (each) | 258 MB | **234 MB** | **9%** |
| Windows (.zip) | 205 MB | **187 MB** | **9%** |
| Windows (.msi) | 171 MB | **150 MB** | **12%** |
| **Total** | **2,305 MB** | **1,636 MB** | **29%** |

## Test plan

- [x] Optimization script tested on local macOS arm64 build — 267 MB saved from fat binary thinning, 70 MB from deduplication, 2 MB from .ts removal
- [x] Verified optimized app bundle launches and runs correctly
- [x] Verified both arm64 and x86_64 thinning work correctly
- [x] Linted and type-checked all changed files
- [x] Existing test suite passes (262 passed, 1 pre-existing failure unrelated to this PR)
- [x] No fat binaries remain in Resources/ after optimization
- [x] CI builds pass on all 5 platforms (macOS ARM, macOS x86, Ubuntu 22.04, Ubuntu 24.04, Windows)
- [x] CI post-build smoke tests pass on all platforms

Relates to #35

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -180,12 +180,25 @@ jobs:
           include-data-files: |
             version.xml=version.xml
           nofollow-import-to: >-
+            numpy,
             sqlalchemy.dialects.mssql,
             sqlalchemy.dialects.mysql,
             sqlalchemy.dialects.oracle,
             sqlalchemy.dialects.postgresql
         env:
           PYTHONPATH: ${{ steps.pythonpath.outputs.PYTHONPATH }}
+
+      - name: Optimize macOS bundle
+        if: runner.os == 'macOS'
+        shell: bash
+        run: |
+          APP_BUNDLE=$(find build/ -name "*.app" -maxdepth 2 | head -n 1)
+          if [ -n "$APP_BUNDLE" ]; then
+            echo "Optimizing $APP_BUNDLE..."
+            python packaging/optimize_macos_bundle.py "$APP_BUNDLE"
+          else
+            echo "No .app bundle found, skipping optimization"
+          fi
 
       - name: Set FILENAME
         id: filename
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -260,7 +260,8 @@ jobs:
       - name: Rename new build
         run: |
           cd build
-          if [[ "${{ matrix.platform }}" == *"Darwin"* ]]; then
+          platform="${{ matrix.platform }}"
+          if [[ "$platform" == *"Darwin"* ]]; then
             rootname="RimSort.app"
           else
             rootname="RimSort"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,22 +65,25 @@ jobs:
             env:
               BUILD_OUTPUT: "app.dist"
               Executable: RimSort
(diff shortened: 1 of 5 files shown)


LionelColaso (approved):
Looks Good Thanks

PR #1968, opened 2026-05-20
fix(ci): check out PR branch directly in lint workflow

cebarks:
## Summary

- Fix `git-auto-commit-action` failure in the lint job when super-linter reformats files on PR branches

## Root Cause

The lint job's `actions/checkout` defaults to checking out the PR merge commit (`refs/pull/<num>/merge`) in a **detached HEAD** state. When super-linter fixes files in-place (e.g., ruff reformatting), `git-auto-commit-action` tries to `git checkout` the actual PR branch to commit the fixes, but fails because the locally modified files conflict with the checkout.

## Fix

Add `ref: ${{ github.head_ref || github.ref }}` to the checkout step so it checks out the actual PR branch directly. This way `git-auto-commit-action` can simply stage, commit, and push without needing to switch branches.

- `github.head_ref` resolves to the PR's source branch on `pull_request` events
- `github.ref` is the fallback for `push` and `merge_group` events

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,6 +19,7 @@ jobs:
       packages: read
       # To report GitHub Actions status checks
       statuses: write
+      checks: write
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


LionelColaso (approved):
Looks Good Thanks

PR #1970, opened 2026-05-20
Mod tags feature: implement user-defined mod tagging system

VA-Wizard:
### Summary

This PR adds a user-defined tagging system for mods in the main mod list UI. It allows users to assign custom tags such as `framework`, `qol`, `translation`, etc. to mods, display those tags when needed, and filter mod lists by one or more tags.

The goal is to improve navigation and organization for large mod lists where load order matters and folder-like grouping is not practical.

#### User-defined mod tags

Added support for storing custom tags in the auxiliary metadata database.

Users can now manage tags from the mod list context menu:

- **Add new tags...**
- **Replace all tags...**
- **Remove all tags**

Tags are stored per mod and persist across refreshes/restarts.

#### Tag edit dialog

Added a new tag editing dialog that allows users to:

- select from existing tags using checkboxes;
- create new tags by typing comma-separated values;
- reuse existing tags without retyping them every time;
- select all / select none in the tag selection list.

This makes repeated tagging of many mods much faster and less error-prone.

#### Tag filtering UI

Added a tag filter control to both active and inactive mod lists.

The filter supports:

- selecting multiple tags;
- selecting all tags;
- selecting no tags;
- filtering by mods with no tags via the special **No tags** option;
- showing mods that match any selected tag.

This is especially useful for incrementally tagging a large mod list: users can filter by **No tags**, tag a batch of mods, and continue until everything is categorized.

#### Optional tag display in mod rows

Tags can be displayed next to mod names when the tag filter is enabled.

To avoid clutter and layout issues:

- tags are hidden by default;
- tags are shown only while tag filtering is active;
- long tag strings are elided;
- full tag information remains available through tooltips.

#### Tag cleanup

Unused tags are cleaned up when they are no longer assigned to any mod, preventing the tag list from growing indefinitely after tags are removed or replaced.

#### Theme styling

Added styling for the new tag edit dialog so it matches existing dark themes.

The dialog now has themed styling for:

- labels;
- input fields;
- tag list;
- checkbox indicators;
- buttons;
- dialog background/borders.

Theme updates include support for the Modern and RimPy styles.

#### Test compatibility

Adjusted tag loading behavior so UI initialization remains safe in tests that use simplified/dummy settings objects without an auxiliary database path.

### User-facing Behavior

Users can now right-click one or more mods and assign tags through the context menu. Tags can then be used to filter active or inactive mod lists without changing the actual mod load order.

Example use cases:

- mark all framework/library mods as `framework`;
- mark translation mods as `translation`;
- mark quality-of-life mods as `qol`;
- filter to show only untagged mods and categorize them gradually;
- temporarily show tag labels while filtering, without permanently cluttering the list.

### Notes

This feature stores tag data in the auxiliary metadata database. It does not modify `ModsConfig.xml` or the actual RimWorld load order.

### DIFF
--- a/app/sort/mod_sorting.py
+++ b/app/sort/mod_sorting.py
@@ -9,6 +9,7 @@ from PySide6.QtCore import QObject, Signal, Slot
 from app.controllers.metadata_db_controller import AuxMetadataController
 from app.utils.generic import scanpath
 from app.utils.metadata import MetadataManager, ModMetadata
+from app.utils.aux_db_utils import auxdb_get_mod_tags
 
 # Simple in-memory cache for folder sizes: {mod_path: (mtime, size_bytes)}
 _FOLDER_SIZE_CACHE: dict[str, tuple[int, int]] = {}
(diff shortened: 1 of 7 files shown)


cebarks:
Overall, looks pretty good to me. There are a few linting failures that should be fixed, but testing locally works well for me.

A few suggestions too that might be useful along with this functionality:
- update the mod info panel on the left side to include a list of tags for that mod.
- add a way to manage the tags from the sorting drop down.

VA-Wizard:
> Overall, looks pretty good to me. There are a few linting failures that should be fixed, but testing locally works well for me.
> 
> A few suggestions too that might be useful along with this functionality:
> 
>     * update the mod info panel on the left side to include a list of tags for that mod.
> 
>     * add a way to manage the tags from the sorting drop down.

Sounds good. Added info to the left panel and tags to sortings.

cebarks (approved):
LGTM

PR #1987, opened 2026-05-22
fix: add ruff format config to super-linter

cebarks:
## Summary
- `PYTHON_RUFF_CONFIG_FILE` only applies to `ruff check` in super-linter — `ruff format` needs its own `PYTHON_RUFF_FORMAT_CONFIG_FILE`
- Without it, super-linter defaults to `.ruff.toml` (which doesn't exist), so `ruff format` uses line-length 79 instead of the project's configured 88 in `pyproject.toml`
- Reformat all files to align with the expected `pyproject.toml` settings

## Test plan
- [ ] CI lint job should pass with PYTHON_RUFF_FORMAT no longer failing

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -72,6 +72,7 @@ jobs:
           GITHUB_ACTIONS_COMMAND_ARGS: >-
             -ignore 'unknown permission scope "attestations"'
           PYTHON_RUFF_CONFIG_FILE: pyproject.toml
+          PYTHON_RUFF_FORMAT_CONFIG_FILE: pyproject.toml
           PYTHON_MYPY_CONFIG_FILE: pyproject.toml
           FILTER_REGEX_EXCLUDE: "LICENSE.md|super-linter-output/|github_conf/"
           IGNORE_GITIGNORED_FILES: true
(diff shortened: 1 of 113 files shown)


LionelColaso:
Try pushing an unformatted file and if it auto fixes it without issues and runs the autobuild then its good

cebarks:
> Try pushing an unformatted file and if it auto fixes it without issues and runs the autobuild then its good

wait didn't you disable the ruff auto-fix earlier in a PR? Maybe instead of disabling it completely, we have it only format when merged to main?

LionelColaso:
@cebarks  i have removed the commit from the main branch but your pr still has the commit 
image

cebarks:
> @cebarks i have removed the commit from the main branch but your pr still has the commit

alright i'll reset this branch on top of current main and see if it works. even if it does correctly, it still won't trigger CI after the auto-fix commit is added.

LionelColaso:
> > @cebarks i have removed the commit from the main branch but your pr still has the commit
> 
> alright i'll reset this branch on top of current main and see if it works. even if it does correctly, it still won't trigger CI after the auto-fix commit is added.

it pushes the fix correctly but shouldnt it directly create a commit instead of merge commit that merge commit could be the issue

cebarks:
ya looks like the formatting fix commit doesn't trigger it, but if i rebase or merge main to the branch it does trigger correctly.

LionelColaso (approved):
looks good for now

PR #1999, opened 2026-05-24
chore: add jscpd and shfmt just recipes, update contributor docs

cebarks:
## Summary
- Add `just jscpd` recipe — runs `npx jscpd@latest` with CI's zero-tolerance threshold
- Add `just shfmt` recipe — checks shell script formatting
- Include both in `just check` so all CI checks run locally
- Update contributor guidelines with a task runner section and restructured coding style docs
- Add linter/formatter prerequisites (just, npx, shfmt) to the development setup guide

## Test plan
- [x] `just jscpd` runs and reports 0 clones
- [x] `just shfmt` runs against `update.sh` and `packaging/`
- [x] `just check` includes all quality checks

### DIFF
--- a/docs/development-guide/contributor-guidelines.md
+++ b/docs/development-guide/contributor-guidelines.md
@@ -102,13 +102,14 @@ Key recipes for contributors:
 
 | Command | Description |
 | :--- | :--- |
-| `just check` | Run all code quality checks (lint, format, typecheck, jscpd) |
+| `just check` | Run all code quality checks (lint, format, typecheck, jscpd, shfmt) |
 | `just fix` | Auto-fix linting and formatting issues |
 | `just test` | Run tests with coverage reporting |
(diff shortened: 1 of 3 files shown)


LionelColaso:
@cebarks it would be better if you also add it as a dev dependency in pyproject.toml  and update the docs\development-guide\contributor-guidelines.md

cebarks:
I'll update the contributor doc, but can you add a node dependency to pyproject.toml?

LionelColaso:
> I'll update the contributor doc, but can you add a node dependency to pyproject.toml?

It can't be directly added as requirement since UV does not have it and at least on Windows Node.js is required to run jscpd
So on second thought just updating the docs would be good to help user know how to install and use in case they don't understand how to install and check locally

LionelColaso (approved):

PR #2000, opened 2026-05-24
fix: resolve macOS .app bundle issues (SteamworksPy path, build config, auto-updater)

cebarks:
## Summary

Fixes three independent macOS `.app` bundle issues:

- **SteamworksPy library not found in compiled builds**: Add `AppInfo.libs_folder` property that returns the correct library path in both source mode (`libs/` subdirectory) and compiled Nuitka mode (application folder root). Update all callers to use it.
- **CI builds missing `SteamworksPy.dylib`**: The CI build workflow uses the Nuitka GitHub Action directly, bypassing `distribute.py`'s `post_build_fixup_macos_steamworks()`. Fold the fixup into `optimize_macos_bundle.py` which already runs as a post-build step.
- **Auto-updater unwraps `.app` bundles**: The `_should_unwrap_directory()` method checked for generic keywords like `"app"` before platform-specific rules, causing `RimSort.app` to be incorrectly unwrapped during update extraction since its name contains `"app"`. Move platform-specific preserve rules first.

### SteamworksPy library path (`MissingSteamworksLibraryException` crash)
- `app/utils/app_info.py` — New `libs_folder` property: returns `application_folder` in compiled mode, `application_folder / "libs"` in source mode
- `app/views/main_content_panel.py` — Use `AppInfo().libs_folder` instead of hardcoded path
- `app/utils/steam/webapi/wrapper.py` — Use `AppInfo().libs_folder` instead of hardcoded path

### Build pipeline
- `packaging/optimize_macos_bundle.py` — Add `fixup_steamworkspy()` step that copies `SteamworksPy_arm.dylib` → `SteamworksPy.dylib` in the bundle if missing. Always re-sign bundle after modifications.
- `rimsort.nuitka-package.config.yml` — No net change (reverted attempted simplification that broke cross-arch builds)

### Auto-updater
- `app/utils/update_utils.py` — Reorder `_should_unwrap_directory()` to check platform-specific preserve rules (e.g. keep `.app` bundles intact on macOS) before generic keyword patterns

## Test plan
- [x] `ruff format` / `ruff check` — clean
- [x] `mypy` — no issues
- [x] `pytest` — 343/344 pass (1 pre-existing failure in `test_dialogue.py` on `main`)
- [x] Verify macOS CI build succeeds and includes `SteamworksPy.dylib` in bundle
- [x] Test Steam DB build from compiled `.app` on macOS
- [x] Test auto-updater extraction on macOS

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -191,31 +191,6 @@ jobs:
         env:
           PYTHONPATH: ${{ steps.pythonpath.outputs.PYTHONPATH }}
 
-      - name: Fixup SteamworksPy in macOS bundle
-        if: runner.os == 'macOS'
-        shell: bash
-        run: |
-          APP_BUNDLE=$(find build/ -name "*.app" -maxdepth 2 | head -n 1)
-          if [ -z "$APP_BUNDLE" ]; then
-            echo "No .app bundle found, skipping SteamworksPy fixup"
-            exit 0
-          fi
-          MACOS_DIR="$APP_BUNDLE/Contents/MacOS"
-          GENERIC="$MACOS_DIR/SteamworksPy.dylib"
-          if [ -f "$GENERIC" ]; then
-            echo "SteamworksPy.dylib already exists in bundle"
-            exit 0
-          fi
-          # Find arch-specific variant and copy to generic name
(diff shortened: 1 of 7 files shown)


cebarks:
This is worth an immediate bugfix release after merge as it fixes a crash on macos when any steamworks related feature is used.

LionelColaso (approved):
Looks Good Thanks

PR #2006, opened 2026-05-25
Skip BinaryChoiceDialog test on macOS

cebarks:
## Summary
- Skip `windowTitle()` assertion on macOS in `TestBinaryChoiceDialog::test_initialization`
- `QMessageBox.windowTitle()` always returns empty on macOS because Qt uses native message box styling that doesn't track the window title
- CI only runs on `ubuntu-latest` so this was never caught — fails locally on macOS

## Test plan
- [x] All 367 tests pass locally on macOS
- [x] Lint and format checks pass
- [x] CI passes on Linux (no behavior change there)

### DIFF
--- a/tests/views/test_dialogue.py
+++ b/tests/views/test_dialogue.py
@@ -1,3 +1,4 @@
+import sys
 from typing import Generator
 from unittest.mock import MagicMock, patch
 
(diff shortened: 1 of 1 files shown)


LionelColaso (approved):
Looks Good Thanks

PR #2018, opened 2026-05-30
[TO BE REPLACED] fix(metadata): skip missing publishedfileid warning for local mods

cebarks:
## Summary

Fixes #1853

- Local mods (not from Steam Workshop) no longer generate false "missing publishedfileid" warnings on every launch
- The `_get_missing_publishfieldid_uuids()` method now filters by `data_source == "workshop"` so only workshop mods are checked
- The `_row_has_missing_pfid()` non-explicit branch now checks the SOURCE column against `STEAM`/`STEAM_CMD` constants
- Adds tests for both code paths and a `tests/views/conftest.py` for steamworks mock

## Test plan

- [x] Launch RimSort with local mods installed — no false publishedfileid warnings should appear
- [x] Verify `uv run pytest tests/views/test_missing_publishfieldid.py -v` passes
- [x] Verify full test suite passes: `uv run pytest tests/ -x -q`

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -648,7 +648,8 @@ class MainContent(QObject):
 
         Workshop mods without a Publish Field ID cannot support automatic
         redownloads or update checking. This check intentionally excludes
-        RimWorld core content and DLC since they are not published mods.
+        RimWorld core content, DLC, and local mods since they are not
+        published to Steam Workshop.
 
         :return: List of internal UUIDs for mods with missing Publish Field ID.
         """
@@ -656,7 +657,8 @@ class MainContent(QObject):
         return [
             uuid
             for uuid, mod_metadata in self.metadata_manager.internal_local_metadata.items()
-            if mod_metadata.get("publishedfileid") is None
+            if mod_metadata.get("data_source") == "workshop"
+            and mod_metadata.get("publishedfileid") is None
             and mod_metadata.get("packageid") not in app_constants.RIMWORLD_PACKAGE_IDS
             and mod_metadata.get("packageid") not in ignored_mods
         ]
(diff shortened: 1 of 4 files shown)


LionelColaso:
>     * Local mods (not from Steam Workshop) no longer generate false "missing publishedfileid" warnings on every launch
> 
>     * The `_get_missing_publishfieldid_uuids()` method now filters by `data_source == "workshop"` so only workshop mods are checked

I think the better fix would be to just check local mods folder but then again i have seen mods downloaded using steam and steamcmd missing publishfieldid
that was the reason for adding "ignore.json"
a clean way to do it would be to check all mods folder combined and then differentiating mods which are being developed and not from mods which were downloaded using steam and steamcmd
I have an idea on how to do this an easy way while keeping the current  "ignore.json" behaviour so steam and steamcmd mods without publishfieldid can be ignored
Without publishfieldid steam and steamcmd can't update or redownload the mod thats why we have a steam database so we can add map and add publishfieldid  to mods which dont have it
 but if steamdb also does not have publishfieldid for the mod and the mod not having publishfieldid itself there is still a way to assign publishfieldid  to the mod (i think i have actually added it) 
 
 so the way you are doing it breaks the whole concept of why the user was being notified in the first place 
 
 Edit1:
 mods cant be updated or redownloaded without publishfieldid
 example of scenerio when publishfieldid is not available https://discord.com/channels/1120855389651599400/1510291798042480650
 
 Edit2:
 also the reason is so that mod authors add publishfieldid to their mods so steamdb does not need to be made since it will already have data for old mods

LionelColaso (changes requested):
see comment

cebarks:
Hmm ok, i see your point. You said you have an idea to fix this, so should i close this PR then so you can tackle it?

LionelColaso:
> Hmm ok, i see your point. You said you have an idea to fix this, so should i close this PR then so you can tackle it?

Keep the PR open till i push the fix so that i don't forget about it

cebarks:
I'll convert it to a draft for now

PR #2019, opened 2026-05-30
fix(ui): fix save button not blinking on mod drag-drop

### LINKED ISSUE #1819: Save button not blinking when it should
### Release Type

Compiled (Windows x86-64)

### Version

1.0.74

### Operating System

Windows 11

### Relevant logs

_No response_

### Describe the bug

Moving a mod somewhere within the active list, or moving a mod between Active and Inactive doesn't make the Save button blink to indicate the modlist is not currently saved.

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

cebarks:
## Summary

Fixes #1819

- Save button now blinks when mods are dragged between active/inactive lists (cross-list drop)
- Root cause: `dropEvent` manually inserted UUIDs into `self.uuids` AND the queued `handle_rows_inserted` also inserted them, creating duplicates that broke the `len(self.uuids) == self.count()` guard — so `list_update_signal` was never emitted
- Fix: only manipulate UUIDs in `dropEvent` for within-list reorder (`source_widget == self`); cross-list drops let `handle_rows_inserted` handle UUID insertion exclusively (same pattern as PR #1922)
- Removes unused `save_btn_animation_signal` and `list_updated_signal` from `ModsPanel` (defined and emitted but never connected to any handler)

## Test plan

- [x] Drag a mod within the active list to reorder — save button should blink
- [x] Drag a mod from inactive to active — save button should blink
- [x] Drag a mod from active to inactive — save button should blink
- [x] Click Save — save button should stop blinking
- [x] Verify full test suite passes: `uv run pytest tests/ -x -q`

### DIFF
--- a/app/controllers/main_window_controller.py
+++ b/app/controllers/main_window_controller.py
@@ -325,8 +325,6 @@ class MainWindowController(QObject):
                     self.main_window.main_content_panel.mods_panel.active_mods_list.uuids = list(
                         active_mods
                     )
-                    # Trigger list updated signal to refresh UI
-                    self.main_window.main_content_panel.mods_panel.list_updated_signal.emit()
 
                 # If there are mods to download, check SteamCMD setup first
                 if mods_to_download:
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Works Thanks

PR #2020, opened 2026-05-30
fix(settings): persist inactive mods sort key and fix sorting checkbox disable bug

### LINKED ISSUE #1771: Sort key for inactive mods not saved to settings
### Release Type

Python Interpreter

### Version

4149ab8

### Operating System

Windows 11

### Relevant logs

_No response_

### Describe the bug

The sort key a user selects is not saved to settings. But there is an entry for it in settings.

Image

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

cebarks:
## Summary

Fixes #1771

**Sort key persistence (folder size):**
- The heavy-sort (Folder Size) branch was missing save logic — the sort key was never persisted to settings even when "Save inactive mods sort state" was enabled
- Added the same save logic that already existed in the non-heavy sort branch

**Sort UI visibility:**
- The sort combobox and Asc/Desc button in the mods panel now hide/show dynamically when the "Enable inactive mods sorting" setting changes (via `EventBus.settings_have_changed`)
- Fixed init-time visibility to properly hide sort widgets when sorting is disabled at startup

**Note:** The checkbox disable-recursion fix was dropped from this PR — that was independently resolved by #2069.

## Test plan

- [x] With "Save inactive mods sort state" checked: change sort key to Folder Size → restart → sort key persists
- [x] With "Save inactive mods sort state" unchecked: change sort key → restart → resets to Modified Time / Desc
- [x] When "Enable inactive mods sorting" is unchecked, the sort combobox and Asc/Desc button in the mods panel disappear
- [x] When re-checked, the sort UI reappears immediately (no restart needed)
- [x] Full quality gate passes: `just check`
- [x] Full test suite passes: `just test` (729 tests)

### DIFF
--- a/app/views/mods_panel.py
+++ b/app/views/mods_panel.py
@@ -3977,7 +3977,15 @@ class ModsPanel(QWidget):
         Update the inactive mods sort UI elements from settings.
 
         Restores the saved sort key and direction to the UI combobox and button.
+        Toggles visibility of sort widgets based on whether sorting is enabled.
         """
+        sorting_enabled = self.settings_controller.settings.inactive_mods_sorting
+        self.inactive_mods_sort_combobox.setVisible(sorting_enabled)
+        self.inactive_mods_sort_order_button.setVisible(sorting_enabled)
+
+        if not sorting_enabled:
+            return
+
         self.inactive_mods_sort_key = (
             self.settings_controller.settings.inactive_mods_sort_key
         )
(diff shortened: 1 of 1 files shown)


LionelColaso:
@cebarks there is a bug cannot enable sorting if its disabled in settings

cebarks:
> @cebarks there is a bug cannot enable sorting if its disabled in settings

I think that's a pre-existing bug but i'll fix it in this PR as well. Should be a simple fix (moving the checkbox out of the disabled group).

cebarks:
Fixed the bug, good to go with your approval @LionelColaso

LionelColaso:
@cebarks Sorry the marked UI  should not load when sorting inactive mods setting is disabled 
between i noticed now that option to save the sorting state as well is missing (this was specifically asked some users wanted to save state some wanted default state thats the reason there were two checkboxes)
image

cebarks:
Ohh I understand the intended behavior now. I'll update this PR to address that.

cebarks:
@LionelColaso ok, it should be the correct behavior now!

LionelColaso:
@cebarks 
image

cebarks:
> @cebarks image

that's the correct behavior right? Or are you thinking it should be more deeply nested like the sections above?

LionelColaso:
> that's the correct behavior right? Or are you thinking it should be more deeply nested like the sections above?

image

cebarks:
> > that's the correct behavior right? Or are you thinking it should be more deeply nested like the sections above?
> 
> image

ahhh gotcha, i see what you were thinking now. wanna just push that change directly?

cebarks:
@LionelColaso Is this replaced by #2069?

LionelColaso:
> @LionelColaso Is this replaced by #2069?

Yes, but the whole implementation can be improved. i saw scope for improvements but did not have an opportunity to implement it at that moment

PR #2025, opened 2026-05-30
fix(tests): prevent tests from launching Steam via unpatched steam:// URLs

cebarks:
## Summary

- Fix 4 tests in `test_troubleshooting.py` that patched `platform_specific_open` at the wrong import location (`app.utils.generic` instead of `app.controllers.troubleshooting_controller`), causing `steam://install`, `steam://validate`, and `steam://workshop_download_item` URLs to reach the real OS and launch Steam on every test run
- Add a `subprocess.Popen` safety net fixture in `conftest.py` that blocks any `steam://` URL from reaching the OS during tests, so future regressions fail loudly instead of silently launching Steam

## Test plan

- [x] All 367 tests pass
- [x] No `steam://` URLs leak through (verified with diagnostic instrumentation)
- [x] Steam no longer launches during `uv run pytest`
- [x] Ruff format and lint pass

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,5 @@
 import subprocess
-from typing import Generator, Union
+from typing import Any, Generator, Union
 from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Yeah Thanks For This really annoying when running tests should have fixed it sooner

PR #2027, opened 2026-05-30
build(i18n): add translation compilation pipeline

cebarks:
## Summary

- **Stop tracking compiled `.qm` files** — they're now generated from `.ts` sources via `pyside6-lrelease`, eliminating binary diffs in translation PRs and preventing stale translations
- **Add `just i18n-compile` / `just i18n-update` recipes** with `[unix]`/`[windows]` platform variants, wired into `dev-setup`, `build`, and `build-version`
- **Add CI build step** — compiles `.ts` → `.qm` before Nuitka packaging so translations are always fresh
- **Fix justfile shell default** — replaced global `set shell := ["pwsh", ...]` (required PowerShell 7 on all platforms) with `set windows-shell := ["powershell.exe", ...]` so Unix uses default `sh` and Windows uses the built-in PowerShell
- **Update translation docs** — both English and Chinese guides updated to reflect that `.qm` files are generated artifacts

## Details

The Nuitka package config (`rimsort.nuitka-package.config.yml`) already bundles the `locales/` directory — no changes needed there. The only gap was that `.qm` files were committed as source rather than generated during the build.

### New recipes

| Recipe | Description |
|--------|-------------|
| `just i18n-compile` | Compile all `.ts` → `.qm` via `pyside6-lrelease` |
| `just i18n-update` | Extract translatable strings from source into `.ts` files |

Both run automatically: `i18n-compile` is called by `dev-setup` (after `uv sync`) and is a dependency of `build`/`build-version`.

### Justfile shell fix

The global `set shell := ["pwsh", "-NoProfile", "-Command"]` was added in #2007/#2013 to fix a Windows issue but required PowerShell 7 on macOS/Linux where it isn't installed by default. Since all recipes use plain cross-platform commands (or already have `[unix]`/`[windows]` guards), the fix is to use `set windows-shell` instead — which only affects Windows and uses the built-in `powershell.exe`.

## Test plan

- [x] `just i18n-compile` generates all 11 `.qm` files from `.ts` sources
- [x] `just --list` shows new recipes
- [x] `just ruff` and `just ruff-format` pass (verify shell change doesn't break recipes)
- [x] Generated `.qm` files are gitignored (`git status locales/` shows clean)
- [x] Compilation is idempotent (running twice produces identical `.qm` files)
- [x] `just clean` removes generated `.qm` files
- [x] CI build succeeds (compile translations step runs before Nuitka)


LionelColaso:
@cebarks This is Amazing This was a goal discussed by me and @oceanfrost to automate this so Thank You for doing this indeed a great help
just curious about why u removed .qm files since Google does not translate files correctly all the time

cebarks:
the `.qm` files are just the binary, compiled versions of the `.tm` source files. This doesn't automatically update the translations, it just ensures all the `.qm` files are fresh before the nuitka builds kick off.

The `just i18n-update` just scans the `app/` source code and finds Qt translation calls that don't exist in the translation files and adds them without a translation, so its clear which strings haven't be done yet.

oceanfrost (approved):

LionelColaso:
> the `.qm` files are just the binary, compiled versions of the `.tm` source files. This doesn't automatically update the translations, it just ensures all the `.qm` files are fresh before the nuitka builds kick off.
> 
> The `just i18n-update` just scans the `app/` source code and finds Qt translation calls that don't exist in the translation files and adds them without a translation, so its clear which strings haven't be done yet.

If possible add a note in the docs to run `just clean` to remove stale `.qm` files.  or automate the clean process in the i18n command

cebarks:
Went ahead and updated the compile recipe to delete the .qm before recompiling.

Issue #1676, opened 2026-01-03
Sorting appears to fail on large mod lists


tenmeiiii:
### Release Type

Compiled (Windows x86-64)

### Version

v1.0.65

### Operating System

Windows 10 Pro build 19045.6691

### Relevant logs

https://files.catbox.moe/g8ufnb.txt

### Describe the bug

When pressing the sort button on [this modlist](https://github.com/user-attachments/files/24414227/celetechplus.xml) VEF is loaded below VVE, even though RimSort acknowledges that this sort is wrong.

### Additional context

<img width="589" height="108" alt="Image" src="https://github.com/user-attachments/assets/ad8c79bf-f3da-446c-85d2-d40bb01b36f5" />

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

cebarks:
From my debugging, this appears to be an issue related to alphabetical sort and how it's non-deterministic (a slightly different mod list can change the sort order). Can you try switching to topographical sort and see if the issue persists @tenmeiiii?

PR #2034, opened 2026-05-31
feat(sort): soft-deprecate alphabetical sort

cebarks:
## Summary

- Add deprecation tooltip to the Alphabetical sorting radio button in settings: *"Alphabetical sorting may produce incorrect results with complex mod lists. Topological sorting is recommended."*
- Log `logger.warning()` when alphabetical sort is selected

## Context

Investigation of #1676 found that the alphabetical sort has no cycle detection — when conflicting load order rules from different metadata sources create dependency cycles, it silently produces incorrect results. The topological sort already detects and reports these cycles via `CircularDependencyError`. Since the topological sort also sorts alphabetically within each dependency level, it provides similar ordering with correctness guarantees.

This is a soft deprecation — the alphabetical sort option remains functional. No UI banners, no blocking, no removal.

Ref #1676

## Test plan

- [x] All 367 tests pass
- [x] Lint, format, mypy, jscpd all clean
- [x] Tooltip appears on hover over "Alphabetically" radio button in settings
- [x] Warning logged when alphabetical sort is selected

### DIFF
--- a/app/controllers/sort_controller.py
+++ b/app/controllers/sort_controller.py
@@ -28,6 +28,10 @@ class Sorter:
             logger.info(f"Created sorter instance with {sort_method} sort method")
 
             if sort_method == SortMethod.ALPHABETICAL:
+                logger.warning(
+                    "Alphabetical sort is deprecated and may produce incorrect results "
+                    "with complex mod lists. Consider switching to Topological sort."
+                )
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):

LionelColaso (approved):
Looks Good

PR #2062, opened 2026-06-01
fix(ui): add confirmation before deleting optimized textures

cebarks:
## Summary
- Add a confirmation dialog to the global "Delete .dds Textures" menu action (`Tools > Textures > Delete .dds Textures`)
- Previously this action ran immediately without any user confirmation, unlike other destructive operations
- Uses the same `show_dialogue_conditional` pattern as ACF import and other destructive actions in the codebase

## Test plan
- [x] Open RimSort, go to Tools > Textures > Delete .dds Textures
- [x] Verify confirmation dialog appears with title, description, and "Delete textures" button
- [x] Click cancel/close — verify no deletion occurs
- [x] Click "Delete textures" — verify todds clean runs as before

### DIFF
--- a/app/views/main_content_panel.py
+++ b/app/views/main_content_panel.py
@@ -2234,6 +2234,22 @@ class MainContent(QObject):
 
         The operation runs asynchronously in a UI panel showing progress.
         """
+        answer = dialogue.show_dialogue_conditional(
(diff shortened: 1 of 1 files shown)


LionelColaso (approved):
Looks Good

PR #2075, opened 2026-06-02
fix(ci): Nuitka version mismatch and broken status-check

cebarks:
## Summary

- **Nuitka version alignment**: `build.yml` pinned Nuitka `4.0.8` while `pyproject.toml` specified `4.1.1`. Updated the workflow to match, preventing local/CI build divergence.
- **status-check actually checks results**: The `status-check` job in `auto_build.yml` used `if: always()` but never inspected `needs.*.result`, so it unconditionally reported success — even when upstream jobs failed. Now iterates over all upstream results and fails on `failure`/`cancelled`, while allowing `skipped` (since `attested-build` and `non-attested-build` are mutually exclusive).

## Test plan

- [x] Verify CI passes on this PR (non-attested build path exercises the status-check logic)
- [x] Confirm Nuitka version in build logs shows 4.1.1

### DIFF
--- a/.github/workflows/auto_build.yml
+++ b/.github/workflows/auto_build.yml
@@ -70,5 +70,19 @@ jobs:
       - non-attested-build
     if: always()
     steps:
-      - name: Set status check
-        run: echo "All required jobs have completed."
+      - name: Verify all required jobs succeeded
+        run: |
+          results=(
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Looks Good

PR #2077, opened 2026-06-02
refactor(watchdog): self-contained lifecycle, fix elif shutdown bug

cebarks:
## Summary

- **Fix elif shutdown bug**: `MainWindow.shutdown_watchdog()` used `elif` between ACF and mods observer shutdown — only one observer was ever stopped per call. Both observers now independently checked via `if`/`if`.
- **Move lifecycle into WatchdogHandler**: Added `start()` and `stop()` methods so the handler owns its own lifecycle. `MainWindow` is now a thin consumer that calls `start()`/`stop()`.
- **Fix signal accumulation bug**: `stop_watchdog_signal.connect()` was called inside `initialize_watchdog()`, adding a new connection on every instance switch. Moved to `MainWindow.__init__()` so it connects exactly once.
- **Fix timer dict iteration race**: `stop()` now snapshots `cooldown_timers` before clearing/cancelling to prevent `RuntimeError` if a pending `Timer` callback modifies the dict during iteration.
- **Cleanup**: Removed unused `targets` parameter from `WatchdogHandler.__init__()`, removed empty `on_moved`/`on_closed`/`on_opened` stubs, removed redundant `stop_watchdog_if_running()` wrapper, fixed pre-existing `update_utils.py` import issue.

## Test plan

- [x] 14 new unit tests for `WatchdogHandler.start()`/`stop()` lifecycle (`tests/utils/test_watchdog.py`)
- [x] 2 delegation tests for `MainWindow.shutdown_watchdog()` (`tests/views/test_main_window_watchdog.py`)
- [x] 3 existing `closeEvent` tests updated and passing (`tests/views/test_main_window_close.py`)
- [x] All quality gates pass (ruff, ruff-format, mypy, pyright, jscpd, shfmt, markdownlint)

### DIFF
--- a/app/utils/update_utils.py
+++ b/app/utils/update_utils.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import os
 import platform
 import re
--- a/app/utils/watchdog.py
+++ b/app/utils/watchdog.py
@@ -22,9 +22,7 @@ class WatchdogHandler(FileSystemEventHandler, QObject):
     mod_deleted = Signal(str, str, str)
     mod_updated = Signal(bool, bool, str, str, str)
 
-    def __init__(
-        self, settings_controller: SettingsController, targets: list[str]
-    ) -> None:
+    def __init__(self, settings_controller: SettingsController) -> None:
         """Initialize the WatchdogHandler.
 
         The WatchdogHandler is a subclass of :class:`watchdog.events.FileSystemEventHandler`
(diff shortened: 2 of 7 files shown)


LionelColaso:
@cebarks Interesting i have not checked it may be in the morning but if i remember correctly player_log_tab.py also has its own watchdog observer if possible you could check that as well not sure if another file uses it but just do a search in case anything else not using the common watchdog

cebarks:
> @cebarks Interesting i have not checked it may be in the morning but if i remember correctly player_log_tab.py also has its own watchdog observer if possible you could check that as well not sure if another file uses it but just do a search in case anything else not using the common watchdog

it does have its own watchdog observer but its VERY different from this one. It does have some of the same issues as this one did so I'll tackle it in a follow-up PR.

LionelColaso (approved):
looks good

PR #2086, opened 2026-06-03
refactor(settings): extract SortingTabController from SettingsController

cebarks:
## Summary

- Introduces `BaseTabController` ABC defining the `update_view_from_model` / `update_model_from_view` / `connect_signals` contract for per-tab settings controllers
- Extracts all Sorting tab logic from the monolithic `SettingsController` (2760 → 2675 lines) into a self-contained `SortingTabController`
- Adds 7 tests covering both model→view and view→model sync directions
- Pure refactor — zero behavior change, all 722 existing tests pass

This is a proof-of-concept for the tab extraction pattern. Once validated, the remaining 12 tabs can be extracted incrementally using the same approach.

## Test plan

- [x] All 722 tests pass (`just ci`)
- [x] All quality checks pass (ruff, mypy, pyright, jscpd, shfmt, markdownlint)
- [x] 0 jscpd clones (test fixture extracted to conftest.py)
- [x] Manual verification: launched app, opened Settings → Sorting tab, toggled options, confirmed save/load works


cebarks:
@LionelColaso I'm curious on your opinion on this approach to refactoring the different tabs from the monolithic settings controller?

LionelColaso:
> @LionelColaso I'm curious on your opinion on this approach to refactoring the different tabs from the monolithic settings controller?

Do you want to convert each tab into classes?

cebarks:
> > @LionelColaso I'm curious on your opinion on this approach to refactoring the different tabs from the monolithic settings controller?
> 
> Do you want to convert each tab into classes?

That was the thought, should make it much cleaner to modify any of the tabs and make diffs cleaner and easier to review (and also minimize conflicts that need to be resolved). Should also make future GUI testing significantly easier as well.

LionelColaso:
I did want to rewrite each tab setting. As class but not sure if we want to breakitdown into 13-15 individual files

cebarks:
> I did want to rewrite each tab setting. As class but not sure if we want to breakitdown into 13-15 individual files

I was also thinking we should probably condense the number of settings tab in general. There are so many and it can get confusing to find what you actually need. Breaking them down into files would be a good opportunity to do that.

LionelColaso:
The below can be merged; not sure about others, though, yet.
1) Theme and Launch State
2) Auxiliary DB and Advanced

LionelColaso (approved):
Looks good, just resolve conflicts and hopefully don't break anything :-P

cebarks:
> The below can be merged; not sure about others, though, yet.
> 
>     1. Theme and Launch State
> 
>     2. Auxiliary DB and Advanced

Agreed on `Aux DB` and `Advanced` but I was thinking `Launch State` and `Game Launch` should definitely be together as well, so maybe all three combined into `Advanced`?

I was thinking `todds`, `steamcmd` into `External Tools` as well.

Thoughts?

LionelColaso:
> > The below can be merged; not sure about others, though, yet.
> > ```
> > 1. Theme and Launch State
> > 
> > 2. Auxiliary DB and Advanced
> > ```
> 
> Agreed on `Aux DB` and `Advanced` but I was thinking `Launch State` and `Game Launch` should definitely be together as well, so maybe all three combined into `Advanced`?
> 
> I was thinking `todds`, `steamcmd` into `External Tools` as well.
> 
> Thoughts?

The main issue for having tabs was to support various display settings and screen sizes and not congest the content. And support a small display size monitor to a 4K display. 
If tabs are merged, they take display space. And the settings panel won't show correctly on all displays beyond the current size
That's why I liked the idea of a scroll bar in the database tab (somehow I never thought about it or was lazy)

`External Tools` has more things to add if I recall.  Will have to check issues. 
toods and steamcmd keep them in separate tabs
Steamcmd and Todd's being separate is good 
Game launch and launch state need to be separate since game launch and instance UI need to be added together when made. Hopefully, also check issues for the exact issue
 launch state more windows needs to be added to launch state

cebarks:
I'd say it's much more congested to have a ton of tabs with a few items compared to having scroll-able panels with more settings. We can have named sections in the larger merged tabs for each grouping of settings, instead of having them all flat.

> toods and steamcmd keep them in separate tabs
> Steamcmd and Todd's being separate is good

Why? IMO it significantly clutters up the settings tabs and is confusing when we have a tab already called "External Tools". That's exactly what they are.

PR #2088, opened 2026-06-03
fix(justfile): exclude .venv and submodules from shfmt checks

cebarks:
## Summary
- `shfmt -d .` / `shfmt -w .` walked into `.venv/` and diffed vendored scripts (e.g. Nuitka's `compile-python-for-nuitka-mac.sh`), causing `just check` to fail
- CI avoided this via super-linter's `IGNORE_GITIGNORED_FILES: true`, but the local justfile recipes did not — creating a local/CI desync
- Use `fd` to scope shfmt to project-owned `.sh` files, excluding `.venv` and `submodules`

## Test plan
- [x] `just check` passes locally
- [x] `just shfmt` passes locally
- [ ] CI lint job passes

### DIFF
--- a/justfile
+++ b/justfile
@@ -77,11 +77,11 @@ jscpd:
 
 # Check shell script (.sh) formatting (shfmt) — diff-only, no changes made
 shfmt:
-    shfmt -d .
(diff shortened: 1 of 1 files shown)


LionelColaso (approved):
looks good

Issue #2051, opened 2026-05-31
(major) Replace MetadataManager with MetadataController/MetadataMediator


cebarks:
## Summary

Complete the metadata rework started in PR #403 by replacing the old `MetadataManager` (`app/utils/metadata.py`, ~2,895 lines) with the `MetadataController`/`MetadataMediator` system that was built but never integrated.

**Design spec:** `docs/superpowers/specs/2026-05-31-metadata-manager-replacement-design.md`

## Motivation

The old `MetadataManager` is a 2,895-line singleton using untyped `dict[str, Any]` with 35+ string keys and regenerated random UUIDs. PR #403 (merged 2025-03-19) built a replacement backend with typed dataclasses (`ListedMod`, `AboutXmlMod`), msgspec-based schema parsing, multi-threaded `MetadataMediator`, and SQLAlchemy/SQLite auxiliary metadata. The final integration step — "Integrate new metadata system" — was explicitly deferred.

## Key Architectural Changes

- **Identifier model:** Filesystem PATH replaces random UUIDs (deterministic, aligns with AuxMetadataController)
- **Typed data structures:** `dict[str, Any]` → `ListedMod`/`AboutXmlMod` dataclasses
- **Standalone dependency graph:** `CompiledDependencyData` structure built by MetadataController, consumed by sort algorithms as parameters (no singleton access)
- **ByVersion precedence fix:** Factory updated from merge to non-additive override (matching old system behavior)
- **Sort algorithm decoupling:** Sort functions become pure functions receiving data as parameters

### PR 1: Port authority into MetadataController
- Port `compile_metadata()` logic, add Qt signals, add utility methods
- Fix ByVersion precedence in factory (`create_base_rules`)
- Add No Version Warning + "Use This Instead" loading to MetadataMediator
- Add WebAPI timestamp caching in aux DB
- Parity tests: run both systems, assert matching outputs
- **No consumers migrate yet — nothing breaks**

### PR 2: Migrate sort algorithms (fixes #2041, #2042)
- Sort functions receive `CompiledDependencyData` instead of accessing MetadataManager
- Delete `gen_deps_graph()`/`gen_rev_deps_graph()` (absorbed into `compile()`)
- #2041: Tier classification uses `KNOWN_TIER_*_MODS.copy()` — constants never mutated
- #2042: Add cycle protection to `get_reverse_dependencies_recursive()`

### PR 3: Migrate views and controllers
[...]

PR #2094, opened 2026-06-04
refactor(metadata): MetadataController API cleanup for PR 3 migration

cebarks:
## Summary

Adds missing API surface, fixes a key mapping bug, and adds performance caching to `MetadataController` — preparing it for PR 3 of the metadata migration (#2051). **Purely additive: no consumer migrations.**

- **Bug fix:** `steamdb_packageid_to_name` was using published file IDs (dict keys) as mapping keys instead of actual `entry.packageId` values. Any downstream `.get("some.package.id")` returned `None`.
- **Bug fix:** `MetadataMediator` paths (`local_mods_path`, `game_path`) were only synced from settings once at init. After autodetecting and saving new paths, the mediator retained stale `None` values, causing `ValueError: Essential paths are missing` on refresh. Fixed by calling `reset_paths()` at the top of `refresh_metadata()`.
- **Bug fix:** `MetadataMediator.refresh_metadata()` used `iterdir()` to collect mod paths, which returned non-directory entries like `Place mods here.txt` and `Place official expansions here.txt`. These placeholder files caused `ValueError: Path must be a directory` crashes in the parser workers. Fixed by filtering `iterdir()` results to directories only.
- **Bug fix:** `create_base_rules()` logged noisy warnings for empty/null XML dependency entries (`{}`, `@isNull="True"`, blank strings) that `xmltodict` produces from common placeholder elements in mod `About.xml` files. These are now silently skipped.
- **Bug fix:** `_ParserWorker` error and warning logs printed the entire batch list (`self.mod_path`) instead of the individual failing path, making errors hard to diagnose. Fixed to log the specific `path`.
- **Caching:** `packageid_to_paths` and `steamdb_packageid_to_name` are now lazily cached with invalidation in `refresh_metadata()`, `reset_paths()`, and `delete_mod()`
- **New API:** `user_rules` property, `workshop_acf_path` property (`Path | None`), `get_mod_name_from_package_id()` method, `workshop_acf_data`/`steamcmd_acf_data` attributes
- **DRY:** Extracted inline workshop ACF path computation from `refresh_metadata()` into `workshop_acf_path` property

### Behavioral notes for PR 3 reviewers
- `steamdb_packageid_to_name` now prefers `steamName` (Workshop display title) over `name` — intentional UX improvement
- `workshop_acf_path` returns `Path | None` (old MetadataManager used `str`) — type migration needed in PR 3
- `get_mod_name_from_package_id()` adds Steam DB fallback before raw package_id — richer than old MetadataManager

## Test plan
- [x] 18 unit tests covering all new API surface (caching, invalidation, edge cases, bug regression)
- [x] 791/791 full test suite passing
- [x] `just check` clean (ruff, mypy, pyright, jscpd, shfmt, markdownlint)
- [x] Four independent code review passes (spec compliance, quality, final adversarial, human)
- [x] Manual verification: autodetect paths → save → mods load without ValueError
- [x] Manual verification: no `Path must be a directory` errors or empty dependency warnings on app startup with real mod library

### DIFF
--- a/app/controllers/metadata_controller.py
+++ b/app/controllers/metadata_controller.py
@@ -40,6 +40,8 @@ class MetadataController(QObject):
     mod_metadata_updated_signal = Signal(str)
     show_warning_signal = Signal(str, str, str, str)
 
+    # ---- Lifecycle ----
+
     def __init__(
         self,
         settings_controller: SettingsController,
@@ -89,10 +91,6 @@ class MetadataController(QObject):
             cls._instance = cls(settings_controller, metadata_db_controller)
         return cls._instance
 
-    def _invalidate_caches(self) -> None:
-        self._packageid_to_paths_cache = None
-        self._steamdb_packageid_to_name_cache = None
-
     @Slot()
     def refresh_metadata(self) -> None:
         """Refresh the metadata."""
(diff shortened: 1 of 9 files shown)


LionelColaso (approved):
Looks Good

PR #2099, opened 2026-06-05
fix: recursively search Assemblies directory for DLLs

### LINKED ISSUE #1070: Empty assemblies folder in mod root causes RimSort to believe the mod is XML only, even if a DLL is loaded elsewhere
### Release Type

Compiled (Windows x86-64)

### Version

1.0.30

### Operating System

Windows 11

### Relevant logs

_No response_

### Describe the bug

The mod SlaveBar is a DLL mod, however, it shows up in RimSort as an XML mod, despite having a DLL. This is because it has an empty Assemblies folder in the root directory.

Ideally, Rimsort should be able to ignore such empty directories, and continue checking in subfolders.

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

cebarks:
## Summary
- Fixes mods with DLLs nested inside subdirectories of `Assemblies/` (e.g. `Assemblies/net472/Mod.dll`) or with an empty root `Assemblies/` folder being incorrectly classified as XML-only
- Changed `glob` → `rglob` in `subfolder_contains_candidate_path` for recursive search
- Replaced the ~30-line inline C# detection block in `metadata.py` with the shared `subfolder_contains_candidate_path` helper (which already handles subfolder iteration correctly)

## Test plan
- [x] Added test for DLLs nested inside subdirectories of Assemblies (`Assemblies/net472/mod.dll`)
- [x] Added test for empty root `Assemblies/` with DLLs in versioned subfolder (`1.5/Assemblies/mod.dll`)
- [x] All existing tests pass
- [x] Manual verification with a mod like SlaveBar that has an empty root Assemblies folder

Closes #1070

### DIFF
--- a/app/utils/files.py
+++ b/app/utils/files.py
@@ -46,9 +46,9 @@ def subfolder_contains_candidate_path(
         candidate_path = subfolder_path / candidate_directory
         if candidate_path.exists() and candidate_path.is_dir():
             if case_sensitive:
-                return any(candidate_path.glob(glob_pattern))
+                return any(candidate_path.rglob(glob_pattern))
             else:
-                for file_path in candidate_path.glob(glob_pattern):
+                for file_path in candidate_path.rglob(glob_pattern):
                     if fnmatch.fnmatch(file_path.name.lower(), glob_pattern.lower()):
                         return True
         return False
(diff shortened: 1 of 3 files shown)


LionelColaso (approved):
Looks Good

PR #2105, opened 2026-06-07
ci: auto-create Edge releases on push to main

cebarks:
## Summary
- Pushes to main now automatically create a GitHub Edge release (non-draft, prerelease) when the build, pyright, and tests all pass
- Edge releases are named "Edge Build (Unstable)" with a warning banner linking to stable releases
- Only one Edge release exists at a time — the previous one is deleted before creating the new one
- Changelog is auto-generated by GitHub, showing all changes since the last stable release
- Removed dead Fedora/RPM CI code from `build.yml` and `release.yml`
- Pinned `ncipollo/release-action` to commit SHA in both workflows

## How it works
- `pre-build` now passes `type: Edge` and `format_only: false` for pushes to main (full version computation), and `type: Auto-Build` / `format_only: true` for PRs (unchanged)
- New `edge-release` job gates on `attested-build`, `pre-pyright`, and `pre-test` all succeeding
- Artifact packaging mirrors the existing `release.yml` logic (tar→tar.gz/zip, MSI, AppImage)
- Manual Edge releases via `release.yml` still work as a fallback

## Test plan
- [ ] Merge a PR that touches `app/` and verify the Edge release is created automatically
- [ ] Verify the release is marked as prerelease with the "Edge Build (Unstable)" title
- [ ] Verify the warning banner and changelog appear in the release body
- [ ] Verify PR builds are unaffected (no release created, Auto-Build version format)
- [ ] Verify manual `release.yml` Edge and Stable releases still work

### DIFF
--- a/.github/workflows/auto_build.yml
+++ b/.github/workflows/auto_build.yml
@@ -1,5 +1,5 @@
 # Auto build on push to main and PR.
-# Not recommended to release these builds due to versioning.
+# Pushes to main also create an Edge release on GitHub.
 
 name: Auto Build
 on:
@@ -78,7 +78,8 @@ jobs:
   pre-build:
     uses: ./.github/workflows/get_version_info.yml
     with:
-      format_only: true
+      type: ${{ github.event_name == 'push' && 'Edge' || 'Auto-Build' }}
+      format_only: ${{ github.event_name != 'push' }}
 
   attested-build:
     needs: [detect-changes, pre-build]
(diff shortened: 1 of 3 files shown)


LionelColaso (changes requested):
If there are multiple PRs being merged, then there would be edge releases being created multiple times a day. 
This would be confusing to users; you could rather manually trigger Edge Builds using an edge branch https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs

cebarks:
> If there are multiple PRs being merged, then there would be edge releases being created multiple times a day. This would be confusing to users; you could rather manually trigger Edge Builds using an edge branch https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs

These edge releases are marked as pre-release and have a disclaimer that it is built from the main branch and potentially unstable. It also deletes the previous edge build while creating the new one, so there will only ever be one auto-built edge release at a time.

LionelColaso (approved):
Looks Good

PR #2106, opened 2026-06-07
fix: macOS self-update broken by nested .app bundle and wrong arch detection

cebarks:
## Summary

- **Nested `.app` bundle**: After extraction, `update.sh` rsync'd the temp directory (containing `RimSort.app/`) into the target `RimSort.app/`, producing `RimSort.app/RimSort.app/Contents/...` — an invalid bundle that macOS won't recognize. Added a `.app` subdirectory check so rsync copies bundle *contents* instead.
- **Wrong architecture download**: `platform.architecture()[0]` returns `"64bit"` on both Intel and ARM Macs, so arch-specific asset matching always failed and fell back to whichever Darwin asset appeared first in the API response. Now uses `platform.machine()` on macOS and updated patterns to match the actual asset naming convention (`arm`/`i386` from `distribute.py`).

## Test plan

- [x] `just check` — all quality gates pass (ruff, mypy, pyright, jscpd, shfmt, markdownlint)
- [x] `just test` — 758 tests pass
- [x] Verify the updated app launches correctly after update (no nested `.app` structure)

### DIFF
--- a/app/utils/update_utils.py
+++ b/app/utils/update_utils.py
@@ -420,8 +420,8 @@ class UpdateManager(QObject):
         "Darwin": {
             "patterns": ["Darwin", "macOS", "Mac"],
             "arch_patterns": {
-                "64bit": ["x86_64", "intel"],
-                "ARM64": ["arm64", "apple"],
+                "arm64": ["arm64", "arm", "apple"],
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Looks Good

PR #2126, opened 2026-06-10
fix: honor exclude options in file search (Skip translations, .git, Source, Textures)

### LINKED ISSUE #913: Add one or two settings — Only search files matching current game version and/or language
### Origin of idea

When searching files in projects that support multiple game versions and multiple languages, search results often include irrelevant files from other versions or languages. This makes it harder to locate the correct files, increases noise in search results, and may lead to accidental edits in the wrong version or language folder. Restricting search to only the current game version and language would help reduce mistakes and improve efficiency, especially for large modding projects.

Add a new setting to file search that allows restricting searches to:

- Only the folders matching the current game version.

- Only the folders matching the current game language.

When enabled, the search will automatically exclude files that belong to other versions or languages, based on the current game configuration or user-defined settings.

The current game version can be directly read from the `Version.txt` file located in the game installation directory.
The current game language can be read from the `langFolderName` key inside the `Prefs.xml` file located in the game's configuration directory.
These files provide reliable and automatic detection, enabling the search feature to filter files accordingly without manual input.

### Alternatives considered

No. The current search results often include too many irrelevant files, which leads to confusion and makes it difficult to locate the correct files efficiently.

### Additional context

This feature would be especially helpful for mod developers who maintain content for multiple game versions and languages simultaneously. It would reduce the risk of accidentally editing outdated or irrelevant files, and streamline the workflow when working with large mod projects where the file structure includes many versioned folders and language subdirectories.

Additionally, restricting the search scope may slightly reduce the performance cost of searches, even though the benefit may be relatively minor.

**Note:** Some language folders may have aliases. For example, for Chinese, both `ChineseSimplified` and `ChineseSimplified (简体中文)` can exist as valid folder names. The search filtering should take these variations into account.

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

cebarks:
## Summary

- Fix broken "Skip translations", "Skip .git folder", "Skip Source folder", and "Skip Textures folder" toggles in the file search dialog — the UI collected preferences correctly but `FileSearch._generic_search()` ignored `exclude_options` entirely
- Add directory pruning via `os.walk()` in-place `dirnames` modification so excluded directories are never descended into
- Remove dead `_should_exclude()` method from `SearchWorker` (was defined but never called)

## Details

The exclude toggles in the file search dialog were wired up correctly in the view (`get_search_options()` builds `exclude_options` dict) and passed through the controller to the search worker, but `_generic_search()` in `app/utils/file_search.py` discarded directory names from `os.walk()` (used `_` instead of `dirnames`) and never checked `exclude_options`.

The fix builds a set of excluded directory names once before the walk loop, then prunes them in-place with `dirnames[:] = [d for d in dirnames if d not in excluded_dirs]` — the standard Python idiom for preventing `os.walk` from descending into subdirectories. This is both correct and more performant than post-filtering results.

Closes #913

## Test plan

- [x] 3 new tests added covering all exclude options across `xml_search` and `standard_search` paths
- [x] 1029/1029 tests pass
- [x] Manual smoke test: toggling "Skip translations" on/off correctly shows/hides `Languages/` results

### DIFF
--- a/app/controllers/file_search_controller.py
+++ b/app/controllers/file_search_controller.py
@@ -517,28 +517,6 @@ class SearchWorker(QThread):
             scope == "inactive mods" and not is_active
         )
 
-    def _should_exclude(self, file_path: str) -> bool:
-        """Check if a file or directory should be excluded based on exclude_options."""
-        exclude_options = self.options.get("exclude_options", {})
-
-        # Skip translations
-        if exclude_options.get("skip_translations") and "Languages" in file_path:
-            return True
-
-        # Skip .git folders
-        if exclude_options.get("skip_git") and ".git" in file_path:
-            return True
-
-        # Skip Source folders
-        if exclude_options.get("skip_source") and "Source" in file_path:
-            return True
-
-        # Skip Textures folders
-        if exclude_options.get("skip_textures") and "Textures" in file_path:
-            return True
-
-        return False
-
     def run(self) -> None:
         try:
             logger.info(f"Starting search with text: {self.pattern}")
--- a/app/utils/file_search.py
+++ b/app/utils/file_search.py
@@ -86,8 +86,21 @@ class FileSearch:
         preview = options.get("preview", False)
         return_dict = options.get("return_dict", False)
 
+        exclude_options = options.get("exclude_options", {})
+        excluded_dirs: set[str] = set()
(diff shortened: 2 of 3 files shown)


LionelColaso (approved):
Ohh Nice catch

PR #2127, opened 2026-06-10
feat: bidirectional incompatibility flagging (#996)

### LINKED ISSUE #996: Incompatible mods flag them all
### Origin of idea

When a mod have an incompatibility rule, only the mod that has the rule is flagged with the red 'error', so only the one that has the rule appear when filtering for the error. Making it more difficult when a user wants to remove the other incompatible mods, and not the one that has the attached rule.

Flag all mentioned incompatible mods. So all of them appear when the user click the 'error' filter so they can be freely removed from the active list.

For example
Regrowth core is incompatible with Animated Marshes mod
Both of them should be flagged with an error, so the user could choose which one to keep. 

Potentially, the message could be different, to make it clear the flag was done due to the other mod rule. In the mentioned case, the tooltip in Regrowth stays the same, while Animated Marshes should state something in the lines of "Incompatible as per their rules: Regrowth core"

### Alternatives considered

_No response_

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

cebarks:
## Summary

Closes #996

When a mod declares an incompatibility rule, **both** mods are now flagged with an error — not just the one that declared the rule. This makes it easy to find and remove either incompatible mod using the error filter.

- **Bidirectional incompatibilities**: If Mod A declares incompatibility with Mod B, Mod B also gets flagged as incompatible with Mod A (in both the old `MetadataManager` and new `CompiledDependencyData` systems)
- **Tooltip differentiation**: The declaring mod shows "Incompatibilities: [other mod]" while the non-declaring mod shows "Incompatible (per other mod's rules): [declaring mod]" — so users can tell which mod owns the rule
- **Direction tracking**: A `declared_incompatibilities` field tracks what each mod explicitly declared, enabling the tooltip split. Mutual declarations (both mods declare each other) correctly show both under "Incompatibilities:"
- **Performance**: `add_incompatibility_to_mod()` now takes `packageid_to_uuids` as a parameter (matching `add_load_rule_to_mod` pattern) instead of rebuilding the lookup on every call

### Files changed

| File | What changed |
|------|-------------|
| `app/models/metadata/metadata_structure.py` | Added `declared_incompatibilities` field + reverse entry in `CompiledDependencyData.build()` |
| `app/utils/metadata.py` | Rewrote `add_incompatibility_to_mod()` with reverse logic + direction tracking + `packageid_to_uuids` param; updated 4 callsites |
| `app/views/mods_panel.py` | `_check_incompatibilities()` returns `(declared, reverse_only)` tuple; new tooltip section; reverse entries count as errors |

## Test plan

- [x] 12 new tests covering bidirectional behavior, direction tracking, mutual declarations, nonexistent mods, and parity invariants
- [x] All 1038 tests pass
- [x] ruff, ruff-format, mypy clean
- [x] Manual smoke test: both mods flagged, tooltips differentiated, error filter shows both
- [x] Verify with known incompatible mod pair (e.g., Regrowth Core + Animated Marshes)

### DIFF
--- a/app/models/metadata/metadata_structure.py
+++ b/app/models/metadata/metadata_structure.py
@@ -501,6 +501,7 @@ class CompiledDependencyData:
     tier_one_mods: set[str] = field(default_factory=set)
     tier_three_mods: set[str] = field(default_factory=set)
     incompatibilities: dict[str, set[str]] = field(default_factory=dict)
+    declared_incompatibilities: dict[str, set[str]] = field(default_factory=dict)
 
     @classmethod
     def build(
@@ -562,6 +563,8 @@ class CompiledDependencyData:
                 if incompat not in all_package_ids:
                     continue
                 compiled.incompatibilities.setdefault(pid, set()).add(incompat)
+                compiled.incompatibilities.setdefault(incompat, set()).add(pid)
+                compiled.declared_incompatibilities.setdefault(pid, set()).add(incompat)
 
             if rules.load_first and pid not in compiled.tier_zero_mods:
                 compiled.tier_one_mods.add(pid)
--- a/app/utils/metadata.py
+++ b/app/utils/metadata.py
@@ -740,6 +740,7 @@ class MetadataManager(QObject):
                         self.internal_local_metadata[uuid],
                         matched_versioned_incompat,
                         self.internal_local_metadata,
+                        self.packageid_to_uuids,
                     )
                 else:
                     logger.debug(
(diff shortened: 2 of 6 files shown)


LionelColaso:
This is awesome; hope it works

cebarks:
> This is awesome; hope it works

Just finished testing, worked great with Combat Expanded as a test mod!

LionelColaso (approved):
Cool

PR #2129, opened 2026-06-10
feat: single-instance warning dialog

### LINKED ISSUE #588: Suggestion: Warning dialog when the user tries to open RimSort when it is already open
Currently, due to shared resource limitations, RimSort can only be opened once. This is fine operationally, but in this case, RimSort just crashes very early on when loading the app info. It would be better if, instead, a dialog pops up to warn the user that RimSort is already running.

cebarks:
## Summary

- Adds a `SingleInstanceLock` class wrapping Qt's `QLockFile` for cross-platform lock file management with stale lock recovery
- Integrates lock into `__main__.py` startup sequence — shows a warning dialog when another instance is already running instead of crashing on SQLite contention
- Registers a SIGTERM handler for clean lock release on Unix; degrades gracefully if lock mechanism fails

Closes #588

## Test plan

- [x] Unit tests for `SingleInstanceLock`: acquire, release, contention, stale recovery, safe no-op release (5 tests)
- [x] Full test suite passes (1017 tests)
- [x] Linting clean (ruff, ruff-format, mypy)
- [x] Manual smoke test: first instance launches normally, second instance shows warning dialog, lock file cleaned up on exit, stale lock recovered after kill

### DIFF
--- a/app/__main__.py
+++ b/app/__main__.py
@@ -28,10 +28,8 @@
 
 import os
 import platform
-import signal
 import sys
 import traceback
-import types
 from logging import WARNING, getLogger
 from multiprocessing import freeze_support, set_start_method
 from types import TracebackType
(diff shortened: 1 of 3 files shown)


LionelColaso (approved):
Looks Good

PR #2162, opened 2026-06-13
feat: add Stable/Beta/Edge update streams

cebarks:
## Summary

- Adds opt-in update streams (Stable / Beta / Edge) so users can choose their update cadence and risk tolerance
- Switches edge builds from push-triggered to a 12-hour scheduled cron, skipping builds when there are no new commits
- Adds Beta release type to CI workflows for manually dispatched release candidates
- Update checker now queries `/releases` (instead of `/releases/latest`) and filters by the user's selected stream, with fallback chain: edge → beta → stable
- Downgrade handling: when switching to a less-bleeding-edge stream, prompts the user before installing an older version
- Version badge (`[BETA]` / `[EDGE]`) displayed next to the game version at the bottom of the main window

## Changes

| Area | Files | What |
|------|-------|------|
| Settings model | `app/models/settings.py` | `update_stream: str = "stable"` |
| Settings UI | `settings_dialog.py`, `advanced_tab_controller.py` | Update Channel dropdown in Advanced tab with escalation warning |
| Update checker | `app/utils/update_utils.py` | `filter_releases_by_stream()`, `extract_version_from_edge_release()` (normalizes `-edge` → `.dev` for PEP 440), stream-aware release lookup, downgrade dialog |
| Version display | `main_window.py`, `main_window_controller.py` | RimSort version label with stream badge |
| CI - Edge | `auto_build.yml` | 12h cron schedule, `check-edge-needed` job, schedule-aware change detection |
| CI - Beta | `get_version_info.yml`, `release.yml` | Beta version format, release type option |
| Tests | `tests/utils/test_update_streams.py` | 16 tests: stream filtering, edge version extraction, downgrade outcomes |

## Design spec

`docs/superpowers/specs/2026-06-13-update-streams-design.md`

## Test plan

- [x] 1070 tests passing (16 new)
- [x] ruff lint clean
- [x] ruff format clean
- [x] mypy clean
- [x] Manual UI verification: dropdown appears in Advanced tab, escalation warnings show/revert correctly, version badge updates live
- [ ] Verify edge cron schedule fires correctly (requires merge to main)
- [ ] Test Beta release dispatch via workflow_dispatch

### DIFF
--- a/.github/workflows/auto_build.yml
+++ b/.github/workflows/auto_build.yml
@@ -1,8 +1,10 @@
 # Auto build on push to main and PR.
-# Pushes to main also create an Edge release on GitHub.
+# Scheduled builds every 12 hours create Edge releases on GitHub.
 
 name: Auto Build
 on:
+  schedule:
+    - cron: '0 */12 * * *'
   pull_request:
     types: [opened, synchronize]
   push:
@@ -44,6 +46,12 @@ jobs:
             CHANGED=$(git diff --name-only "$PR_BASE"..."$PR_HEAD")
           elif [[ "$GH_EVENT" == "merge_group" ]]; then
             CHANGED=$(git diff --name-only "$MG_BASE"..."$MG_HEAD")
+          elif [[ "$GH_EVENT" == "schedule" ]]; then
+            # For scheduled builds, always consider app changed
+            # (check-edge-needed job gates whether we actually build)
+            echo "app_changed=true" >> "$GITHUB_OUTPUT"
+            echo "Scheduled build — skipping change detection"
+            exit 0
           else
             CHANGED=$(git diff --name-only "$PUSH_BEFORE" "$PUSH_SHA")
           fi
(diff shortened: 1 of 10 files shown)


LionelColaso:
Intersesting But,
Why do we need 3 channels? Just Stable and Edge is Enough. We could rename EDGE to BETA
We need to add proper support in the updater to support EDGE/BETA updates. and also so that the user can can upgrade or downgrade to stable builds as well safely if possible
Mostly RimSort is backward compatible unless there is any drastic change to settings, for example, the case of run_arguments.

cebarks:
My thought behind having beta and edge differentiated is:

Edge: for users to be able to test specific fixes or get bleeding-edge features if they choose to (making it easier for new bugs to be caught earlier)

Beta: essentially would be release candidate builds so users who are willing to help test for bugs before new stable versions are actually released can.

This whole idea came from the last release or two where bugs made it into stable versions that a significant amount of users upgrade to could have been caught in a beta version instead. even if 1% of users use the beta builds and report issues, that would drastically increase the number of people testing new features before they are released.

PR #2151, opened 2026-06-11
refactor(metadata): migrate all consumers to MetadataController and remove old metadata system (#2051 PR 3)

cebarks:
## Summary

Migrate all remaining MetadataManager consumers to MetadataController, completing the metadata migration (#2051). After this PR, MetadataManager is deleted — no PR 4 needed.

**Three mechanical transformations applied across 50 files:**

- **UUID → PATH identity** — `.uuid` → `.path`, `.uuids` → `.paths` throughout widget system; `_get_mod_path()` deleted; mapper dicts eliminated
- **dict[str, Any] → typed ListedMod/AboutXmlMod** — all `.get("key")` / `["key"]` access replaced with typed property access + `isinstance` guards
- **MetadataManager.instance() → MetadataController.instance()** — singleton swap; `refresh_cache()` → `refresh_metadata()`, `compile_metadata()` → `compile()`, `_abort_requested` → `is_abort_requested`

**Additional changes:**
- `published_file_id`: `int` with `-1` sentinel → `str | None`
- `SteamDbEntry.tags` field added (was silently dropped by msgspec)
- `DividerData.path` alias property for uniform access in mixed lists
- Watchdog fully rewritten — no more UUID generation, signals simplified from 3/5-param to 2-param
- Sort functions renamed `uuid_to_*` → `path_to_*` with typed access
- `ModInfo.from_listed_mod()` factory for typed mod conversion
- MetadataController gains `process_creation`/`process_deletion`/`process_update` + `is_abort_requested` + `update_workshop_timestamps()`
- MetadataManager and `app/utils/metadata.py` fully deleted

**Bug fixes:**
- Fix UTF-8 BOM in `PublishedFileId.txt` causing mod parsing failures (#2150)
- Fix Use This Instead database never loading (wrong path resolution, wrong data format, wrong lookup key)
- Fix Use This Instead panel crashes (`AboutXmlMod` has no `.copy()`, replacement metadata missing `packageid`)
- Fix crash when clicking invalid mods (no `About.xml`) in mod info panel
- Fix `local_mods_path` derivation when not explicitly set (#2154)
- Fix rule editor crash — typed access prevents `'dict' object has no attribute 'lower'` (#2130)
- Fix mods disappearing from active list on sort — dual metadata system mismatch eliminated (#2150)
- Fix `modDependenciesByVersion` not flowing through to dependency checks and sorting (#2095)
- Fix #2041 structurally (constants never mutated by sort functions)

## Test plan

[...]

PR #2163, opened 2026-06-13
fix(metadata): don't crash on startup when local mods folder is unset

cebarks:
## Summary

- Fixes fatal `ValueError: Essential paths are missing, invalid, or not directories` crash on startup when the local mods folder path is not configured
- The UI guard (`check_if_essential_paths_are_set`) only validates `game_folder` + `config_folder`, but `MetadataMediator.refresh_metadata` also required `local_mods_path` — which defaults to `""` in the Instance model and becomes `None` after path conversion
- Makes `local_mods_path` optional throughout the metadata pipeline (`MetadataMediator`, `_ParserWorker`, `create_listed_mod_from_path`, `_set_mod_type`) since it is not truly essential — the mod search loop already skipped `None` paths, and the type classification comparison handles `None` safely

## Test plan

- [x] `just typecheck` — 0 issues
- [x] `just pyright` — 0 errors
- [x] `just test` — 1054 passed

### DIFF
--- a/app/models/metadata/metadata_factory.py
+++ b/app/models/metadata/metadata_factory.py
@@ -362,7 +362,7 @@ def _parse_optional(
 
 
 def _set_mod_type(
-    mod: ListedMod, local_path: Path, rimworld_path: Path, workshop_path: Path | None
+    mod: ListedMod, local_path: Path | None, rimworld_path: Path, workshop_path: Path | None
 ) -> ListedMod:
     """
     Set the mod type based on the paths given.
@@ -715,7 +715,7 @@ def create_rules_from_external_rules(external_rule: ExternalRule) -> Rules:
 def create_listed_mod_from_path(
     path: Path,
     target_version: str,
-    local_path: Path,
+    local_path: Path | None,
     rimworld_path: Path,
     workshop_path: Path | None,
     prefer_versioned: bool = True,
--- a/app/models/metadata/metadata_factory.py
+++ b/app/models/metadata/metadata_factory.py
@@ -362,7 +362,10 @@ def _parse_optional(
 
 
 def _set_mod_type(
-    mod: ListedMod, local_path: Path | None, rimworld_path: Path, workshop_path: Path | None
+    mod: ListedMod,
+    local_path: Path | None,
+    rimworld_path: Path,
+    workshop_path: Path | None,
 ) -> ListedMod:
     """
     Set the mod type based on the paths given.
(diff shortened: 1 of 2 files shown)


LionelColaso:
This is not right; the local mods folder is automatically set when the game folder is selected unless the user manually cleared it. 
Check [_on_game_choose.](https://github.com/RimSort/RimSort/blob/main/app/controllers/settings_tabs/locations_tab_controller.py#L163)
The solution would be to check if game paths are set; if they are set, then add the mods to the game path. To avoid the crash 
The local mod's path should never be NONE. If game paths are set

LionelColaso (changes requested):
see comments

cebarks:
> This is not right; the local mods folder is automatically set when the game folder is selected unless the user manually cleared it. Check [_on_game_choose.](https://github.com/RimSort/RimSort/blob/main/app/controllers/settings_tabs/locations_tab_controller.py#L163) The solution would be to check if game paths are set; if they are set, then add the mods to the game path. To avoid the crash The local mod's path should never be NONE. If game paths are set

ahh gotcha, it seemed like a common issue with people so i assumed it didn't have to be set if so many people had configs with it missing

LionelColaso:
> > This is not right; the local mods folder is automatically set when the game folder is selected unless the user manually cleared it. Check [_on_game_choose.](https://github.com/RimSort/RimSort/blob/main/app/controllers/settings_tabs/locations_tab_controller.py#L163) The solution would be to check if game paths are set; if they are set, then add the mods to the game path. To avoid the crash The local mod's path should never be NONE. If game paths are set
> 
> ahh gotcha, it seemed like a common issue with people so i assumed it didn't have to be set if so many people had configs with it missing

If you check all location paths are validated before settings can be saved
just did not add a validation for mods folder since i made it to set automatically using game folder since i never thought user would clear it after its set should probabably add validation for it as well

cebarks:
> If you check all location paths are validated before settings can be saved just did not add a validation for mods folder since i made it to set automatically using game folder since i never thought user would clear it after its set should probabably add validation for it as well

Ah makes sense, I think this fix should probably be merged into #2151 then since it'll need to be fixed in the new metadata system anyway.

LionelColaso:
> > If you check all location paths are validated before settings can be saved just did not add a validation for mods folder since i made it to set automatically using game folder since i never thought user would clear it after its set should probabably add validation for it as well
> 
> Ah makes sense, I think this fix should probably be merged into #2151 then since it'll need to be fixed in the new metadata system anyway.

What ever you prefer

cebarks:
closing in favor of #2151

PR #2165, opened 2026-06-13
Add ability to sort inactive mods by the date they were updated on the Steam Workshop

### LINKED ISSUE #2153: Allow sorting mods by when they were updated on Steam
### Origin of idea

As Steam doesn't always updated the modified time of a folder when updating a mod (at least on Windows), it would be great if we could sort mods based on when they were updated on Steam.

Add sorting option to sort by the date displayed as Workshop Times/Steam Updated.
I've already implemented this at https://github.com/tobi1449/RimSort/tree/sortby-mod-updated
If accepted, I'll open a PR.

### Alternatives considered

_No response_

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

tobi1449:
Add ability to sort inactive mods by the date they were updated on the Steam Workshop

Closes #2153

### DIFF
--- a/app/sort/mod_sorting.py
+++ b/app/sort/mod_sorting.py
@@ -287,7 +287,7 @@ def uuid_to_mod_updated(
     else:
         metadata = get_mod_metadata(uuid)
 
-    return metadata.get("internal_time_updated") if metadata else None
+    return metadata.get("internal_time_updated") if metadata else 0
 
 
 def get_dir_size(path: str) -> int:
--- a/app/sort/mod_sorting.py
+++ b/app/sort/mod_sorting.py
@@ -287,7 +287,7 @@ def uuid_to_mod_updated(
     else:
         metadata = get_mod_metadata(uuid)
 
-    return metadata.get("internal_time_updated") if metadata else 0
+    return metadata.get("internal_time_updated", 0) if metadata else 0
 
 
 def get_dir_size(path: str) -> int:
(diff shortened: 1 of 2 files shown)


cebarks (changes requested):
Maybe just change the sort text from "Mod Updated" to something to reflect it's based on the workshop last updated date, maybe something like: "Workshop Updated" (open to suggestions here)

cebarks:
Also a few linting errors that should be fixed (don't worry about the codecov report)

app/sort/mod_sorting.py
```diff
@@ -266,6 +266,30 @@ def uuid_to_mod_tags(
...
+
+    Returns:
+        Update timestamp as int, or 0 if not available
+    """
+    if cached_metadata is not None:
+        metadata = cached_metadata.get(uuid)
+    else:
+        metadata = get_mod_metadata(uuid)
+
+    return metadata.get("internal_time_updated") if metadata else 0
```
cebarks: `dict.get(key)` without a default returns `V | None`, so pyright sees the return type as `Any | Literal[0] | None` which isn't assignable to `int`. Adding a default of `0` eliminates the `None` branch:

```suggestion
    return metadata.get("internal_time_updated", 0) if metadata else 0
```

cebarks (commented):
Thanks for the contribution! Just one small pyright fix needed — see the inline suggestion below.

EDIT: just went ahead and resolved it since it was quite simple.

cebarks (approved):

tobi1449:
> Thanks for the contribution! Just one small pyright fix needed — see the inline suggestion below.
> 
> EDIT: just went ahead and resolved it since it was quite simple.

Thanks, for some reason this didn't show up with "just check".

PR #2171, opened 2026-06-14
tags: feat: Rework TagEditDialog for easier tag editing/creation

### LINKED ISSUE #2172: tags: Improve tag editing/creation workflow in the popup dialog
### Origin of idea

Tag editing/creation requires remembering all tags precisely to find the desired tag to set for a new mod, and one cannot assign tags by just typing. The checkbox is unnecessarily small and hard to click. The UX is not exactly ergonomic, as one has to swap between the mouse and the keyboard.

Refactor the tag editing pop-up window to enable easier tag creation/editing via the keyboard, filtering of the tag list, and easier tag assignment.

### Alternatives considered

_No response_

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

Adda0:
Reworks `TagEditDialog` to support the following, allowing for easier tag creation and selection.

- Filtering existing tags (by writing into the text field area).
- Quickly creating new tags (typing the tag name, creation upon pressing Enter or comma keys), while still allowing filtering on next tags (clearing the text input area after creation).
- Pressing down arrow when the text area is focused moves the cursor to the filtered list of tags, allowing selection by pressing spacebar. Shift-Tab moves back to the text area.
- The tag list is using multiselection and the `QListWidget` selection mechanism instead of the previous checkboxes for easier selection (one does not have to hunt for the small checkbox with their cursor; clicking anywhere on the line with the tag will suffice).
- The list of tags is automatically sorted; there is no need for manual sorting in functions.

From my testing, everything seems to work.

Fixes #2172. Apologies for not waiting for discussion on the FR before implementing the changes. I already made the changes for myself. Feel free to modify/close the PR as needed.


cebarks (changes requested):
Overall great UX improvement, thanks!

A couple things before merging:
- down arrow while the text area is selected does not work as described while i was testing (I'm on macos currently)
- some unit tests for `upsert_typed_tag()` and `filter_tags_list` would be awesome if you're up to it, otherwise I'll try and cover them in a later testing improvement pass.

Personally I'd love to see more of these kinds of UX improvements, if you're up to doing them for other parts of the app!

Feel free to hit me up on Discord if you'd like to help and need any suggestions 😄

Adda0:
Heya.

> down arrow while the text area is selected does not work as described while i was testing (I'm on macos currently)

Hmm, I am on Linux, and it works just fine. Will have a look and try to find what could cause it.

> some unit tests for upsert_typed_tag() and filter_tags_list would be awesome if you're up to it, otherwise I'll try and cover them in a later testing improvement pass.

Will try to come up with something.

cebarks:
Maybe on macOS the down arrow key isn't the same as `Qt.Key.Key_Down` on other platforms?

Adda0:
I added tests for all the new functions and tried to fix the macOS issue. I cannot verify that the fixes are working because I do not have a Mac. I tried to be firmer in forcing the focus on the correct dialog element. Also, to improve the workflow, I modified the behaviour of the Enter and Return keys so that when the tag input text field is focused, Enter and Return upsert the typed text, and when the tags list is focused, they toggle the currently focused tag item. Alt-{Enter,Return} accept the changes made in the dialog. Also rebased on the latest master.

Adda0:
(Fixing the linting issues.)

PS: The lint action gives quite a few errors in its output for Python tool runs. Might be worth checking out.

Adda0:
I think the remaining linting errors stem from the JSCPD check, which found some duplication in the tests. However, I believe the errors are invalid, since the duplication checks the same issues, but in different contexts, which is useful.

cebarks:
JSCPD can be too picky sometimes IMO, go ahead and wrap the duplicates like this:
```
  # jscpd:ignore-start
  # duplicated code here
  # jscpd:ignore-end
```

I have a feeling there's some weird macos-specific platform issue with the down arrow code so just file an issue to track it and we can fix it later.

Adda0:
> JSCPD can be too picky sometimes IMO, go ahead and wrap the duplicates like this:

OK, I will do so.

> I have a feeling there's some weird macos-specific platform issue with the down arrow code so just file an issue to track it and we can fix it later.

So still not fixed? Hmm, does it at least work, for example, for Tab switching focus from the text field and Alt-Enter accepting the changes? To see whether the key presses are captured at all on macOS.

cebarks:
> So still not fixed? Hmm, does it at least work, for example, for Tab switching focus from the text field and Alt-Enter accepting the changes? To see whether the key presses are captured at all on macOS.

tab and alt-enter work, as well as the comma, return, space upserting, it's literally only the arrow key :upside_down_face: 

i'll do a little testing and see if i can fix it on my macbook and if not i'll go ahead and merge this as is (once the JSCPD ignores are added).

cebarks (approved):

Adda0:
Added the ignores. It really looks like the key detection is thus working. Possibly, there are then issues with changing the focus on macOS, but I do not know. It all works fine on Linux. If you could debug it yourself, that would be best. I cannot try anything without having you as the test runner :wink: To get a clean slate for experimenting where everything already works on Linux, all configuration related to FocusPolicy and FocusReason can be removed (but not the setFocus() functions themselves).

cebarks:
OK i got it working with an Event Filter! It still worked on linux from my testing as well.

Adda0:
Heh, the down key now stopped working for me on the latest master. Everything else still works, even the new up arrow (nice idea). Will create a new PR with a fix that seems to work for me.

cebarks:
god damn it oops, i should have waited. what DE are you on?

Adda0:
Fix for Linux in https://github.com/RimSort/RimSort/pull/2187. It is an issue with operator precedence. How it could work for you, I cannot tell.

cebarks:
I'm not sure either lol, must have been some file cache somewhere

Adda0:
Hey. So, I updated my RimSort to 1.8.0 from my PR version from here, and the changes done in this PR disappeared (the tag edit dialog keyboard events, the removal of the small checkboxes and its replacement with the MultiSelection mode, the tests etc.; not sure which parts exactly were removed). I *think* they were removed in https://github.com/RimSort/RimSort/commit/1928f297a19d99670b2ad7d7a1035cf2dd2c7680, maybe other commits as well. And I presume that is why [`cc5fdcf` (#2151)](https://github.com/RimSort/RimSort/pull/2151/changes/cc5fdcfa604e7b12789491ec58475b2b50ae9295) was made, but actually only fixes the broken remains of this PR that were not removed previously. Was that intentional?

cebarks:
Absolutely not intentional, my bad. I'll fix it now

PR #2178, opened 2026-06-15
fix(ddsutility): File Deletion path can be mangled in DDSUtility if the path contains `.dds` outside of the file name

tomaioo:
## Summary

Security: Unsafe File Deletion in DDSUtility

## Problem

**Severity**: `Medium` | **File**: `app/utils/dds_utility.py:L30`

The DDSUtility.delete_dds_files_without_png method deletes files based on path construction using string replacement (.dds to .png). This could potentially delete wrong files if paths contain '.dds' in unexpected places. More critically, it performs file deletion without proper validation of the file's location or ownership, which could be exploited if paths are manipulated.

## Solution

Use proper path manipulation with Pathlib instead of string replacement. Validate that files are within expected directories before deletion. Consider adding confirmation dialogs before mass deletion operations.

## Changes

- `app/utils/dds_utility.py` (modified)

### DIFF
--- a/app/utils/dds_utility.py
+++ b/app/utils/dds_utility.py
@@ -1,5 +1,6 @@
 import os
 from glob import glob
+from pathlib import Path
 
 from loguru import logger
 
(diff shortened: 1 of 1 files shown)


cebarks (changes requested):
Definitely an improvement for an unlikely but technically possible bug, but definitely not a security vulnerability. All paths that would be passed to this function are user-controlled and inputted. 

If you're already swapping part of this to `pathlib`, please go all the way and replace the `os.path` uses in lines 33-38 as well.

PR #2276, opened 2026-06-28
Add delete button to workshop mod updater panel

### LINKED ISSUE #2266: Add a delete button on the mod update screen
### Origin of idea

"Update mods via steam" never works properly for me, it either does nothing or only updates a single mod, making me do the process manually.

Simply having a button to "Delete mod completely" or Subscribe/Unsubscribe on Steam would allow users to delete the mod(s) and Verify Game Files on Steam to force an update, greatly reducing cortisol levels.

### Alternatives considered

_No response_

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

LionelColaso:
Adds a delete button to WorkshopModUpdaterPanel so users can delete mods that fail to update properly via Steam, then force Steam to re-download them.

The button provides:
- Delete mod completely
- Delete mod and resubscribe using Steam (when Steam integration enabled)
- Delete mod and unsubscribe from Steam (when Steam integration enabled)

Table auto-refreshes after deletion via the standard _refresh_metadata_and_panel callback. The key is already set correctly in the existing ModInfo.from_metadata(uuid, ...) call, so the deletion menu can properly look up selected mods.

Closes #2266

### DIFF
--- a/app/windows/workshop_mod_updater_panel.py
+++ b/app/windows/workshop_mod_updater_panel.py
@@ -75,6 +75,11 @@ class WorkshopModUpdaterPanel(BaseModsPanel):
                 )
             )
 
+        # Add delete button
+        button_configs.append(
+            self._create_delete_button_config(self.tr("Delete Selected Mods"))
+        )
+
         # Set up buttons based on configurations
         self._setup_buttons_from_config(button_configs)
 


hawbaudah:
you're amazing

PR #1841, opened 2026-03-17
Harden network requests

cebarks:
### HTTP request utility
- Create `app/utils/http.py` with `get()`/`post()`/`head()` wrapper functions that apply a default 15s timeout
- Refactor all 15 `requests.get/post/head` call sites in `app/` to use the shared utility
- Use `timeout=(5, 60)` for batch Steam API calls (up to 5000 IDs per chunk) — fast connect, slow read
- Add timeout to `HttpClient.make_request()` in rentry wrapper
- Add `timeout=15` to `distribute.py` directly (can't import from `app/`)

### URL validation
- Fix Rentry URL validation: use `urlparse()` hostname comparison instead of `startswith()` which matched `rentry.co.evil.com`
- Fix Steam collection URL validation: same `startswith()` subdomain spoofing bug
- Remove unnecessary `try/except` around `urlparse` (it never raises)

## Security Impact

**Severity: Medium** — 9 HTTP calls had no timeout, allowing indefinite application hangs. URL validation accepted spoofed subdomains. The `http.py` utility ensures no future HTTP call can accidentally omit a timeout, and provides a single point for adding retry logic later.

## Test plan

- [x] `tests/utils/test_http.py` — 6 tests passing
- [x] `tests/utils/test_url_validation.py` — 7 tests passing
- [ ] Verify Steam Workshop metadata queries still work
- [ ] Verify Rentry import still works
- [ ] Verify SteamCMD download still works
- [ ] Verify auto-update check still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2280, opened 2026-07-01
feat(http): add retry logic with backoff for transient failures

shuja-waraich-03:
Builds on #1841 (the `app/utils/http.py` timeout wrappers) to add automatic
retry-with-backoff for transient HTTP failures, as requested in #1845.

A `urllib3` `Retry` adapter is mounted on a fresh per-request `Session`, so
requests through `get()` / `post()` / `head()` now retry transient failures
automatically.

## Behavior

- **Retries:** 4 attempts, exponential backoff (factor 1).
- **Transient conditions:** connection errors + status codes `429, 500, 502, 503, 504`.
- **Idempotent-only:** retries limited to `GET`/`HEAD`; `POST` is not retried.
- **Streaming-safe:** `read=0` so partial reads on streaming downloads aren't re-attempted.
- **Rate limits:** honors `Retry-After` headers (`respect_retry_after_header=True`).
- **Unchanged contract:** `raise_on_status=False` keeps `response.raise_for_status()` behaving as before.
- **Thread-safe:** a fresh `Session` per request, so no state is shared across threads.

## Testing

`tests/utils/test_http.py` covers wrapper dispatch, default/explicit timeouts,
adapter mounting, the status forcelist, idempotent-only methods, no read
retries, `Retry-After` handling, and the non-raising contract. Locally: ruff,
ruff-format, mypy, and pyright all pass; 9 tests pass.

### DIFF
--- a/app/utils/http.py
+++ b/app/utils/http.py
@@ -1,33 +1,72 @@
-"""HTTP utility module providing request functions with default timeouts.
+"""HTTP utility module providing request functions with retries and timeouts.
 
 All HTTP requests in the application should use these functions instead of
 calling requests.get/post/head directly, to ensure consistent timeout behavior.
+
+Transient failures (connection errors and the status codes in
+RETRY_STATUS_CODES) are retried automatically for idempotent methods using
+exponential backoff. A fresh Session is created per request so no state is
+shared between threads.
 """
 
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Looks Good Thanks

Issue #2284, opened 2026-07-02
Mod Dividers Are Weird


neronix17:
### Release Type

Compiled (Windows x86-64)

### Version

1.7.3

### Operating System

Window 10

### Describe the bug

Mod dividers seem to be implemented using an index number and do not save that in mod lists, so loading a different mod list screws it all up. I'd previously been using empty mods as a divider and while the feature of having them collapsible and being visually distinct is nice it's pretty useless if you ever use the load/save mod lists functionality as it currently is, which I hope isn't the intended behaviour.

### Additional context

Reproducing:
- Create two mod lists with dividers, have one list be different numbers of mods within any given divider.
- Loading either list has the dividers in the same position they were last left in, when they should be above the mods they were when saving the list.

### Code of Conduct

I agree to follow this project's Code of Conduct

### Duplicate Issue Check

I've checked for similar issues and didn't find anything

### Wiki/FAQ Check

I've checked the Wiki for a solution and didn't find a solution

neronix17:
Further messing around I thought maybe these would work better with instances, nope, same issue, they are not saved as part of the instance either.

zohryu:
Yeah, same problem here.
Loading a saved modlist and the dividers are just emptied out and ... basically don't do anything after.

PR #2286, opened 2026-07-04
fix: clear stale dividers when importing mod lists

gingerninja85:
## Summary

Fixes stale active mod-list dividers being reused after importing a normal RimWorld XML mod list.

RimWorld XML mod lists only carry package IDs; they do not carry RimSort UI-only divider metadata. Before this change, importing a different XML list left `settings.active_mods_dividers` untouched, so the previous list's divider rows could be reinserted at old numeric positions in the newly imported list.

## What changed

- Clear `settings.active_mods_dividers` after a successful XML import and before inserting the imported list.
- Save settings immediately so stale divider state is not restored later.
- Add a regression test covering the stale-divider import case.

## Verification

```bash
QT_QPA_PLATFORM=offscreen PYTHONPATH=. python -m pytest \
  tests/views/test_main_content_divider_import.py \
  tests/views/test_main_content_run.py \
  tests/models/test_mod_list.py::TestModListConstruction::test_empty \
  -q -p no:xvfb
```

Result:

```text
6 passed
```

## Notes

This does not attempt to serialize dividers into RimWorld XML, because those files are package-ID mod lists and do not naturally preserve RimSort-specific UI divider metadata. This patch prevents the stale-state behavior described in #2284.


neronix17:
Since my issue is mentioned with this, I'll add that the stale state behaviour is preferable to having to create them all again. Both options suck but redoing them entirely every time is way worse.

LionelColaso:
@gingerninja85 This is Half Solution but solves one issue creating new issue but its a solution for now
Better solution would be to save the divider state somewhere and re add it after importing modlist but there will be additional logic that would need to be changed also making sure that it does not get added to the .xml etc which is bigger task

@neronix17 i will look into it can you create a new Issue with the solution you looking for so i or someone can look into it in their free time

Thanks

LionelColaso (approved):

neronix17:
I'd just copy and paste what I already said so no thanks. I've just went back to using custom empty mods as dividers, works 1000x better.

PR #2287, opened 2026-07-05
feat: add recently updated mods indicator

### LINKED ISSUE #2104: A warning for mods that recently updated may help a lot
### Origin of idea

Some newly updated mods just crashed my game, and I didn't know which one it was. 

Just like the title

### Alternatives considered

_No response_

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

carmoran0:
## Summary

Closes #2104.

When Steam auto-updates workshop mods, a freshly-updated mod can crash the game with no easy way to tell which one changed. The issue reporter had exactly this happen. This PR adds an **opt-in** indicator that flags Steam Workshop / SteamCMD mods whose workshop update time falls within a configurable window, so the likely culprits are visible at a glance.

## What it does

- **New settings** (Settings → Sorting → *Mod list options*): a **"Show recently updated mods indicator"** checkbox (off by default) and a **"Days to consider a mod recently updated"** spinbox (1–30, default 3, disabled while the checkbox is off).
- **Per-row icon** on affected mods in both the active and inactive lists, with a relative-time tooltip ("Updated 5 hours ago…"). Clicking it opens that mod's Steam Workshop changelog page.
- **"N updated" counter** in the active list's summary bar (next to "N new"); clicking filters the list to just those mods, mutually exclusive with the existing warnings/errors/new filters.

## How it works

- The "recently updated" timestamp prefers `acf_time_updated` (the update time of the content **actually installed on disk**, refreshed on every metadata refresh) and falls back to the Steam WebAPI `external_time_updated`. This is intentionally the reverse of `resolve_aux_timestamps`, since the WebAPI value can be newer than the installed content (an update is available but not downloaded) — a false positive for crash diagnosis.
- Only `STEAM_WORKSHOP` / `STEAM_CMD` mods are ever flagged; local / git / Ludeon mods are skipped.
- Reuses existing infrastructure: the aux metadata DB already stores these timestamps, and the per-item lookup piggybacks on the existing `CustomListWidgetItemMetadata` aux queries — gated on the setting, so it costs nothing when the feature is off. The window check in the recalculation loop is a pure integer comparison (no DB access).

## Notes for reviewers

- **Icon**: adds `themes/default-icons/updated.png` (icons8 "update" icon). I added a short **Credits** section to the README for attribution (icons8's free license requires a visible link) — happy to move it elsewhere if you'd prefer.
- **Translations**: only the 7 new strings were hand-added to `locales/en_US.ts`. I deliberately did **not** run a full `update-ts` regeneration, because the committed `en_US.ts` is already out of sync with the codebase and a wholesale regen swept in ~57 unrelated string changes from prior refactors. Worth a separate housekeeping pass.
- **Timing caveat**: `acf_time_updated` updates when RimSort refreshes its metadata, so a mod Steam auto-updates *while RimSort is open* only shows the icon after the next Refresh. This is mentioned in the setting's tooltip.

## Testing

- Added unit tests for the two pure helpers (`resolve_workshop_updated_timestamp`, `is_recently_updated`), the timestamp-resolution skip rule (`get_updated_timestamp_by_path` — Steam vs. local/git/Ludeon/missing/KeyError), the new settings defaults, and the sorting-tab controller round-trip.
- Full suite passes; `ruff` clean. The GUI-integration layer (icon visibility, counter, filter) is not unit-tested, consistent with the repo's existing coverage for list-widget classes.


carmoran0:
This is how it looks like. Let me know what do you all think
Captura de pantalla_20260705_153259
Captura de pantalla_20260705_153000

LionelColaso (approved):
Cool amd Awesome Thanks

PR #2289, opened 2026-07-06
Complete Spanish translations and fix i18n string-extraction bugs

carmoran0:
## Summary

While completing the Spanish (`es_ES`) translation, I found that `en_US.ts` was out of sync with the codebase in two ways that go beyond a normal missed-string gap:

- **`app/services/instance_service.py`** aliased `tr = QCoreApplication.translate`. `pyside6-lupdate`'s static extraction can't resolve that alias — it misreads `tr(context, text)` as `obj.tr(sourceText, disambiguation)` and swaps them, so ~53 user-facing strings in that file (instance backup/restore/clone/delete dialogs) were **never captured into any translation file, in any language**, since the file was added.
- **`app/utils/button_factory.py`** had `self.panel.tr(text)` calls where `text` was a variable, not a literal — also unextractable by lupdate. That put several already-translated button labels (Refresh, Select all, Deselect all, Subscribe/Unsubscribe selected, Download with SteamCMD) at risk of silently disappearing from the `.qm` the next time someone runs a full translation regen.

Fixing both surfaced a full, clean `pyside6-lupdate` diff for the first time, so this PR also completes the Spanish translation against the corrected source.

Separately, while manually testing the language-switch flow, restarting the app after a language change crashed with `ModuleNotFoundError: No module named 'app'`. That's a pre-existing bug in `restart_application()`: `python -m app` sets `sys.argv[0]` to the resolved path of `app/__main__.py`, not `-m app`, so the non-frozen restart path was re-launching that path as a plain script instead of as a module. Included the fix since it currently blocks verifying any translation change in a dev environment.

## What changed

1. **`fix(i18n)`**: replace the `tr` alias in `instance_service.py` and the indirect `self.panel.tr()` calls in `button_factory.py`/`use_this_instead_panel.py` with direct `QCoreApplication.translate(context, text)` calls at the literal call sites. Adds a regression test guarding against the `tr = QCoreApplication.translate` alias pattern reappearing anywhere in `app/`, plus button/label coverage for the affected call sites.
2. **`fix`**: `restart_application()` now reconstructs the `-m app` invocation explicitly for the non-frozen (dev) case instead of blindly reusing `sys.argv`.
3. **`i18n`**: regenerates `en_US.ts` (66 strings newly extractable/added) and translates all of them into Spanish, plus fixes two pre-existing broken translations found along the way where an earlier auto-translate pass had mangled a `{placeholder}`'s position, losing the surrounding text. `es_ES` is now 100% translated per `translation_helper.py check` and passes `translation_helper.py validate` with no placeholder/HTML tag mismatches. All 11 locale `.qm` files recompile cleanly with `just i18n-compile`.

## Test plan

- [x] `uv run pytest` — 1128 passed (1 pre-existing, unrelated failure: a symlink test that requires elevated Windows privileges, fails identically on a clean `main` checkout)
- [x] New tests: `tests/test_i18n_source_extraction.py`, `tests/utils/test_button_factory.py`, `tests/windows/test_use_this_instead_panel_button.py`, `tests/utils/test_generic.py::TestRestartApplication`
- [x] `python translation_helper.py check es_ES` → 100.0% translated, 0 unfinished, 0 missing, 0 obsolete
- [x] `python translation_helper.py validate es_ES` → valid, no mismatches
- [x] `ruff check` / `ruff format --check` / `mypy` clean on all changed files
- [x] Manually launched the app, switched language to Español, restarted, and confirmed the new strings render correctly and the restart no longer crashes


LionelColaso (approved):
Looks Good Thanks

PR #2294, opened 2026-07-07
Show per-mod startup load time from Loading Progress reports

### LINKED ISSUE #1160: Startup Impact Integration
### Origin of idea

It can be very useful to see how have your mods are. There is a mod for it: https://github.com/WimStienstra/StartupImpactContinued/tree/master
I patched it and made a draft integration that shows load time per mod.

The mod is patched to dump metrics to XML file, that can be parsed by RimSort and show in active mod list.

Image

### Alternatives considered

_No response_

### Additional context

Is it valuable at all? Should I use this approach (a bit wonky and complex)? Do we have alternatives?

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

carmoran0:
Closes #1160

## Description

Adds an opt-in mod list indicator showing how much each mod contributes to RimWorld's startup time, using the `StartupImpactData.xml` report that the [Loading Progress](https://github.com/ilyvion/loading-progress) mod saves to the RimWorld save data folder (the parent of the instance's `Config` folder).

How it works: enable **Settings → Sorting → "Show startup load time per mod"** (off by default), save a startup impact report in game (Loading Progress → "Track startup loading impact", then Save in its startup impact window), and refresh RimSort.

- Each mod row shows its impact as whole milliseconds (e.g. `36387ms`) at the far right, after the updated/warning/error icons: green below 1s, orange from 1s, red from 5s.
- The tooltip breaks down off-thread (loading screen) time, the top 3 impact categories, and when the report was measured plus the total game startup time.
- `app/utils/startup_impact.py` parses the Scribe-serialized report (timings are stored in milliseconds and exposed as seconds), tolerates Scribe's omitted-default elements, and caches by file mtime; the cache is invalidated on refresh, so re-saving a report in game and refreshing picks up new timings.
- Report entries are matched by normalized packageId when present, falling back to a case-insensitive mod name match. Current Loading Progress releases only write `modName`; an upstream PR (ilyvion/loading-progress#2) adds `modPackageId` and an auto-save option — this parser already reads `modPackageId` opportunistically, so no RimSort changes will be needed when that lands.

## Also included

- Fix `just i18n-update`: pyside6-lupdate's directory scan does not include `.py` in its default extension list, so it found 0 strings and marked every entry obsolete. The recipe now passes `-extensions py`. The `.ts` files are regenerated (new feature strings plus strings that had been pending since the recipe broke) and the `.qm` files recompiled.


carmoran0:
Here's what it looks like with my mods. The next step is to make a PR in "loading progress" to enable automatic saving of the reports. Right now you have to save the report manually in the mod settings for it to work.
Captura de pantalla_20260707_172538

LionelColaso (approved):
Looks Good

PR #2304, opened 2026-07-11
fix(tags): restore TagEditDialog improvements from PR #2171

cebarks:
## Summary

- Re-applies the `TagEditDialog` UX improvements from PR #2171 that were accidentally reverted by the metadata refactor in PR #2151 (commit `1928f297`)
- Restores MultiSelection mode (replaces checkboxes), keyboard-driven tag creation, arrow-key navigation, automatic list sorting, and corresponding tests
- Includes the operator precedence fix from commit `0e84d751`

Fixes https://github.com/RimSort/RimSort/pull/2171#issuecomment-4940264127


cebarks:
@Adda0 can you verify that this PR fixes the regression?

Adda0:
Thanks for the quick response and fix. It seems to restore the previous behaviour correctly. Except that all the doc comments, comments (including e.g., `# jscpd:ignore-start`; admittedly, the comments in the keyboard events are now superfluous, but the others were meaningful, I'd say), and some asserts in the code are missing. The test (which was actually misnamed [filter_tags](https://github.com/RimSort/RimSort/pull/2171/changes#diff-9593223a7c8ee010c66185cf4a9262ef66c8248feb86ea58a0d501035e1fd454R147) instead of `test_filter_tags` in the original PR) has been split into multiple small tests, which breaks one feature of the test because it also tests how filters are cleared and updated on subsequent changes. Otherwise, looks correct.

Adda0:
I found, however, that the up arrow event you added separately is missing.

Adda0:
Yeah, all works as expected for me now. Thanks for the fix :heart:

(Be aware that the PR now includes a large number of unrelated changes, which may or may not be intended.)

LionelColaso:
@cebarks the Pr needs to be cleaned i think merge may have issues

LionelColaso:
it would be better if you rebase your branch

LionelColaso (approved):
Looks Good Thanks

Adda0:
Hey. Thank you for the fix. All seems to work as expected now in v1.10.2.

PR #2320, opened 2026-07-17
fix: keep local git/SteamCMD mods active over their Workshop copy

### LINKED ISSUE #2300: RimSort prefers Workshop mod over locally installed mod
### Release Type

Compiled (Windows x86-64)

### Version

1.8.0

### Operating System

Win11

### Relevant logs

_No response_

### Describe the bug

RimSort.log: https://github.com/user-attachments/files/29812378/RimSort.log

I have noticed that when I copy my GitHub project folder containing my in-development mod into the game's mod folder and then toggle my published mod as inactive within Rimsort, the local mod (which then has a Git icon) that I toggle as active, sort, and try to save does not save correctly.

When I press "Run", I receive a message saying that my changes haven't been saved. However, if I click either "Save and Run" or "Run Anyway," my in-development mod loads correctly.

If I then exit the game and restart Rimsort, my in-development mod is placed in the inactive table, and my published mod is active again.

### Additional context

It seems that the duplicate of mod names, which occurs during the development and testing of one's own mods, conflicts with something.

Also, when you restart RimSort, it prefers the published mod to the locally installed mod.

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

keyz182:
## Summary

Fixes #2300. Activating a local copy of a mod that is also installed from the
Workshop (same package ID) did not stick. On restart the Workshop copy became
active again, and pressing Run reported unsaved changes immediately after a save.

## Fix

The duplicate-resolution priority lists in `MetadataController.get_mods_from_list`
only held LUDEON, LOCAL and STEAM_WORKSHOP. A local copy typed `ModType.GIT` or
`ModType.STEAM_CMD` matched no entry, so resolution fell through to the Workshop
copy. `app/models/mod_list.py` already carried the complete lists, so the two
had drifted.

Consolidated both orderings into `metadata_structure.py` next to `ModType` as a
single definition (GIT and STEAM_CMD ahead of STEAM_WORKSHOP for a bare ID, after
it for a `_steam` ID) and imported them from `metadata_controller.py` and
`mod_list.py`, deleting the local copies so they cannot drift again.

The same resolver is behind the false unsaved-changes prompt: `_do_save` builds
its "last saved" baseline by re-running resolution over what it just wrote, so
the baseline held the Workshop path while the active list held the git path. One
fix covers both symptoms.

## Testing

- Controller tests: git beats Workshop and SteamCMD beats Workshop for a bare ID,
  and a `_steam` ID still picks Workshop when a git copy is present.
- `mod_list` test: git beats Workshop via `ModList.from_mods_config`.
- metadata_controller, mod_list, metadata_parity and metadata_structure suites
  pass; ruff check and format clean.

### DIFF
--- a/app/controllers/metadata_controller.py
+++ b/app/controllers/metadata_controller.py
@@ -14,6 +14,8 @@ from PySide6.QtCore import QMutex, QObject, Signal, Slot
 from app.controllers.metadata_db_controller import AuxMetadataController
 from app.models.metadata.metadata_mediator import MetadataMediator
 from app.models.metadata.metadata_structure import (
+    SOURCE_PRIORITY_DEFAULT,
+    SOURCE_PRIORITY_STEAM,
     AboutXmlMod,
     CompiledDependencyData,
     ListedMod,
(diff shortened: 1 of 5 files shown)


LionelColaso (approved):
Looks Good Thanks

PR #2342, opened 2026-07-29
Fix thread priority setting logic in SearchWorker

Spagles:
Adjust the logic for setting thread priority to ensure it only applies when the thread is running, preventing unnecessary warnings.


LionelColaso (approved):
Looks Good Thanks

PR #2346, opened 2026-08-01
Fix silent update failures on Linux requiring elevation

Spagles:
fixes a critical issue where the auto-updater would silently fail on Linux if the application was installed in a protected directory. The primary launch method used `subprocess.Popen` with `shell=True`, which successfully spawned a shell but failed internally when `sudo` lacked a terminal prompt. By explicitly forcing an exception in this scenario, the intended fallback terminal emulator is successfully launched so the user can authenticate.


LionelColaso (approved):
Looks Good Thanks

PR #2362, opened 2026-08-14
fix signal emit in settings failure dialog, auto-update prompt check, and appearance branch logic

Spagles:
- added missing parentheses to `EventBus().reset_settings_file.emit()` in `SettingsFailureDialog`. Previously, clicking "Reset Settings" only referenced the bound method without invoking it, which meant the signal was never sent to `SettingsController` to overwrite the corrupted config and reload defaults.
- updated `if refresh_now:` to `if refresh_now == QMessageBox.StandardButton.Yes:` in `_on_auto_updates_complete()`. Because `show_dialogue_conditional` returns `QMessageBox.StandardButton` enums (`Yes` = 16384, `No` = 65536), checking it as a bare boolean evaluated to `True` even when clicking "No", causing an unintended mod list refresh.
- changed `if browser_state == "normal":` to `elif` in `update_view_from_model()`. The detached `if` statement caused a `"maximized"` setting to fall through the rest of the checks and execute the fallback `else:` block unnecessarily.

### DIFF
--- a/app/controllers/main_content_controller.py
+++ b/app/controllers/main_content_controller.py
@@ -296,7 +296,7 @@ class MainContentController(QObject):
                 "Refresh now?"
             ).format(count=len(succeeded), summary="<br>".join(summary_parts)),
         )
-        if refresh_now:
+        if refresh_now == QMessageBox.StandardButton.Yes:
             EventBus().do_refresh_mods_lists.emit()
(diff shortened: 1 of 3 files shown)


LionelColaso (approved):
Looks Good Thanks

PR #2397, opened 2026-08-30
Stop unsubscribe/subscribe from always blocking for 30 seconds

keyz182:
Subscribing, unsubscribing, and resubscribing to Workshop mods always took the full 30-second timeout to complete, even when Steam confirmed the operation almost instantly. `_wait_for_callbacks` already has an early-exit check for this:

```python
    if self.callbacks_total and self.callbacks_count >= self.callbacks_total:
        return True
```

But `callbacks_total` is hardcoded to `None` in `run()`, so that check has been dead code,  it can never fire.

That was introduced in #2119, which replaced a per-item callback count with a flat timeout. "Steam may not fire callbacks reliably" (per that PR's own comment). That tradeoff was made to address #1460, where the dev later confirmed the real issue:

> steam sometimes does not initiate downloads when cmd is sent... steam sometimes does not respond to the command sent by RimSort.

That's about Steam silently dropping a command, not about callback counts being unreliable when Steam does respond. That said, @LionelColaso - maybe you can comment on if I've missed a separate reason that justifies keeping the timeout only?

This restores `callbacks_total` (mod count, doubled for resubscribe since that queues an unsubscribe + subscribe callback per mod) so the existing early-exit actually runs. When Steam responds normally, we return as soon as all expected callbacks arrive instead of always waiting 30s. When Steam doesn't respond, nothing changes, same 30s fallback as before. This isn't a fix for #1460's dropped-command issue; it's a separate perf regression introduced as a side effect of that fix.


LionelColaso:
@keyz182 Please remove the SteamworksPy.so binary its not required to generate it except for that The fix restores intended behavior with proper fallback. and is good for merge
i still need to update a couple of things and revist this properly when i get time

keyz182:
> @keyz182 Please remove the SteamworksPy.so binary its not required to generate it except for that The fix restores intended behavior with proper fallback. and is good for merge i still need to update a couple of things and revist this properly when i get time

Will do - didn't mean to check that in.

keyz182:
> @keyz182 Please remove the SteamworksPy.so binary its not required to generate it except for that The fix restores intended behavior with proper fallback. and is good for merge i still need to update a couple of things and revist this properly when i get time

Done :+1:

LionelColaso (approved):
Looks Good Thanks

PR #2404, opened 2026-09-03
feat: mod list history with snapshot diff viewer (closes #1678)

### LINKED ISSUE #1678: Allow current active modlist to be compared to a saved modlist
### Origin of idea

Saw a recommendation for debugging where someone said to create a steam collection of all your active mods and unsubscribe from the rest of them in order to see which mods are going in better. Looked for an option to do that in Rimsort and couldn't find one, so I just didn't do that. 
After thinking about it a little more, I thought that allowing the user to see a diff of their current active modlist and a modlist saved in a file is probably better, as that way you don't have to unsubscribe from everything else.

Allow current active modlist to be compared to a saved modlist (perhaps like a filter view, or just a red-green diff overlay)

### Alternatives considered

could do the other way and let it be saved as a steam collection

### Additional context

_No response_

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

DaraynLB:
### Summary

Implements #1678 — keep a history of the active mod list so a user can diff
two versions and narrow down which mod change coincided with the onset of
errors.

### Changes

- **New `ModlistHistoryService`** (`app/services/modlist_history_service.py`).
  On every save it writes a timestamped JSON snapshot of the active + inactive
  mod lists to `modlists/history/<instance>/`. Each snapshot is identified by a
  SHA-256 of the ordered active package IDs and chained to its predecessor via
  `previousId`, forming a breadcrumb trail. A save whose active list is
  identical to the last snapshot is skipped; `modlist_history_retention_count`
  prunes the oldest. Snapshot files are valid RimSort JSON mod lists (plus a
  `rimsortSnapshot` metadata block), so they re-import through the existing
  parser. A human-readable `history.log` line is appended per save.
- **New "File → Mod List History…" dialog** (`app/windows/modlist_history_panel.py`).
  Lists the snapshots and shows the added / removed / reordered mods between
  any two of them (select one to diff against its predecessor, or Ctrl-select
  two). Can restore an old list into the UI (unsaved, same behaviour as file
  import), export a snapshot, attach a note, or open the history folder.
- `MainContent._do_save` hooks the snapshot write (guarded — a failure here
  never blocks saving). The post-parse body of `_do_import_list_file_xml` is
  factored into `_apply_imported_package_ids` and reused for restore.
- Settings → Advanced gains a "Mod list history" group: `modlist_history_enabled`
  (default on) and `modlist_history_retention_count` (default 100, -1 = keep all).
- `EventBus.do_open_modlist_history` + the File-menu action wiring.

### Tests

- `tests/services/test_modlist_history_service.py` — id determinism and
  order-sensitivity, snapshot write + dedupe, `previousId` chaining, prune
  (keep-N and keep-all), and add/remove/reorder diffing.
- `tests/models/test_settings_defaults.py` — the two new defaults.
- `tests/views/test_menu_bar.py` — the new menu action emits its signal.

### Verification

- `just test` — 1346 passed, 13 skipped
- `just typecheck` clean; `ruff check` / `ruff format` clean;
  `check_deferred_imports.py` clean
- Manual: repeated save cycles produce snapshots; an unchanged save writes
  nothing; the diff view reflects activate/deactivate/reorder; restore, export,
  notes, retention pruning, and the on/off toggle all behave as expected;
  opening the dialog with no snapshots shows an empty state without crashing.

### Notes

- Only `locales/en_US.ts` is updated (new source strings, extracted with
  `-locations none` to keep the diff minimal); other locales are left for the
  usual translation-sync PRs.
- Scope nuance: #1678 phrases it as comparing the *current* active list against
  a *saved modlist file*. This PR diffs between automatically-captured snapshots
  instead. Happy to add a "compare current list to an arbitrary file" entry
  point as a follow-up if that's preferred.

### DIFF
--- a/app/controllers/menu_bar_controller.py
+++ b/app/controllers/menu_bar_controller.py
@@ -72,6 +72,9 @@ class MenuBarController(QObject):
         self.menu_bar.export_to_rentry_action.triggered.connect(
             EventBus().do_export_mod_list_to_rentry
         )
+        self.menu_bar.modlist_history_action.triggered.connect(
+            EventBus().do_open_modlist_history.emit
+        )
 
         for action in self.menu_bar.upload_log_actions:
             action.triggered.connect(
--- a/app/controllers/settings_tabs/advanced_tab_controller.py
+++ b/app/controllers/settings_tabs/advanced_tab_controller.py
@@ -27,6 +27,10 @@ class AdvancedTabController(BaseTabController):
             self._on_include_mod_notes_in_mod_name_filter_changed
         )
 
+        self.dialog.modlist_history_enabled_checkbox.toggled.connect(
+            self._on_toggle_modlist_history_enabled
+        )
+
     def update_view_from_model(self) -> None:
         try:
             self.dialog.show_save_comparison_indicators_checkbox.setChecked(
@@ -64,6 +68,13 @@ class AdvancedTabController(BaseTabController):
             self.settings.enable_backup_before_update
         )
         self.dialog.max_backups_spinbox.setValue(self.settings.max_backups)
+
+        self.dialog.modlist_history_enabled_checkbox.setChecked(
+            self.settings.modlist_history_enabled
+        )
+        self.dialog.modlist_history_retention_count_spinbox.setValue(
+            self.settings.modlist_history_retention_count
+        )
 
         # Auxiliary DB
         self.dialog.aux_db_time_limit.setText(str(self.settings.aux_db_time_limit))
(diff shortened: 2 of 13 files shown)


LionelColaso (approved):
Approved. The automatic snapshot approach is great — better UX than the original ask. Since comparing against an arbitrary saved modlist file isn't covered yet. I'd suggest tracking that as a follow-up issue so keeping  #1678 stays open until the file-compare feature lands.
Thanks

PR #2412, opened 2026-09-06
fix: preserve arbitrary git repository paths

### LINKED ISSUE #2396: Add Git Mod fails if git URL does not contain exactly two path segments
### Release Type

Compiled (Windows x86-64)

### Version

v1.12.0

### Operating System

Windows 10

### Relevant logs

https://pastebin.com/AcbbT08q

### Describe the bug

RimSort fails to clone a git repository if the URL contains fewer than or more than exactly 2 path elements.

Examples:
`http://repo.host/repository` (1 path element) → silently fails (the log calls this an "Invalid git URL")
`http://repo.host/path/to/repository` (3 path elements) → strips the URL down to 2 path elements (`http://repo.host/path/to`) and attempts to clone that. Obviously, this fails.

### What I expect

As these are all perfectly valid remote git URLs, the clone process should succeed and the repository should be cloned into RimWorld's mods folder.

### Additional context

I use my own private git server for, among other things, hosting custom personal-use mods. As this is a personal hosted git server, it does not follow the GitHub-esque `/organization/repository` format that RimSort seems to be expecting; rather mods can be found at any arbitrary depth depending on how I set them up. (In this specific case I put them under `/mods/rimworld/modname`.)

----

(Potentially related issue: RimSort does not recognize valid git URLs that start with `git://` instead of `http://`.)

### Code of Conduct

### Duplicate Issue Check

### Wiki/FAQ Check

be-student:
Fixes #2396.

Preserves the complete repository path for regular HTTP(S) and `git://` remotes while retaining GitHub/GitLab browse-URL normalization. Private-host path segments are no longer interpreted as branch markers.

Validation: 1,362 tests passed with 1 skip; targeted Ruff, format, and mypy checks passed.

### DIFF
--- a/app/utils/git_utils.py
+++ b/app/utils/git_utils.py
@@ -575,7 +575,12 @@ class ParsedGitUrl:
     repo_name: str
 
 
-_BROWSE_PATH_SEGMENTS = {"tree", "blob", "commit", "releases", "issues", "pull"}
+_TWO_SEGMENT_BROWSE_HOSTS = {
+    "github.com",
+    "www.github.com",
+    "gitlab.com",
+    "www.gitlab.com",
+}
 
 
 def parse_git_url(repo_url: str) -> ParsedGitUrl | None:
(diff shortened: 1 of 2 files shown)


LionelColaso (approved):
Looks Good Thanks

Issue #2419, opened 2026-09-08
Linux: bundled SteamCMD overwrites desktop Steam libraryfolders.vdf


Exorion1er:
### Release Type

AppImage (x86-64)

### Version

1.12.0

### Operating System

Fedora 44

### Describe the bug

On Linux, RimSort launches its bundled SteamCMD with the same user HOME/config area as the desktop Steam client. SteamCMD then writes its own one-library `libraryfolders.vdf` over the desktop client's file. The next normal Steam launch no longer knows about secondary Steam libraries, although their game data and manifests are intact.

1. Configure native Steam with its default library plus a secondary library on another mounted drive.
2. Quit Steam and launch RimSort; allow its SteamCMD-backed functionality/startup to run.
3. Start normal Steam again.

### Additional context

Steam's content log showed:

- Normal Steam loaded the secondary library and triggered a write for **2 libraries**.
- RimSort's SteamCMD then loaded 0 apps from its own `.../RimSort/instances/Default/steam/steamapps` path and triggered a write for **1 library**.
- The next normal Steam launch loaded only its default library. Re-adding the secondary library made all already-installed games reappear.

This repeats after each RimSort launch. The affected secondary drive was mounted from `/etc/fstab` by UUID before Steam started, so it is not a mount-timing failure.

Valve documents the underlying Linux behavior: SteamCMD and the Steam client share per-user configuration unless SteamCMD runs under a separate user/config home:

https://steamcommunity.com/groups/SteamClientBeta/discussions/0/4840896974232786377/

### Expected behavior

RimSort's SteamCMD should use an isolated config home under its own configured SteamCMD prefix, without changing the user's desktop Steam configuration.

### Suggested fix

When spawning `steamcmd.sh` on Linux, provide a dedicated environment for the subprocess, for example:
```text
HOME=<steamcmd_prefix>/home
XDG_DATA_HOME=<steamcmd_prefix>/home/.local/share
XDG_CONFIG_HOME=<steamcmd_prefix>/home/.config
```

This should be passed by RimSort's process runner rather than requiring users to patch the downloaded steamcmd.sh. The SteamCMD download/install location can remain unchanged, so existing workshop content and symlinks are preserved.

### Local workaround and verification

[...]

PR #2421, opened 2026-09-09
fix: isolate Linux SteamCMD user configuration

be-student:
## Change

Run Linux workshop downloads with `HOME`, `XDG_CONFIG_HOME`, and `XDG_DATA_HOME` under a dedicated `home` directory inside the configured SteamCMD prefix. SteamCMD no longer inherits the desktop Steam client's user-configuration paths. Refs #2419.

The first download, queued batches, and manual restart retain these overrides. Other inherited variables, including `PATH`, remain available. Windows and macOS retain their existing environment. Relative prefixes resolve to absolute isolation paths before launch. User documentation describes the dedicated directory.

## Validation

Regression tests cover process wiring, absolute paths, relative prefixes, inherited variables, queued batches, restart, and non-Linux behavior.

- Final tests against base source: 6 failed, 14 passed.
- Same targeted files with the fix: 20 passed.
- `uv run pytest --doctest-modules --no-qt-log -q`: 1369 passed, 1 skipped.
- Full mypy: no issues in 282 source files; pyright: 0 errors/warnings.
- Changed-file Ruff and formatting, markdownlint, duplicate-code scan, deferred-import guard and `git diff --check`: passed.
- Whole-tree Ruff still reports an unrelated `EXE001` on untouched `app/__main__.py` (its tracked file mode is not executable).

These checks ran on macOS. They verify environment construction and QProcess wiring; no real Linux SteamCMD run or desktop-library preservation trial was performed. Directory-creation errors fail before the subprocess starts rather than falling back to the desktop environment.

## Agent disclosure

Generated and independently reviewed by OpenAI Codex for be-student. No human review has occurred before submission; maintainer review and Linux runtime verification are requested.

### DIFF
--- a/app/utils/steam/steamcmd/wrapper.py
+++ b/app/utils/steam/steamcmd/wrapper.py
@@ -293,6 +293,27 @@ class SteamcmdInterface:
     def console_log_path(self) -> Path:
         return Path(self.steamcmd_install_path) / "logs" / "console_log.txt"
 
+    def _get_process_environment(self) -> dict[str, str] | None:
+        """Return an isolated environment for SteamCMD on Linux.
+
+        SteamCMD otherwise shares the desktop Steam client's configuration under
+        the user's home directory and can overwrite its ``libraryfolders.vdf``.
+        Other platforms retain their existing inherited environment.
+        """
+        if self.system != "Linux":
+            return None
+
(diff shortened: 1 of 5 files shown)


LionelColaso (approved):
Nice catch looks Good Thanks

PR #2448, opened 2026-09-16
feat(autodetect): discover non-Steam RimWorld installations (GOG, Heroic)

### LINKED ISSUE #2447: Autodetect does not find RimWorld installed via GOG (macOS and Linux)
**Problem**

When RimWorld is installed through GOG (GOG Galaxy or the offline installer), the **Autodetect** button in Settings → Locations fails to fill the *Game location* and *Local mods location* fields.

`PathAutodetectService` currently only looks at Steam installations: it parses `libraryfolders.vdf` and falls back to hardcoded Steam paths. With no Steam installation present, it returns non-existent Steam fallback paths, the fields stay empty, and the essential-paths check keeps failing on every launch.

The same applies on Linux for GOG installs (official offline installer under `~/GOG Games/<title>/game`, or Heroic/Lutris under `~/Games/...`).

**Environment**

- RimSort built from current `main` (up to `64d9ce7`)
- macOS 15 (Apple Silicon), RimWorld 1.6.4871 via GOG Galaxy, installed at `/Applications/RimWorld.app`
- Steam not installed

**Notes on the GOG layout (macOS)**

The GOG build ships as a single bundle with game content at the bundle root:

```
/Applications/RimWorld.app
├── Contents/
├── Data/          (Core + DLCs)
├── Mods/
└── Version.txt
```

So the bundle itself is a valid "game folder" for RimSort's macOS handling (`get_executable_path()` already expects the `.app` directory), it is just never looked for outside Steam.

**Expectation**

Autodetect should also discover non-Steam installations so first-run configuration works without manually locating the game folder (or maintaining symlinks as a workaround).

I have a working implementation and can open a PR. Sketch of the approach: keep Steam detection unchanged, then fall back to launcher metadata (Heroic's GOG `installed.json`) and a bounded, content-validated search over platform-standard roots (`/Applications`, `~/Applications`, `~/GOG Games`, `~/Games`), keeping the existing hardcoded paths as the last resort.

0-danielviktorovich-0:
Resolves #2447

**Problem**

`PathAutodetectService` only detects Steam installations. When RimWorld comes from GOG (Galaxy on macOS defaults to `/Applications/RimWorld.app`; the offline installer / Heroic / Lutris on Linux use other roots), autodetection returns non-existent Steam fallback paths, leaves the location fields empty, and the essential-paths check keeps failing on launch.

**What this PR does**

1. **Non-Steam discovery pipeline** in `PathAutodetectService`, shared by macOS, Linux and Windows, tried only when Steam detection does not produce an existing game folder:

   Steam VDF parsing (unchanged) → launcher metadata → bounded search over standard roots → legacy hardcoded fallback.

   - *Launcher/registry metadata:* Heroic Games Launcher's GOG `installed.json` (`install_path` entries, path resolved per-platform) covers custom Heroic install locations; on Windows, the GOG Galaxy registry key (`HKLM\SOFTWARE\WOW6432Node\GOG.com\Games`, per-game `executable` value) covers custom install drives without any filesystem scanning.
   - *Standard roots:* `/Applications` and `~/Applications` on macOS; `~/GOG Games` (GOG offline installer, minigalaxy) and `~/Games` (Heroic, Lutris) on Linux; `C:/GOG Games`, `~/GOG Games` and `~/Games` on Windows. Depth ≤ 2, so `~/GOG Games/<title>/game` and `~/Games/<provider>/<title>` are covered without hardcoding game folder names.

2. **Content validation instead of name matching.** A candidate must carry RimWorld markers — a parsable `Version.txt` (e.g. `1.6.4871 rev573`) or a known game executable; macOS bundles additionally need `Contents/` plus the Ludeon bundle-root layout (`Data/`, `Mods/`). A look-alike folder is rejected regardless of its name.

3. **`RimWorldLinux64`** is now accepted alongside the legacy `RimWorldLinux` binary in `get_executable_path()` and in the settings validation message (current Steam/GOG Linux builds ship the 64-bit binary).

4. **Silent first-run fill** in `MainContent.check_if_essential_paths_are_set()`: before showing the "Essential path(s)" dialog, the missing essential paths (game, config, local mods; workshop only if it exists) are auto-filled from the same service and saved. Existing values are never overwritten, Steam-integration checkboxes are not touched, and refreshes with `prompt=False` (e.g. after "Clear All Locations") do not re-trigger autodetection.

**Security considerations** (the search runs on the user's filesystem, so it is deliberately conservative):

- strictly read-only: nothing is executed, nothing is written — results are only pre-filled in the settings dialog / instance fields for the user to keep;
- symlinked directories are skipped (`os.scandir` with `follow_symlinks=False`, plus explicit `is_symlink()` checks), so symlink loops cannot occur;
- searches are bounded: depth ≤ 2, ≤ 500 entries per directory, hidden entries skipped;
- launcher metadata is parsed with the stdlib JSON parser from a fixed user-owned config location, size-capped (5 MB), shape-checked, and malformed files are skipped rather than trusted;
- no new dependencies.

**Testing**

- New unit tests: `tests/controllers/test_path_autodetection.py` (GOG/Heroic discovery on all three platforms — the Windows registry lookup is covered by a read-only fake `winreg`, so it runs on every platform — content validation, bounded/symlink-safe search, `RimWorldLinux64`) and `tests/views/test_essential_paths_autodetect.py` (silent fill, no-overwrite, no-refill after clearing, dialog only when autodetection finds nothing). Shared view fixtures moved to `tests/views/conftest.py`. One Linux-specific assertion (non-executable binary rejection) is skipped on Windows where `os.access(X_OK)` does not reflect the executable bit.
- Full suite: `pytest` — 1414 passed, 1 skipped; ruff, mypy and pyright clean on the touched files.
- Manually verified on a real GOG (macOS, Apple Silicon) install with Steam absent: on a fresh profile the app auto-fills `/Applications/RimWorld.app`, the RimWorld config folder and the bundle-root `Mods` folder, and the metadata scan lists all six Ludeon mods (Core + Anomaly, Biotech, Ideology, Odyssey, Royalty). The Autodetect button behaves identically when invoked manually.

Happy to adjust the root lists or the validation markers if maintainers prefer different trade-offs.


app/services/path_autodetect_service.py
```diff
@@ -215,8 +274,372 @@ def get_windows_paths(self) -> tuple[Path, Path, Path]:
...
+                subkey_count = 0
+                while subkey_count < self._MAX_METADATA_ENTRIES:
+                    try:
+                        subkey_name = winreg.EnumKey(games_key, subkey_count)
+                    except OSError:
+                        break
+                    subkey_count += 1
+                    try:
+                        with winreg.OpenKey(games_key, subkey_name) as game_key:
+                            exe_value = winreg.QueryValueEx(game_key, "executable")
```
LionelColaso: The value name here doesn't match what GOG Galaxy actually writes. GOG Galaxy stores each game under `HKLM\SOFTWARE\WOW6432Node\GOG.com\Games\<id>` with values `path` (install dir) and `exe`/`EXE` (executable path); nothing writes `executable`. Independent consumers read `path` (ModOrganizer2 basic_games `gog_utils.py`, Fluorine-Manager, keypirinha-allmygames) or `exe` (keypirinha). As written, `QueryValueEx(game_key, "executable")` raises `OSError` and hits `continue`, so this branch never finds a game on Windows.

Suggested fix: read `path` (already the game dir) with an `exe`/`EXE` fallback (`Path(value).parent`):
```python
try:
    install_path = winreg.QueryValueEx(game_key, "path")
    candidate = Path(install_path[0])
except OSError:
    exe_value = winreg.QueryValueEx(game_key, "exe")
    candidate = Path(exe_value[0]).parent
```
Please also update the `_FakeWinreg` fixture/test that asserts the `executable` value name, and add a case for `path`/`exe`.

app/services/path_autodetect_service.py
```diff
@@ -215,8 +274,372 @@ def get_windows_paths(self) -> tuple[Path, Path, Path]:
...
+            if metadata_file.stat().st_size > self._MAX_METADATA_FILE_BYTES:
+                logger.warning("Heroic GOG metadata file is too large; skipping")
+                return None
+            installed = json.loads(metadata_file.read_text(encoding="utf-8"))
+        except (OSError, ValueError):
+            logger.warning(
+                "Failed to read Heroic GOG metadata; skipping", exc_info=True
+            )
+            return None
+        if not isinstance(installed, list):
```
LionelColaso, line 401: Heroic's `gog_store/installed.json` is an electron-store DB — a JSON object with a single `installed` key holding the array (`new TypeCheckedStoreBackend('gogInstalledGamesStore', { cwd: 'gog_store', name: 'installed' })`, then `.set('installed', [...])`), not a top-level array. See Heroic `src/backend/storeManagers/gog/electronStores.ts`. As written, `isinstance(installed, list)` is always `False` on real installs, so the Heroic branch never fires on Linux/Windows. Entries do use `install_path` (and `executable`), so only the top-level shape needs handling.

Suggested fix (keeping a legacy top-level list working):
```python
if isinstance(installed, dict):
    installed = installed.get("installed", [])
if not isinstance(installed, list):
    return None
```
Please also update `test_finds_heroic_metadata_custom_location`, which currently writes a top-level list and therefore passes despite the mismatch.

app/views/main_content_panel.py
```diff
@@ -399,39 +457,43 @@ def check_if_essential_paths_are_set(self, prompt: bool = True) -> bool:
...
-                self._show_settings_dialog("Locations")
-            return False
+
+        logger.warning("Essential path(s) are invalid or not set!")
+
+        # First-run convenience: silently run path autodetection and save any
+        # missing essential paths before bothering the user. Deliberate
+        # "Clear All Locations" flows reach the non-prompting refresh path
+        # (prompt=False) and are not affected by this.
+        if prompt and self._autodetect_missing_essential_paths():
```
LionelColaso: `_autodetect_missing_essential_paths()` fills only paths that already exist, but this returns `True` for any fill — it doesn't re-check that game + config + local are now all set and exist (what `_instance_essential_paths_ready` requires). In the target first-run scenario (fresh GOG install, game never launched) there is no config folder yet: `game_folder` and `local_folder` (`game/Mods`) get filled, `config_folder` stays empty, this returns `True`, and the "Essential path(s)" dialog is skipped — the app proceeds without a config.

Suggested fix: re-evaluate readiness after autofill, e.g.
```python
if prompt and self._autodetect_missing_essential_paths():
    if self._instance_essential_paths_ready(
        self.settings.instances[self.settings.current_instance]
    ):
        logger.info("Essential paths were completed by silent autodetection")
        return True
```
Please add a test where the config folder does not exist yet and assert the dialog path is taken.

LionelColaso (changes requested):
Thanks for this — the autodetection pipeline (Steam VDF → launcher metadata → bounded root scan → legacy fallback), the read-only/symlink-safe scanning, and content-based validation are a solid approach, and the test breadth is good.

However, the two launcher-metadata parsers don't match the formats actually written by GOG Galaxy and Heroic, so those branches won't fire on real installs, and the silent-autofill exit has a gap. CI is green here, but only because the new tests encode the same assumptions (fake registry `"executable"` value; Heroic `installed.json` as a top-level list) — the codecov report also shows ~37 uncovered lines concentrated in `path_autodetect_service.py`.

Requesting changes for three items (inline):
1. `app/services/path_autodetect_service.py:497` — GOG Galaxy registry value is `path`/`exe`, not `executable`.
2. `app/services/path_autodetect_service.py:398` — Heroic `gog_store/installed.json` is `{"installed": [...]}`, not a top-level list.
3. `app/views/main_content_panel.py:472` — `check_if_essential_paths_are_set` returns `True` on partial autofill without re-verifying the config folder.

Suggest fixing the parsers and backing them with fixture-based tests using the real schemas, which would also close the codecov gap.

app/services/path_autodetect_service.py
0-danielviktorovich-0: Fixed in `56b5a2f`. The registry lookup now checks `path` as the install directory, then `exe`/`EXE` as executable-path fallbacks. Each candidate is content-validated, so a stale `path` value can fall through to a valid executable entry. Tests cover all three value names, stale-path fallback, invalid values, and unrelated games.

app/services/path_autodetect_service.py
0-danielviktorovich-0: Fixed in `56b5a2f`. Heroic metadata now reads the electron-store object via its `installed` key while retaining compatibility with the legacy top-level list. The fixture now uses the real `{"installed": [...]}` shape, and negative fixtures cover an invalid `installed` value plus malformed entries.

app/views/main_content_panel.py
0-danielviktorovich-0: Fixed in `56b5a2f`. After silent autodetection, the code now re-runs `_instance_essential_paths_ready` before returning success. A regression test covers the fresh-install case where game and Mods exist but config does not: partial values are saved, the method returns `False`, and the configuration dialog is shown.

0-danielviktorovich-0:
Coverage hardening is complete in `895de4c`, `1e40710`, and `37e22bf`. The added tests cover cross-platform fallback behavior, registry/Heroic edge cases, defensive filesystem failures, and GOG provenance without relying on a locally installed game. Codecov now confirms: **All modified and coverable lines are covered by tests** (100% patch coverage). The final local full suite passes with **1447 passed, 1 skipped**; Ruff, mypy, Pyright, and duplicate detection are clean. These follow-ups change tests only, not production behavior.

app/services/path_autodetect_service.py
LionelColaso: Verified — thanks. `_find_gog_registry_game_folder` now reads `path` (install dir), then `exe`/`EXE` (`.parent`), with per-value content validation via `_looks_like_rimworld_windows_dir`, so a stale/missing value falls through instead of aborting. Tests cover all three value names, the stale-`path` fallback, invalid/vacuous values, a missing root key, and an unreadable subkey — all passing locally at `1e40710`. Resolving.

app/services/path_autodetect_service.py
LionelColaso: Verified — thanks. `_find_heroic_game_folder` now reads the electron-store object via `metadata.get("installed", [])` while retaining legacy top-level-list compatibility; fixtures use the real `{"installed": [...]}` shape, and negative cases cover an invalid `installed` value, malformed entries, and oversized files — per-codecov 100% patch coverage. Resolving.

app/views/main_content_panel.py
LionelColaso: Verified — thanks. `check_if_essential_paths_are_set` no longer keys off the autofill return value; after silent autodetect it re-runs `_instance_essential_paths_ready` and falls through to the Essential path(s) dialog when the config folder is still missing. `test_prompts_when_autodetect_only_partially_fills_paths` covers exactly that fresh-GOG-install case, asserting partial paths are saved, the method returns `False`, and the dialog is shown. Resolving.

LionelColaso (approved):
All three blockers are resolved and verified against the head at `37e22bf`:

1. **GOG Galaxy registry** now reads `path`/`exe`/`EXE` with per-value content validation and stale-value fallback, matching the real Galaxy schema and independent consumer implementations (`path`/`exe`, not `executable`).
2. **Heroic metadata** parses the electron-store `{"installed": [...]}` shape while tolerating a legacy top-level list; fixtures now use the real schema with negative cases.
3. **Silent autofill** re-checks `_instance_essential_paths_ready` after filling, so a missing config folder still shows the Essential path(s) dialog, with a regression test for the fresh-install case.

Verified locally (95 passed; the 3 symlink failures were a local Windows privilege limitation, all green on CI), and the follow-up commits since the fixes are test-only (100% patch coverage per codecov). Lint, pytest (macOS/ubuntu/windows), pyright, builds, CodeQL and codecov all pass on the final head. Nice work — approving.

PR #1914, opened 2026-04-16
[fix] Memory leak and hanging process with built-in browser

codebykoi:
Fixes #1881, #1871

### Root Cause

SteamBrowser class had no `closeEvent()` override. QWebEngineView / QWebEnginePage require explicit cleanup otherwise they leave zombie processes and circular references that prevent garbage collection. Main window also kept a strong reference that was never nulled.

### Fix Applied:
1. Added proper `closeEvent()` handler to SteamBrowser:
   - Sets `WA_DeleteOnClose` attribute
   - Stops any active page loading
   - Disconnects all web view signals
   - Properly destroys QWebEnginePage, QWebEngineProfile, QWebChannel
   - Recursively cleans up all layouts and widgets
   - Clears all internal references

2. Fixed main_content_panel browser handling:
   - Added cleanup of existing browser instance before creating new one
   - Added `destroyed` signal handler that automatically nulls the reference
   - Added safety checks to prevent multiple instances

Feel free to edit or cherry pick the fix for yourself, I just got tired of restarting app just to get rid of that bug.

PR #2451, opened 2026-09-17
fix(steambrowser): preserve mod downloader wait-list across SteamCMD failures

pjgoliveira21:
## Problem

The Mod Downloader's wait-list (the mods you queue up while browsing the
Workshop before hitting "Download mod(s) (SteamCMD)") gets wiped out
whenever something goes wrong during the SteamCMD download — forcing users
to rebuild the entire list from scratch by re-browsing every mod.

## Root cause

Clicking "Download mod(s) (SteamCMD)" calls `self.steam_browser.close()`
*before* SteamCMD even starts (`main_content_panel.py`). `SteamBrowser`'s
`closeEvent` unconditionally clears `downloader_list_mods_tracking` /
`downloader_list_dupe_tracking` and marks the widget `WA_DeleteOnClose`,
destroying it outright.

This happens regardless of whether the download later succeeds or fails —
so the instant you click download, the list backing it is already gone.
If SteamCMD then errors on any mod, there's no queue left to fall back to
or retry from.

(This `closeEvent` teardown itself is intentional — it was added in #1914
to fix a memory leak/hanging-process bug — so the fix here avoids touching
that cleanup and instead preserves the list data outside the widget's
lifecycle.)

## Fix

- Before closing the browser to start a download, snapshot its wait-list
  (`SteamBrowser.get_download_list_snapshot()`) into a spot on
  `MainContentPanel` that survives the widget being destroyed.
- Add a `steamcmd_mod_download_succeeded` event, emitted per-mod in
  `RunnerPanel` the moment SteamCMD confirms a successful download. Each
  success removes that one mod from the preserved snapshot.
- If the Mod Downloader is reopened later (a fresh `SteamBrowser` instance
  is created, since the old one was destroyed), the snapshot is restored
  into its list via `SteamBrowser.restore_download_list()`.

Net effect: mods that download successfully are dropped from the list as
they complete; mods that fail (or never get attempted, e.g. SteamCMD
crashing outright) stay queued and reappear next time the Mod Downloader
is opened, instead of being silently lost. The existing "Clear List"
button still works as the manual way to empty it out.

Applied the same snapshot/restore around the Steamworks-subscribe download
path for consistency, since it closes the same browser window the same way.

## Testing

- `ruff check` / `ruff format --check` pass on all touched files.
- Manually verified: queued several mods, triggered a SteamCMD download,
  interrupted/failed a subset, reopened the Mod Downloader, and confirmed
  only the mods that hadn't successfully downloaded were still listed.


LionelColaso:
Solid fix — root cause (the download button emits `downloader_list_mods_tracking` itself, so `closeEvent` wipes it before SteamCMD starts) is real, and the snapshot/restore + per-mod success drop is a clean design. A few change requests before merge:

**1. Steamworks path can still silently drop queued mods on failure (medium).**
`_do_steamworks_api_call_animated` pops every pfid out of `_pending_downloader_snapshot` unconditionally after the API call returns. But `_do_steamworks_api_call` can fail *quietly* — e.g. `check_steam_available()` returns `False` and it early-returns without a raised exception (main_content_panel.py:2428–2430) — and the mods are dropped from the preserved list though they were never handled. The SteamCMD path only drops mods on a confirmed success signal; this path can't distinguish, defeating the fix's premise for the "Steam app" button. Suggest only dropping on actual success (e.g. skip the pop when `check_steam_available` fails / when the operation reports failure).

**2. int vs str pfid mismatch can break the snapshot drop (low–medium).**
Snapshot keys come from `item.data(Qt.ItemDataRole.UserRole)`, which is str for JS-bridge/URL adds (`JavaScriptBridge.add_mod_from_js` is `Slot(str, str)`) but **int** for collection adds — `__compile_collection_datas` reads `metadata["publishedfileid"]` straight from Steam's WebAPI JSON, which returns numbers (browser.py:610–611). Meanwhile `steamcmd_mod_download_succeeded` always carries a str (`steamcmd_current_pfid` from stdout). So a collection-queued mod's int key never matches the str pop/in, and it reappears in the list even after a successful download. Normalizing pfids to `str` when adding (or in `get_download_list_snapshot`/`restore_download_list`) would harden this; it also aligns with the pre-existing `steamcmd_download_tracking.remove(current_pfid)` str path.

**3. `get_download_list_snapshot` can produce a `None` key (low).**
If an item has no UserRole data, `publishedfileid` is `None` and becomes a dict key despite the `dict[str, str]` annotation — `continue` on a falsy pfid (or coerce to str) is a cheap guard.

Non-blocking nits:
- `_on_steamcmd_mod_download_succeeded` calls another class's private `_remove_mod_from_list` — a public `remove_mod_if_queued()`-style helper would be cleaner.
- Behavior change worth a changelog line: closing the Mod Downloader now preserves the list even when no download is running (previously close = discard). Matches the intent, but users closing to abandon the queue get the old behavior back.

Also nice-to-have: an end-to-end test that a *failed* SteamCMD run (no success signals) leaves the snapshot fully intact through reopen.

pjgoliveira21:
> Solid fix — root cause (the download button emits `downloader_list_mods_tracking` itself, so `closeEvent` wipes it before SteamCMD starts) is real, and the snapshot/restore + per-mod success drop is a clean design. A few change requests before merge:
> 
> **1. Steamworks path can still silently drop queued mods on failure (medium).** `_do_steamworks_api_call_animated` pops every pfid out of `_pending_downloader_snapshot` unconditionally after the API call returns. But `_do_steamworks_api_call` can fail _quietly_ — e.g. `check_steam_available()` returns `False` and it early-returns without a raised exception (main_content_panel.py:2428–2430) — and the mods are dropped from the preserved list though they were never handled. The SteamCMD path only drops mods on a confirmed success signal; this path can't distinguish, defeating the fix's premise for the "Steam app" button. Suggest only dropping on actual success (e.g. skip the pop when `check_steam_available` fails / when the operation reports failure).
> 
> **2. int vs str pfid mismatch can break the snapshot drop (low–medium).** Snapshot keys come from `item.data(Qt.ItemDataRole.UserRole)`, which is str for JS-bridge/URL adds (`JavaScriptBridge.add_mod_from_js` is `Slot(str, str)`) but **int** for collection adds — `__compile_collection_datas` reads `metadata["publishedfileid"]` straight from Steam's WebAPI JSON, which returns numbers (browser.py:610–611). Meanwhile `steamcmd_mod_download_succeeded` always carries a str (`steamcmd_current_pfid` from stdout). So a collection-queued mod's int key never matches the str pop/in, and it reappears in the list even after a successful download. Normalizing pfids to `str` when adding (or in `get_download_list_snapshot`/`restore_download_list`) would harden this; it also aligns with the pre-existing `steamcmd_download_tracking.remove(current_pfid)` str path.
> 
> **3. `get_download_list_snapshot` can produce a `None` key (low).** If an item has no UserRole data, `publishedfileid` is `None` and becomes a dict key despite the `dict[str, str]` annotation — `continue` on a falsy pfid (or coerce to str) is a cheap guard.
> 
> Non-blocking nits:
> 
> * `_on_steamcmd_mod_download_succeeded` calls another class's private `_remove_mod_from_list` — a public `remove_mod_if_queued()`-style helper would be cleaner.
> * Behavior change worth a changelog line: closing the Mod Downloader now preserves the list even when no download is running (previously close = discard). Matches the intent, but users closing to abandon the queue get the old behavior back.
> 
> Also nice-to-have: an end-to-end test that a _failed_ SteamCMD run (no success signals) leaves the snapshot fully intact through reopen.

Thanks for the detailed pass — addressed all three, plus the nits:

**1. Steamworks path could silently drop queued mods on failure**
`_do_steamworks_api_call` now returns a `bool` indicating whether it actually dispatched the instruction (`False` when `check_steam_available()` fails, another call is already in flight, or the instruction is unsupported). `_do_steamworks_api_call_animated` only pops the instruction's mods from `_pending_downloader_snapshot` when that comes back `True` — a quiet early-return no longer causes queued mods to vanish.

**2. int vs str pfid mismatch**
Normalized `publishedfileid` to `str` right at `_add_mod_to_list`, the single choke point every add path (JS bridge, URL parse, collection expansion) funnels through. A collection-sourced int pfid and a JS-bridge str pfid are now always comparable, so `steamcmd_mod_download_succeeded`'s str pfid correctly matches and drops it.

**3. `get_download_list_snapshot` could produce a `None` key**
Added a guard that skips any list item with no `UserRole` data instead of letting a falsy `publishedfileid` become a dict key.

**Nits:**
- Added `SteamBrowser.remove_mod_if_queued()` — a public, quiet ("not queued" isn't warning-worthy for this caller) wrapper. `MainContent` now calls that instead of reaching into `_remove_mod_from_list`.
- Added the suggested test: a SteamCMD run reporting zero successes leaves the full original wait-list intact through a browser close + reopen.
- On the changelog note: agreed, worth flagging explicitly. Closing the Mod Downloader now preserves the queue unconditionally, not just while a download is active — previously, closing for any reason (including just deciding not to download) discarded it, which was an implicit second way to clear the list. "Clear List" is now the only way to intentionally empty the queue. Worth a release-note mention so existing users aren't surprised their old queue reappears.

All existing + new tests passing locally, pyright/jscpd clean.

LionelColaso (approved):
Reviewing the updated version: all three requested changes and the nits are addressed and verified.

1. **Steamworks quiet-failure drop** — _do_steamworks_api_call now returns a ool (dispatched vs. skipped), and _do_steamworks_api_call_animated only pops the instruction's mods from the pending snapshot when it was actually dispatched; a silent early-return (check_steam_available fail) no longer discards queued mods. Tested by 	est_ids_are_kept_when_the_call_was_not_dispatched.
2. **int/str pfid normalization** — _add_mod_to_list coerces to str at the single choke point, so WebAPI-collection adds and JS/URL adds are comparable with the SteamCMD success str. Tested.
3. **None snapshot key** — guarded in get_download_list_snapshot. Tested.
- Nits: public 
emove_mod_if_queued() removes the cross-class private call; suggested failed-run preservation test added.

Full CI green (lint/jscpd, pyright x3, pytest x3, CodeQL, builds, codecov/patch). LGTM, approving.

